PR #29, opened 2016-06-28
Disable spinlock when building on ARM.

fancycode:
The current spinlock code uses the "rep" instruction, which is not
supported on the ARM architecture.


kozlovic:
Thanks for the contribution!

PR #34, opened 2016-07-30
[ADDED] Optional global message delivery and async subscriptions with timeout

kozlovic:


src/nats.c
```diff
@@ -879,6 +928,17 @@ nats_Open(int64_t lockSpinCount)
...
     if (s == NATS_OK)
+        s = natsMutex_Create(&(gLib.dlvWorkers.lock));
+    if (s == NATS_OK)
+    {
+        gLib.libHandlingMsgDeliveryByDefault = (getenv("NATS_DEFAULT_TO_LIB_MSG_DELIVERY") != NULL ? true : false);
+        gLib.dlvWorkers.maxSize = 2;
+        gLib.dlvWorkers.workers = NATS_CALLOC(gLib.dlvWorkers.maxSize, sizeof(natsMsgDlvWorker*));
+        if (gLib.dlvWorkers.workers == NULL)
+            s = NATS_NO_MEMORY;
+    }
```
ishbir: I don't see `NATS_DEFAULT_TO_LIB_MSG_DELIVERY` documented anywhere.

src/sub.c
```diff
@@ -164,18 +182,79 @@ natsSub_deliverMsgs(void *arg)
...
+    // a "timeout" control message, do nothing.
+    if (!sub->closed && !sub->timedOut && !sub->timeoutSuspended)
+    {
+        // Prevent from scheduling another control message while we are not
+        // done with previous one.
+        sub->timedOut = true;
+
+        // Set the timer to a very high value, it will be reset from the
+        // worker thread.
+        natsTimer_Reset(sub->timeoutTimer, 60*60*1000);
```
ishbir: What's the point of this if the subscription will be destroyed anyway?

ishbir:
Great work, @kozlovic! It's a huge PR and I couldn't look through it very finely but I did read through it and left some comments. Thanks for your effort.

src/sub.c
kozlovic: It will not, that's the thing. If we are here in the code, it means that the async subscription timed-out. A "control" message is posted to the worker thread. There, the sub's callback will be invoked with a `NULL` message. Then it all depends what the user is doing. If the subscription is destroyed in the callback, proper cleanup will happen. If nothing is done, the subscription can still receive messages or timeout again. This is similar at calling `natsSubscription_NextMsg` that would get a timeout. The subscription is not automatically destroyed. You can call it again and may get messages or another timeout, etc...
Doing an "auto-destroy-free-resource" was too dangerous. This way, it makes it clear and the responsibility of the user to decide when to destroy the subscription.

ishbir: That's what I meant actually. Why do you reset the timer if the subscription will be destroyed in the callback function of the user?

kozlovic: First, I need a reset because I don't want the timer to fire for nothing since the rest is handled in the worker thread. There may be tons of messages queued-up (for this or other subscriptions), and there is no point of having the timer fire to realize that it has nothing to do because the previous timeout event is not yet processed.
Secondly, I do not know if the subscription will be destroyed. The user does not have to destroy it. It can keep it and still receive further messages.

Here is the doc for one of the timeout API:
```
/** \brief Creates an asynchronous subscription with a timeout.
 *
 * Expresses interest in the given subject. The subject can have wildcards
 * (see \ref wildcardsGroup). Messages will be delivered to the associated
 * #natsMsgHandler.
 *
 * If no message is received by the given timeout (in milliseconds), the
 * message handler is invoked with a `NULL` message.<br>
 * You can then destroy the subscription in the callback, or simply
 * return, in which case, the message handler will fire again when a
 * message is received or the subscription times-out again.
 *
 * \note Receiving a message reset the timeout. Until all pending messages
 * are processed, no timeout will occur. The timeout starts when the
 * message handler for the last pending message returns.
 *
 * \warning If you re-use message handler code between subscriptions with
 * and without timeouts, keep in mind that the message passed in the
 * message handler may be `NULL`.
 *
 * @param sub the location where to store the pointer to the newly created
 * #natsSubscription object.
 * @param nc the pointer to the #natsConnection object.
 * @param subject the subject this subscription is created for.
 * @param timeout the interval (in milliseconds) after which, if no message
 * is received, the message handler is invoked with a `NULL` message.
 * @param cb the #natsMsgHandler callback.
 * @param cbClosure a pointer to an user defined object (can be `NULL`). See
 * the #natsMsgHandler prototype.
 */
```
Let me know if this clears things up.

src/nats.c
kozlovic: I wanted to use that as a backdoor for those like you that would want the behavior to always be "a global message delivery thread pool" as opposed to having each subscription creates its delivery thread, without having to bother creating and setting the option (through `natsOptions_UseGlobalMessageDelivery`). Incidentally, I am using this for the test suite and valgrind to verify that all tests can use that way of message delivery and still work properly without memory leak (which I have included in the Travis build matrix).
I will see how I can document this in doxygen.

src/sub.c
ishbir: This function is reached if a timer times out and if it already has timed out, then why reset it if you don't want it to be fired again? Sorry but I'm not very familiar with how cnats works on the inside.

kozlovic: That's how my timers work, they will fire again automatically, that is, it is not a one time thing. After creating a timer and assign a callback to the timer, the callback will fire every given interval, without the need to call reset. Here I call reset to "stop" the firing since the worker thread will take care of the reset after the user callback returns.

ishbir: Gotcha! Thanks for clarifying.

kozlovic:
@ishbir I really appreciate that you took the time to review.

I have updated the docs to document the environment variable. I just wanted to know if you would be happy with the added APIs and if this something that would work for you (see screen shot of new doc for one of the API).

I think that these are great additions to the NATS C Client, so I am planning to merge this to master soon. I have some more changes to make to support async INFOs from server's cluster auto-discovery feature and will then probably cut a new release. Again, thanks for the great feedback!

screen shot 2016-08-09 at 9 51 28 am

ishbir:
@kozlovic Thank you for adding the APIs. I'm no longer working on the project I needed them for but I'm glad I could provide feedback backed by an actual use case.

kozlovic:
@ishbir Fair enough! Again, I think this is going to be a good addition, so I will merge. Thanks!

PR #42, opened 2016-09-12
allow disabling of monotonic clock as JRuby FFI thinks it exposes and…

abrandoned:
… it does not

in writing an FFI bridge into cnats (as opposed to using EventMachine with Ruby) I came across inconsistencies in how different Rubies (JRuby and MRI in this case) expose `CLOCK_MONOTONIC` to their respective implementations of FFI

am able to build and use the shared library on both when compiling and just shutting off `CLOCK_MONOTONIC` from the cnats side

Still building out bridge, but mostly working examples using cnats and getting a compile flag to take out `CLOCK_MONOTONIC` would certainly help in cross-ruby support
(https://github.com/abrandoned/cnats/blob/abrandoned/no_monotonic_clock/natsffi.rb)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ option(NATS_BUILD_WITH_TLS "Build with TLS support" ON)
 option(NATS_BUILD_WITH_TLS_CLIENT_METHOD "Use TLS_client_method()" OFF)
 option(NATS_BUILD_LIBUV_EXAMPLE "Compile libuv example" OFF)
 option(NATS_BUILD_LIBEVENT_EXAMPLE "Compile libevent example" OFF)
+option(NATS_BUILD_WITHOUT_CLOCK_MONOTONIC "Compile without respecting CLOCK_MONOTONIC" OFF)
 
 if(NATS_BUILD_WITH_TLS)
 find_package(OpenSSL REQUIRED)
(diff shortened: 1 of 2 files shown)


kozlovic:
Thank you for the contribution! It's cool that you are able to use NATS C Client from Ruby. As you know, we also provide a Ruby [implementation](https://github.com/nats-io/ruby-nats). How does FFI bridge compares to that?

abrandoned:
@kozlovic yeah, very familiar with the Ruby client, some problems that are rough in a larger infrastructure (Eventmachine is not the most loved lib to be packaging with a deploy, it leaks a lot and has some problems) .... also is not thread-safe in publishing (https://github.com/abrandoned/ruby-nats-multithread_publish) have a multithread_publish fork that maintain strats for making it threadsafe in publishing but the abstractions used can be costly to perf

using multithread publish can push 400-500,000 publishes/second through Ruby EM client (which is great), but the tests with cnats FFI and 4 publisher threads are pushing 11.2 million msg/sec .... the cnats bridge is quite small comparatively and we are quite interested in seeing if we can push more the FFI direction for higher throughput using better patterns for messaging from JRuby (which supports threading)

kozlovic:
Thanks for the feedback and good luck with your project. Let me know if you have any issue with the C client!

PR #46, opened 2017-01-04
update docs to reflect changed function name

abrandoned:
Readme not using current name


kozlovic:
@abrandoned Thanks for pointing this out. Actually, the whole paragraph need to be removed since this API has been deprecated and will be removed in the future: http://nats-io.github.io/cnats/deprecated.html.
I am going to close this PR and remove the paragraph from README.md.
Thanks!

PR #49, opened 2017-01-24
Use ISO __func__ instead of __FUNCTION__

andrelucas:
gcc5 complains about __FUNCTION__.

### DIFF
--- a/src/err.h
+++ b/src/err.h
@@ -6,17 +6,18 @@
 
 #include "status.h"
 #include "nats.h"
+#include "natsp.h"
 
 #define NATS_SSL_ERR_REASON_STRING ERR_reason_error_string(ERR_get_error())
 
 #define nats_setDefaultError(e) nats_setError((e), "%s", natsStatus_GetText(e))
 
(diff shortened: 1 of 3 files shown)


kozlovic:
@andrelucas Thank you for your contribution. Unfortunately, I cannot merge this PR as-is: it would fail to compile on Windows (it compiles ok for me on Mac OS/X and Ubuntu).

However, what you could do is replace references of `__func__` in this file with `__NATS_FUNCTION__`, add `#include "natsp.h"` and in `src/include/n-unix.h` add `#define __NATS_FUNCTION__  __func__` and in `src/include/n-win.h` add `#define __NATS_FUNCTION__   __FUNCTION__`. 

We could do all that directly in `err.h`, but doing the way I describe makes this available to other source files where we may need to have access to `__func__/__FUNCTION__` in the future.

Let me know if you would like to make those changes, or I can do it if you prefer.

andrelucas:
I'll gladly do that, thanks for the crystal-clear explanation.

kozlovic:
@andrelucas All good! Thanks again for the contribution.

PR #56, opened 2017-03-18
[FIXED] Crash in _updateStack (index out of range)

mpodanow:
Happens when application keeps reconnecting (e.g. maxReconnect=-1) and  errTL->func is filled with "natsConnection_FlushTimeout"/"natsConnection_Flush" sequences.

### DIFF
--- a/src/nats.c
+++ b/src/nats.c
@@ -1178,6 +1178,7 @@ _updateStack(natsTLError *errTL, const char *funcName, natsStatus errSts,
 
     idx = errTL->framesCount;
     if ((idx >= 0)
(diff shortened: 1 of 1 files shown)


kozlovic:
@mpodanow Thank you for the contribution! I am going to add a test for this in separate PR. Merging this one.

kozlovic:
@mpodanow Thanks for fixing this. The fix is good, but I am now wondering how and why the stack filled with those. I am not sure I am able to reproduce the filling up of the stack with those functions. Would you have a way to reproduce this? Thanks!

PR #60, opened 2017-03-31
State the documentation link near the top of the readme [skip ci]

qrpike:
I kept having to scroll around to find the docs link, usually it's at the very top. It may save some people a little time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # NATS - C Client 
 A C client for the [NATS messaging system](https://nats.io).
 
+Go [here](http://nats-io.github.io/cnats) for the online documentation.
+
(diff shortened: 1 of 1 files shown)


kozlovic:
The button named `doc|Doxygen` links to the documentation. Do you think that if we were to rename the label on this one it would have caught your attention, or is it better to simply have the text you added? (we could do both too)

qrpike:
@kozlovic Ah I didn't even see that. I personally don't look at those much because some repos just add a ton of meaningless badges. Now that I know about the badge it's 100% up to you.

kozlovic:
Ok, I will merge, but could you do me a favor and amend your commit to add `[ci skip]` in the summary (or anywhere in the commit). This is to prevent triggering Travis build that we don't need for such change. From your branch
```
git commit --amend
<edit the summary to add [ci skip] in there> then save
git push --force
```
Thanks!

kozlovic:
Thanks!

PR #61, opened 2017-03-31
Fixed max reconnects plurality

qrpike:


kozlovic:
Thanks for the contribution!

PR #67, opened 2017-05-02
Fix install prefix

dmpriso:
We did some preparing work for the integration to hunter: https://github.com/ruslo/hunter

Manual configuration of CMAKE_INSTALL_PREFIX in CMakeLists.txt is bad practice and conflicts with other toolchains (e.g. hunter)

Moved adapters to src/adapters because of installation; added files to installation. Fixed samples accordingly.


kozlovic:
Thank you very much for the contribution!

Issue #68, opened 2017-05-10
Fixing memory leak reports with VLD when linking statically


dmpriso:
As discussed on slack, VLD reports several small leaks when linking cnats statically. This is because thread-local and global allocations are actually released during cnats `DllMain()` function (or it's unix equivalent). However, `DllMain()` will not be called obviously when static linking (and may also introduce symbol conflicts when creating a DLL which statically links to cnats).
While the allocations are actually uncritical, since they are small and not growing, they may make hunting for actual leaks harder (in the end, everyone wants a clean memory report after debugging)

I've created a fix for that already but one small thing is still left to discuss.

First of all, I've created a `NATS_STATIC` symbol and avoid creating of `DllMain()` when defined. Then, I've created a `NATS_EXAMPLES_STATIC_LINKING` option to cmake for testing.

For the actual fix, I've created and implemented two new functions:

```C
/** \brief Per-Thread cleanup function for static library
*
* When linking to the static library, this should be called at the end of
* each thread where NATS objects have been created or used.
*
* \note This has no effect when called from the shared library.
* nats_Close() implicitly calls that.
*/
NATS_EXTERN void
nats_ReleaseThreadMemory(void);

/** \brief Global cleanup function for static library
*
* When linking to the static library, this should be called at the end of
* the application in order to perform library cleanup.
* IMPORTANT The library may not be re-used again after calling this.
*
* \note This has no effect when called from the shared library.
*/
NATS_EXTERN void
nats_ReleaseStaticLibrary(void);
```

`nats_ReleaseThreadMemory()` is already called by `nats_Close()`, so for many actual implementations it shouldn't be necessary to call that.

In my first try, I also called `nats_ReleaseStaticLibrary()` from `nats_Close()`. However this *may* break existing code: Until now, it was possible to "reuse" the library after `nats_Close()` has been called. However, I assume the intention of calling `nats_Close()` is to perform some final cleanup. If that is the case, we may also call `nats_ReleaseStaticLibraray()` from `nats_Close()`, making the whole static library cleanup automatic (except for multi-threaded applications where `nats_ReleaseThreadMemory()` is still required to be called for obvious reasons).

Background: `nats_ReleaseStaticLibrary()` performs the cleanup which a shared lib does during `DllMain(DLL_PROCESS_DETACH)`, which means deleting the mutex. However, the mutex cannot be reinitialized cause for doing so `gInitOnce` must be reset (which cannot be done in clean code; we could of course hack this)

[...]

PR #69, opened 2017-05-10
Memory cleanup for static

dmpriso:
Possible fixes for #68

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ option(NATS_BUILD_WITH_TLS "Build with TLS support" ON)
 option(NATS_BUILD_WITH_TLS_CLIENT_METHOD "Use TLS_client_method()" OFF)
 option(NATS_BUILD_LIBUV_EXAMPLE "Compile libuv example" OFF)
 option(NATS_BUILD_LIBEVENT_EXAMPLE "Compile libevent example" OFF)
+option(NATS_EXAMPLES_STATIC_LINKING "Use static linking for examples" OFF)
 
 if(NATS_BUILD_WITH_TLS)
 find_package(OpenSSL REQUIRED)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -19,7 +19,9 @@ endif()
 # Get all the .c files in the examples directory
 file(GLOB EXAMPLES_SOURCES RELATIVE ${PROJECT_SOURCE_DIR}/examples *.c)
 
-add_definitions(-Dnats_IMPORTS)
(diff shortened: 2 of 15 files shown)


kozlovic:
@dmpriso Please have a look at #70 to see if it solves your issue.

I took from your PR the call to `nats_ReleaseThreadMemory()` from the `_threadStart` function to reclaim thread-local storage from threads created by the library. There is one example where we create a user thread and perform NATS calls, so I call this function there.

But unlike in your PR, this function is not restricted to static library use. Any thread that the library creates needs to be cleaned-up, and on non Windows build, that was not happening at all (except for the main thread).

I also replaced the gcc's destructor and DllMain with `atexit()` which apparently serves the same purpose. I have verified that the final cleanup is invoked in all cases (examples statically or dynamically linked). The tests running with valgrind do not show any memory leak or still reachable memory. The good thing about this approach is that, as you can see, there is no need for the user to call any call, regardless if using the library dynamically or statically.

Please let me know if PR #70 works for you. I will obviously credit you in the release notes since you have reported the issue and proposed a fix for that. Thanks again!

kozlovic:
By the way, I made 2 commits (the last being for the doc) so that code review can be simpler. Check the 1st commit for code-only changes: https://github.com/nats-io/cnats/pull/70/commits/a4f7830f75573c3412cec34915d25804cbd66342.

Also, in the future if you submit a PR, I would appreciate if you could replace tabs with 4 spaces to match the rest of the project. Thanks!

dmpriso:
@kozlovic I ran my own unit tests of our cnats integration and they do not leak any more with your PR #70. Global cleanup works correctly. As you already mentioned, I still need to call `nats_ReleaseThreadMemory()` from my own threads where I call nats functions in order to prevent all leaks.

While I don't use the shared library myself (as of now), I'd suggest to add `DllMain()` again for windows builds because with a `DllMain()` it is possible to do that per-thread cleanup automatically for threads created outside of the DLL. With your PR #70, existing users which are using the shared library on windows now need to call `nats_ReleaseThreadMemory()` - it is kind of a break of existing code which now leaks where it didn't leak before.

kozlovic:
@dmpriso Thanks for checking and you are right about `DllMain`. I have put it back just for the thread_detach case. For users on Windows using the dynamic library, this PR will result in identical behavior than before.

However, as I state in the comment, users that want to write portable applications that work the same if statically or dynamically linked, they should call `nats_ReleaseThreadMemory()` on the threads they create. In case of dynamic library, it means that `nats_ReleaseThreadMemory()` may be called twice per thread (one from user, one internally), but that's ok since this call does nothing if the thread local storage has already been freed.

Thanks again for your help! I am closing this PR in favor of PR #70.

PR #76, opened 2017-05-12
[CHANGED] Added nats to the include directory during install [ci skip]

### LINKED ISSUE #72: `status.h` and `version.h` don't belong in `$PREFIX/include/`
Currently, cnats installs the cnats-specific header files `status.h` and `version.h` to `$PREFIX/include/`. While the other header might make sense at `/usr/local/include/nats.h` and `/usr/include/nats.h`, `/usr/local/include/status.h` or `/usr/include/status.h` do not. Same goes for `version.h`.

**TL;DR:** They would make sense if they were installed as something like `/usr/include/nats-status.h` or `/usr/include/nats/status.h`

kozlovic:
When doing `make install`, the header files are now copied under
the `$CMAKE_INSTALL_PREFIX/include/nats`.

Resolves #72

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ endif(NATS_BUILD_WITH_TLS)
 #---------------------------------------------------------------------
 # Add to the 'clean' target the list (and location) of files to remove
 
+list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats.h")
 list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/nats.h")
 list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/status.h")
 list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/version.h")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,9 +94,11 @@ endif(NATS_BUILD_WITH_TLS)
 #---------------------------------------------------------------------
 # Add to the 'clean' target the list (and location) of files to remove
 
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/status.h")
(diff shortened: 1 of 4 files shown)


kozlovic:
With this changes, a `make install` with a `CMAKE_INSTALL_PREFIX` set to `/Users/ivan/dev/cnats/install` results in this:

```
Install the project...
-- Install configuration: "Release"
-- Installing: /Users/ivan/dev/cnats/install/lib/libnats.dylib
-- Installing: /Users/ivan/dev/cnats/install/lib/libnats_static.a
-- Up-to-date: /Users/ivan/dev/cnats/install/include/nats/nats.h
-- Up-to-date: /Users/ivan/dev/cnats/install/include/nats/status.h
-- Up-to-date: /Users/ivan/dev/cnats/install/include/nats/version.h
-- Up-to-date: /Users/ivan/dev/cnats/install/include/nats/adapters/libevent.h
-- Up-to-date: /Users/ivan/dev/cnats/install/include/nats/adapters/libuv.h
```
@pritambaral I believe this change addresses your concerns. @dmpriso Hope this still works for you.

My concern is that users that previously did a `make install` and used the `include` directory for their apps will have to update all their apps to pick up the headers. More importantly, they may not know that they are still using the old ones. If I go with this PR, I intend to warn users in the release notes, but not sure if this the right direction. What do you guys think?

pritambaral:
@kozlovic Yes, it does. In addition it also adds the adapter headers!

If worried about backwards compatibility, might I suggest a deprecation warning using the preprocessor?:

```c
// include/nats.h
#ifndef __WIN32
#warning "`#include <nats.h>` is deprecated. Please use `#include <nats/nats.h>`"
#else
#pragma message ("`#include <nats.h>` is deprecated. Please use `#include <nats/nats.h>`")
#endif

#include <nats/nats.h>
```

kozlovic:
@pritambaral Glad to know that this would work for you. I am a bit confused in your comment about preprocessor. Where would you want me to put this?

Keep in mind that the examples that I provide pickup the headers directly from the source code, so that would not help putting this in there.

pritambaral:
I meant the old include location.

If loading `$CMAKE_INSTALL_PREFIX/include/nats.h` could automatically load the new `$CMAKE_INSTALL_PREFIX/include/nats/nats.h`, that would serve to be backwards compatible; but you probably want to move off the former, so you can put that deprecation notice as a compiler warning (where supported).

dmpriso:
@kozlovic works for me, thanks!

PR #92, opened 2017-08-05
[FIXED] Installed adapters headers

### LINKED ISSUE #90: Installed header files include "nats.h", not "nats/nats.h"
Two of the installed header files include "nats.h", rather than "nats/nats.h":

```
$ pwd
/usr/local/include/nats
$ grep -ir "nats\.h" .
./adapters/libevent.h:#include "nats.h"
./adapters/libuv.h:#include "nats.h"
```

This causes warnings (from the pragma in "nats.h"):

```
$ clang  libuv-sub.c  -luv -lnats
In file included from libuv-sub.c:5:
In file included from ./examples.h:7:
/usr/local/include/nats.h:7:2: warning: "`#include <nats.h>` is deprecated. Please use `#include <nats/nats.h>`"
      [-W#warnings]
#warning "`#include <nats.h>` is deprecated. Please use `#include <nats/nats.h>`"
 ^
1 warning generated.
```

kozlovic:
Modified adapters (libevent and libuv) headers to include
`../nats.h`. This will work even in case when we build examples
with cmake that includes `src`, and also for users using
`/usr/local/include` and having their files include `nats/nats.h`
and `nats/adapters/libuv.h` (for instance).

Updated libuv adapter to use uv_poll_init() instead of uv_poll_init_socket()
since the later disappeared from libuv's master repo.

For non Windows builds, if examples are built statically, link
against static libevent/libuv libraries.
For Windows, if say "libuv.lib" is the lib from the static built of
that library, then the example will be statically linked with that.
If the ".lib" file is from the building of a shared library, the
example will need "libuv.dll" to be able to run.

On Windows, changed the "libevent" library name to "event" since
it looks like this is how the library is named now.

Resolves #90


src/adapters/libuv.h
```diff
@@ -191,7 +191,7 @@ uvAsyncAttach(natsLibuvEvents *nle)
         s = NATS_NO_MEMORY;
 
     if ((s == NATS_OK)
-        && (uv_poll_init_socket(nle->loop, nle->handle, nle->socket) != 0))
```
mdevan: natsLibuvEvents.socket is an int, and the third argument to uv_poll_init_socket is a uv_os_sock_t. On Unices uv_os_sock_t is an int, but on Windows it is a SOCKET. So this might not compile on Windows (don't have a machine handy to try out, sorry).

kozlovic: It actually compiles fine without any warning, but I will change the `(*natsEvLoop_Attach)` function prototype to use `natsSock` which represents properly a socket based on the platform it is built on (master is in a new release 1.7.0 schedule, so I am ok to change `nats.h` content). I will do so in upcoming PR.

PR #122, opened 2018-02-08
glibc 2.26 support FIX

martende:
There is a problem with glibc 2.26 compilation. 

They removed xlocale.h support since this version. 
https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27
This fix works only for linux and may be must be changed to make support of all other platforms.


kozlovic:
@martende Could you please try with PR #123? As you expected, your changes did not work on all platforms. Would not build on Travis and on macOS. Thanks for the report!

PR #166, opened 2018-08-23
provide access to STAN connection's NATS connection, but flag it uncl…

edam:
…osable

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2666,6 +2666,7 @@ natsConn_create(natsConnection **newConn, natsOptions *options)
     nc->refs        = 1;
     nc->sockCtx.fd  = NATS_SOCK_INVALID;
     nc->opts        = options;
+    nc->closeLock   = false;
 
     if (nc->opts->maxPingsOut == 0)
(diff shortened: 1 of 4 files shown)


kozlovic:
@edam Please have a look at PR #167. The direction I went was to also add a function that release the use of the NATS Connection. The streaming connection is actually retained. The idea is that if the streaming connection is destroyed but some code has still reference to the NATS connection and tries to pub/sub, it would fail (since the NATS connection is closed), but not crash. Users should call stanConnection_ReleaseNATSConnection() when done with it.

If you feel this is unnecessary and stating in the doc that if you call stanConnection_GetNATSConnection() and uses this connection, you must NOT do any NATS calls after stanConnection_Destroy() has been called, I will remove.

I will update (generate) docs (doxygen) once we are settled in the API(s).

Thanks again for your contribution!

edam:
Perfect, thanks!

kozlovic:
Closing this in favor of #167 that has been merged. You will be credited for your contribution in the release notes. Thanks!

PR #172, opened 2018-09-12
Use cmake variables for installation destinations

dtoubelis:
Location of libraries and development files may be different when building for different platforms. CMake can handle this just fine, however, the current cmake configuration does not use these facilities. This merge request fixes that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,14 +120,14 @@ endif(NATS_BUILD_WITH_TLS)
 #---------------------------------------------------------------------
 # Add to the 'clean' target the list (and location) of files to remove
 
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/nats.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/status.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/version.h")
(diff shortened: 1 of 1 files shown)


kozlovic:
Disclaimer, I am no expert on CMake, so would love if you could explain (or point me to) the difference between say `CMAKE_INSTALL_PREFIX}/lib` and `CMAKE_INSTALL_LIBDIR/`. Are you suggesting that the issue is that I "hardcoded" include and lib for the directories?
As it was, you could change CMAKE_INSTALL_PREFIX to point to location that you wanted, so that was not enough?
Again, nothing against the change, just trying to understand better. Thanks!

dtoubelis:
I'm not an expert on CMake either but we are building the client for a platform that puts libraries in `/usr/lib64`. The build system is yocto and it has `cmake` class that takes care of installing bits and pieces into correct locations. But having destinations hardcoded interferes with the intent.

dtoubelis:
You may want to hold off this patch though. I want to give it to our CMake expert to review. I may have missed something.

kozlovic:
Sure, thanks!

kozlovic:
that being said, your changes seem fine to me. Using those 2 cmake variables seem to work fine (include defaults to `include` and lib to `lib` on my macOS, which is what I was hardcoding with). But probably defaults to other values for other platforms.

dtoubelis:
I may have missed something, it is still not placing the libraries in `/usr/lib64` for me :-(

kozlovic:
Have you modified CMAKE_INSTALL_PREFIX? You can do that from `make edit_cache`.

dtoubelis:
I'm closing the pull request. It is completely wrong. First of all, `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` variables are defined by `GNUInstallDirs` module that is not included. And secondly, the part I modified is used to build uninstall target and has nothing to do with the installation. I will create a new pull request once I figure it out.

PR #174, opened 2018-09-13
[IMPROVED] Use CMake variable for deployment destinations

dtoubelis:
Here is the proper fix for the https://github.com/nats-io/cnats/pull/172.

Location of libraries and development files may be different when building for different platforms. CMake can handle this just fine, however, the current cmake configuration does not use these facilities. This merge request fixes that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,7 +139,6 @@ list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/${NATS_LIBDIR}/libnats
 set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES "${NATS_INSTALLED_FILES}")
 #---------------------------------------------------------------------
 
-if(NATS_UPDATE_VERSION OR NATS_UPDATE_DOC)
 #------------
 # Versionning and Doc
 
@@ -150,6 +149,7 @@ set(NATS_VERSION_SUFFIX "-beta")
 
 set(NATS_VERSION_REQUIRED_NUMBER 0x010100)
 
+if(NATS_UPDATE_VERSION OR NATS_UPDATE_DOC)
 configure_file(
 	${CMAKE_CURRENT_SOURCE_DIR}/src/version.h.in
 	${CMAKE_CURRENT_SOURCE_DIR}/src/version.h
@@ -159,8 +159,8 @@ configure_file(
 	${CMAKE_SOURCE_DIR}/doc/DoxyFile.NATS.Client.in
 	${CMAKE_SOURCE_DIR}/doc/DoxyFile.NATS.Client
 	@ONLY)
-#------------
 endif(NATS_UPDATE_VERSION OR NATS_UPDATE_DOC)
+#------------
 
 #----------------------------
 # Add the project directories
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -33,12 +33,15 @@ add_library(nats SHARED ${SOURCES} ${PS_SOURCES} ${S_SOURCES})
 target_link_libraries(nats ${NATS_OPENSSL_LIBS} ${NATS_EXTRA_LIB} ${NATS_PROTOBUF_LIB})
 add_library(nats_static STATIC ${SOURCES} ${PS_SOURCES} ${S_SOURCES})
 target_link_libraries(nats_static ${NATS_OPENSSL_LIBS} ${NATS_PROTOBUF_LIB})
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -120,19 +127,18 @@ endif(NATS_BUILD_WITH_TLS)
...
 
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/nats.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/status.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/version.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/adapters/libevent.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/include/nats/adapters/libuv.h")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/lib/libnats${CMAKE_SHARED_LIBRARY_SUFFIX}")
-list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/lib/libnats_static${CMAKE_STATIC_LIBRARY_SUFFIX}")
+list(APPEND NATS_INSTALLED_FILES "${CMAKE_INSTALL_PREFIX}/${NATS_INCLUDE_DIR}/nats.h")
```
kozlovic: Actually, I used to have this to remove installed libs and headers on a make clean, which I am not sure this is the right thing to do anyway. So why don't you remove the whole section (from `#----` above to `#---` after the `set_directory_properties`.

With your changes, this would not remove the libs with the version number anyway.

src/CMakeLists.txt
```diff
@@ -33,12 +33,15 @@ add_library(nats SHARED ${SOURCES} ${PS_SOURCES} ${S_SOURCES})
...
 
 # --------------------------------------
 # Install the libraries and header files
 # --------------------------------------
-install(TARGETS nats DESTINATION lib)
-install(TARGETS nats_static DESTINATION lib)
-install(FILES deprnats.h DESTINATION include RENAME nats.h)
-install(FILES nats.h status.h version.h DESTINATION include/nats)
-install(FILES adapters/libevent.h adapters/libuv.h DESTINATION include/nats/adapters)
+install(TARGETS nats LIBRARY DESTINATION ${NATS_LIBDIR})
```
kozlovic: This does not work on Windows. See (https://cmake.org/cmake/help/v3.0/command/install.html):
```
For DLL platforms the DLL part of a shared library is treated as a RUNTIME target
and the corresponding import library is treated as an ARCHIVE target.
All Windows-based systems including Cygwin are DLL platforms. 
```
So either you replace this line with:
```
if(WIN32)
  install(TARGETS nats RUNTIME DESTINATION ${NATS_LIBDIR})
else(WIN32)
  install(TARGETS nats LIBRARY DESTINATION ${NATS_LIBDIR})
endif(WIN32)
```
or simply remove `LIBRARY` and simply have:
```
install(TARGETS nats DESTINATION ${NATS_LIBDIR})
```
I tested and it still installs the shared library on macOS, Linux and Windows. If it does not work on your system and you need to specifically have `LIBRARY`, then you need the `ifdef` as shown above.

dtoubelis:
Sounds good. We will look into it.

dtoubelis:
Please have a look at it now.

src/CMakeLists.txt
```diff
@@ -33,12 +33,15 @@ add_library(nats SHARED ${SOURCES} ${PS_SOURCES} ${S_SOURCES})
...
 
 # --------------------------------------
 # Install the libraries and header files
 # --------------------------------------
-install(TARGETS nats DESTINATION lib)
-install(TARGETS nats_static DESTINATION lib)
-install(FILES deprnats.h DESTINATION include RENAME nats.h)
-install(FILES nats.h status.h version.h DESTINATION include/nats)
-install(FILES adapters/libevent.h adapters/libuv.h DESTINATION include/nats/adapters)
+install(TARGETS nats DESTINATION ${NATS_LIBDIR})
```
kozlovic: I gather than removing `LIBRARY` works (install the shared lib in the appropriate directory) on your platform of choice, right?

antlad: @kozlovic Yes, all is working as expected.

PR #183, opened 2018-09-28
added additional architecture support when mutex locking with spin

wrightleft:
We were running into compile errors in `src/unix/mutex.c`:
```bash
Error: unrecognized opcode: `rep'
```
Which, the comments suggest is expected for ARM and ARM64, but PowerPC also doesn't have the instruction used. 

There are several good kernel macro to use, like `spin_until_cond` or `cpu_relax`, but they either are too new in the kernel, or require the kernel headers, which is something I didn't think the project wanted to depend on.

I also tried to find a good pthread or glibc alternative, but came up short. So I borrowed from [an implementation of `cpu_relax` in boost](https://www.boost.org/doc/libs/1_68_0/boost/fiber/detail/cpu_relax.hpp) to implement a few more branches to support more architectures.

I didn't test this to see if it actually accomplishes anything more than just calling `pthread_mutex_lock`, but I assume the code to spin was there for a reason.

### DIFF
--- a/src/unix/mutex.c
+++ b/src/unix/mutex.c
@@ -70,17 +70,25 @@ natsMutex_TryLock(natsMutex *m)
 void
 natsMutex_Lock(natsMutex *m)
 {
-    // The "rep" instruction used for spinning is not supported on ARM.
-#if !defined(__arm__) && !defined(__aarch64__)
     if (gLockSpinCount > 0)
     {
         int64_t attempts = 0;
-
         while (pthread_mutex_trylock(m) != 0)
         {
             if (++attempts <= gLockSpinCount)
             {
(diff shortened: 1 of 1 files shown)


kozlovic:
@wrightleft Thank you for the contribution. I tested and compiles fine on macOS and and Linux VM and on Windows too. I trust that it works with powerPC since this is why you submitted the PR. The one I cannot test is ARM/ARM64. Are you able to check on that platform?

kozlovic:
@wrightleft Ok, merging for now and will see if we have to iterate. Thanks again for your contribution!

wrightleft:
Sorry for the lack of communication. Thanks for following through.

PR #187, opened 2018-10-16
[ADDED] nats_CloseAndWait() to wait for library tear down

### LINKED ISSUE #184: Race condition likely during unsubscribe
Hi @kozlovic ,

You handled a previous (invalid) test case from me:
https://github.com/nats-io/cnats/issues/154

After continuing to experience problems, though, I updated and improved the tests, and I think I found that the original issue may indeed exist.

Attached are the updated test case. To use:
```
# ensure gnatsd is running, and then ...
make
./test_cnats.out 0
./test_cnats.out 1
./test_cnats.out 2
# etc...
```

What I've found is that tests 9 and 10 fail. A workaround is to introduce a sleep between `natsSubscription_Unsubscribe` and `natsSubscription_Destroy` (line 99 in the provided cnats_iface.c).

I haven't identified the fix yet. It seems the gLib ref count doesn't get decrement completely, perhaps because the thread in natsSub_deliverMsg doesn't always break out at
```
if (sub->closed)
{
    natsSub_Unlock(sub);
    break;
}
```

I'll investigate further if time permits.

Thanks.

[cnats_connect_disconnect_issue.tar.gz](https://github.com/nats-io/cnats/files/2453355/cnats_connect_disconnect_issue.tar.gz)

kozlovic:
This new API closes the library and blocks until all NATS threads
return and resources have been released. This will return an error
if called from a NATS thread (such as message, connection, error
callbacks, etc..)

Resolves #184


kozlovic:
@rudyjantz verified that this PR works.

PR #189, opened 2018-10-17
[IMPROVED] natsConnection_Close() now internally calls FlushTimeout()

### LINKED ISSUE #188: Send buffers not cleared on nats-close
This is related to,
- Issue https://github.com/nats-io/cnats/issues/184
- PR https://github.com/nats-io/cnats/pull/187

Scenario:
A sender process uses publish to send data to a receiver process. Then that publisher immediately closes cnats.

Behavior:
The receiver does not always receive the message.

Expected behavior:
Unsure.
  - Without PR https://github.com/nats-io/cnats/pull/187, it's sort of expected that the publish may not complete. Though it's using the synchronous version of publish, cnats documentation states that data is buffered. Even with the send-asap option set to true, cnats is best effort. So it's understandable if the message does not arrive.
  - With PR https://github.com/nats-io/cnats/pull/187, it's less clear. That's mostly the purpose of raising this issue. Because `nats_CloseAndWait()` is designed to tear down cleanly, should that include completion of any buffered sends?

I've attached a small set of tests. There are 3 executables. 
- test_cnats.out is the "base" version.
- test_cnats_flush.out calls `natsConnection_Flush()` after publishing.
- test_cnats_sendasap.out uses `natsOptions_SetSendAsap(opts, true)`.
All 3 make use of PR https://github.com/nats-io/cnats/pull/187. The tests of interest are test-4 and test-5. To use,
```
$ make
./test_cnats.out 4 # should pass, and does
./test_cnats.out 5 # should pass with PR 187? fails about half of the time or more on my test machine
./test_cnats_sendasap 4 # should pass, and does
./test_cnats_sendasap 5 # should pass with PR 187? fails about half of the time or more on my test machine
./test_cnats_flush 4 # should pass, and does
./test_cnats_flush 5 # should pass, and does
```

If this is expected behavior, then please ignore. But in case it was overlooked, I wanted to raise it before PR https://github.com/nats-io/cnats/pull/187 is merged.

[cnats-pr-187-pub-then-disconnect-test.tar.gz](https://github.com/nats-io/cnats/files/2482700/cnats-pr-187-pub-then-disconnect-test.tar.gz)

kozlovic:
This ensures that anything that is in the buffer is not only flushed
but that the server has processed anything that was sent by the
connection prior to the close call.

Resolves #188

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -80,6 +80,9 @@ _close(natsConnection *nc, natsConnStatus status, bool doCBs);
 static bool
 _processOpError(natsConnection *nc, natsStatus s, bool initialConnect);
 
+static natsStatus
+_flushTimeout(natsConnection *nc, int64_t timeout);
+
 
 /*
  * ----------------------------------------
@@ -2072,6 +2075,12 @@ _close(natsConnection *nc, natsConnStatus status, bool doCBs)
 
     natsConn_lockAndRetain(nc);
 
+    // If connection is not already closed and currently connected, attempt to
+    // send a PING and get a PONG to ensure that not only buffers are flushed
+    // but that server has processed anything that was pending.
+    if (!natsConn_isClosed(nc) && (nc->status == NATS_CONN_STATUS_CONNECTED))
+        _flushTimeout(nc, 2000);
+
     if (natsConn_isClosed(nc))
     {
         nc->status = status;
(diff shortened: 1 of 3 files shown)


kozlovic:
@rudyjantz Tried this PR with his test case and confirmed that it works.

PR #198, opened 2018-12-05
Fixes to make it compile under FreeBSD (targets nats / nats_static).

morenocarullo:
Not perfect yet, as not everything compiles. But at leat nats and nats_static do.


kozlovic:
@morenocarullo Sorry for the delay. Could you try to build from master (without the changes that you have here?) and see if it works for you on FreeBSD? I have made some changes that for instance removes the need for locale. If the only thing that is missing in `include/n-unix.h` is the `#include <sys/socket.h>` I could add that. I don't think that there is a need for a FreeBSD OS flag if that's the case. Thanks!

kozlovic:
@morenocarullo Have you had a chance to try my suggestion? I am not sure we need a special flag for FreeBSD. Again, since I have removed the need from locale, if the only thing missing is the include in n-unix.h, we could simply add that. Thanks!

kozlovic:
Closing for now.. Like I said, I don't think we need a special handling for FreeBSD since I have removed need for locale. If the only one missing is the socket include, we could add that no problem.

PR #211, opened 2019-05-09
add the cmake options NATS_BUILD_STATIC_LIBRARY & NATS_BUILD_SHARED_LIBRARY to control which type of library is built

ksachdeva:
This helps in speeding up the build (by choosing only one of them) when cnats is being used either with cmake ExternalProject or FetchContent especially in Continuous Integration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,9 @@ option(NATS_BUILD_STATIC_EXAMPLES "Statically link examples" OFF)
 option(NATS_BUILD_STREAMING "Build NATS Streaming" ON)
 option(NATS_BUILD_NO_PREFIX_CONNSTS "No prefix for connection status enum" OFF)
 option(NATS_COMPILER_HARDENING "Compiler hardening flags" OFF)
+option(NATS_BUILD_STATIC_LIBRARY "Build static library" ON)
+option(NATS_BUILD_SHARED_LIBRARY "Build shared library" ON)
+
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/coveralls-cmake/cmake)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ option(NATS_BUILD_STATIC_EXAMPLES "Statically link examples" OFF)
 option(NATS_BUILD_STREAMING "Build NATS Streaming" ON)
 option(NATS_BUILD_NO_PREFIX_CONNSTS "No prefix for connection status enum" OFF)
 option(NATS_COMPILER_HARDENING "Compiler hardening flags" OFF)
+option(NATS_BUILD_STATIC_LIBRARY "Build static library" ON)
```
kozlovic: Could we make them `NATS_BUILD_LIB_STATIC` and `NATS_BUILD_LIB_SHARED`? This is not that important, but when doing `make edit_cache` this way they would be close to each other, that's all.

src/CMakeLists.txt
```diff
@@ -29,18 +29,30 @@ endif(NATS_BUILD_STREAMING)
...
 
 # --------------------------------------
 # Install the libraries and header files
 # --------------------------------------
-install(TARGETS nats DESTINATION ${NATS_LIBDIR})
+if(NATS_BUILD_SHARED_LIBRARY)
+  install(TARGETS nats DESTINATION ${NATS_LIBDIR})
+endif(NATS_BUILD_SHARED_LIBRARY)
+
 install(TARGETS nats_static ARCHIVE DESTINATION ${NATS_LIBDIR})
```
kozlovic: You should have the same `if` block with `NATS_BUILD_STATIC_LIBRARY` for that line otherwise it would fail if one sets that option to `OFF`.

kozlovic:
Closing this one in favor of #257 (I updated your branch, you will get the proper credits).

PR #225, opened 2019-05-29
Add more compilers to CI build

rigtorp:
Some of the tests are broken on newer versions of GCC

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ html/
 .project
 .settings/
 
+.vscode/
+
 # Mac
 .DS_Store
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
     - unzip
 
 before_install:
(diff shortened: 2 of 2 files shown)


kozlovic:
@rigtorp Thank you for this PR! This is a lot of builds in there ;-) Maybe we could limit the number of gcc compilers versions, etc.. but I am glad that you posted this so now I know how to select a specific gcc version in Travis :-)
Will investigate how we could consolidate a bit this matrix (I already found that it was taking a long time before...).

rigtorp:
True it might be excessive. Main goal was to get GCC9 and clang 8 builds going so latest ASAN/TSAN/UBSAN can be used.

kozlovic:
@rigtorp Thanks for the contribution. I will merge this PR and follow right away with the fixes and a bit of rework of the list. Will limit to gcc-9 and clang-8 (add both sanitize flags to that one).

Issue #226, opened 2019-06-01
Remove call to ERR_remove_thread_state which is deprecated in OpenSSL


danesh-d:
The call to ERR_remove_thread_state is deprecated in OpenSSL 1.1.0. The thread initialization and deinitialization is now done automatically by OpenSSL (see here https://www.openssl.org/docs/man1.1.0/man3/ERR_remove_state.html).

PR #227, opened 2019-06-01
[FIXED] Remove calls to "ERR_remove_thread_state()" which was depreca…

danesh-d:
…ted in OpenSSL 1.1.0

The thread initialisation and deinitialisation is done automatically in
OpenSSL from version 1.1.0.

### DIFF
--- a/src/nats.c
+++ b/src/nats.c
@@ -21,6 +21,7 @@
 #include <inttypes.h>
 #include <assert.h>
 #include <stdarg.h>
+#include <openssl/opensslv.h>
 
 #include "mem.h"
 #include "timer.h"
@@ -137,7 +138,8 @@ static void
 _cleanupThreadSSL(void *localStorage)
 {
 #if defined(NATS_HAS_TLS)
-    ERR_remove_thread_state(0);
+    if (!(natsOpenSSL_Version(1, 1, 0, OPENSSL_VERSION_NUMBER)))
+        ERR_remove_thread_state(0);
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


danesh-d:
This patch addresses the issue #226.

kozlovic:
@danesh-d Thanks for the contribution! However, I would like the change to take into account the version of OpenSSL. As you can see from the travis build result, since an older OpenSSL library is used, this causes memory leaks:
```
131/176 MemCheck: #131: Test_SSLBasic ...........................   Defects: 6
132/176 MemCheck: #132: Test_SSLVerify ..........................   Defects: 6
133/176 MemCheck: #133: Test_SSLVerifyHostname ..................   Defects: 6
134/176 MemCheck: #134: Test_SSLSkipServerVerification ..........   Defects: 6
135/176 MemCheck: #135: Test_SSLCiphers .........................   Defects: 8
136/176 MemCheck: #136: Test_SSLMultithreads ....................   Defects: 8
137/176 MemCheck: #137: Test_SSLConnectVerboseOption ............   Defects: 6
```

danesh-d:
@kozlovic Sure. I will re-submit updated version of this patch.

danesh-d:
@kozlovic The code uses deprecated function only for older versions of OpenSSL, otherwise the function is not called. I also added unit test for the function that compares OpenSSL version. Tests are also passed.

kozlovic:
@danesh-d Thank you for the contribution! I tested and it works on Windows too.

Issue #228, opened 2019-06-01
Replace calls to "ASN1_STRING_data" with "ASN1_STRING_get0_data"


danesh-d:
According to recommendation from OpenSSL (see here https://www.openssl.org/docs/man1.1.0/man3/ASN1_STRING_data.html) the function "ASN1_STRING_data" is deprecated and should be replaced by "ASN1_STRING_get0_data".

kozlovic:
Closed due to PR #238.

PR #229, opened 2019-06-01
[FIXED] Remove call to "ASN1_STRING_data" which was deprecated

danesh-d:
The call to "ASN1_STRING_data" is deprecated and is replaced by
"ASN1_STRING_get0_data" according to OpenSSL's recommendation.

### DIFF
--- a/src/opts.c
+++ b/src/opts.c
@@ -298,8 +298,13 @@ _hostnameMatchesSubjectAltName(char *hostname, X509 *cert)
         if (namePart->type == GEN_DNS)
         {
             foundAnyEntry = true;
-            foundMatch = _hostnameMatches((char*) ASN1_STRING_data(namePart->d.uniformResourceIdentifier),
-                                           hostname);
+
+            if (!(natsOpenSSL_Version(1, 1, 0, OPENSSL_VERSION_NUMBER)))
+                foundMatch = _hostnameMatches((char*) ASN1_STRING_data(namePart->d.uniformResourceIdentifier),
+                                               hostname);
+            else
+                foundMatch = _hostnameMatches((char*) ASN1_STRING_get0_data(namePart->d.uniformResourceIdentifier),
+                                               hostname);
         }
 
         GENERAL_NAME_free(namePart);


danesh-d:
This patch addresses the issue #228.

danesh-d:
I can build the code without any problem on my local machine. It seems that on the build servers for CI, the libraries (namely OpenSSL) is not updated since the compiler cannot find the declaration of the function and then linker cannot find the implementation. I use gcc 8.3.1-2 and OpenSSL 1.1.1b FIPS.

kozlovic:
@danesh-d Thank you for your contribution. As in the other PR, I'd like to see a change that would allow OpenSSL prior to 1.0/1.0.1 to work too. One way to at least mitigate this would be to select a build option to build the library with or without these calls. But since the OpenSSL library may be dynamically linked, I am not sure this is the right approach.

danesh-d:
@kozlovic Yes exactly. Another way would be to read the current version of OpenSSL which is reported by API and then judge whether deprecated version or the new one should be used. actually I had implemented with this check, but I though maybe it is not intended that the library will work with older versions on OpenSSL as people are always encouraged to keep OpenSSL installed on their machine, updated. I will modify this patch so you will see if it would be something useful or not.

src/opts.c
```diff
@@ -298,8 +298,13 @@ _hostnameMatchesSubjectAltName(char *hostname, X509 *cert)
         if (namePart->type == GEN_DNS)
         {
             foundAnyEntry = true;
-            foundMatch = _hostnameMatches((char*) ASN1_STRING_data(namePart->d.uniformResourceIdentifier),
-                                           hostname);
+
+            if (!(natsOpenSSL_Version(1, 1, 0, OPENSSL_VERSION_NUMBER)))
```
kozlovic: Unfortunately, that does not work since you are detecting at runtime but then have in the code reference to a function that may not even be known at the time of the building of the library. (see travis build report that fails with openssl pre 1.1.0).

danesh-d: Yes exactly! The check should be in compile time. I will fix it if later when I have time if you think it is worth to be implemented. Thanks.

kozlovic:
@danesh-d Closing this since this is no longer relevant following PR #238 where those calls have been removed.

PR #231, opened 2019-06-03
[FIXED] Add a marginal to buffers to silent truncated buffer warnings

danesh-d:


danesh-d:
This patch addressed issue #231.

kozlovic:
@danesh-d Not sure about this fix. Clearly, the two tests could be easily fixed by just copying directly in errProto and if needed make the stack buffer bigger without the need for the constant you use.

What are the warnings you are referring to and how did you get it to print them? I just checked the code and all places where I use the connection's errTxt buffer, I do use `snprintf()`, which as far as I know, will ensure that I do not overflow the buffer and makes sure that a terminating `\0` is added.

danesh-d:
Sure you know better how this can be fixed. I am sorry that I missed to update the commit message. I have two lines to suppress warnings given by gcc but then I realized that clang is also used so decided to remove those #pragma directives to avoid making thins overcomplicated.

rigtorp:
> What are the warnings you are referring to and how did you get it to print them?
@kozlovic Check my PR #225. These are new warnings added to recent GCC and clang versions.

kozlovic:
@danesh-d Should be fixed with #239 and prior.

PR #237, opened 2019-07-16
FIXED: build of client if no openssl installed if -DNATS_BUILD_WITH_T…

vsysoev:
…LS=OFF


kozlovic:
@vsysoev Thank you for the contribution!

PR #242, opened 2019-07-30
[ADDED] natsConnection_Sign() method

kozlovic:
Also fixed issue where JWT was not escaped when sent to the server
which could cause connect failures.


README.md
```diff
@@ -555,6 +555,31 @@ if (s == NATS_OK)
...
+```c
+    s = natsOptions_Create(&opts);
+    if (s == NATS_OK)
+        s = natsOptions_SetUserCredentialsFromFiles(opts, "user.creds", NULL);
+    if (s == NATS_OK)
+        s = natsConnection_Connect(&nc, opts);
+
+    // Sign some arbitrary content
+    const unsigned char *content   = (const unsigned char*) "hello";
+    int                 contentLen = 5;
```
derekcollison: Formatting looks off.

README.md
```diff
@@ -555,6 +555,31 @@ if (s == NATS_OK)
...
+    unsigned char		*sig       = NULL;
+    int					sigLen     = 0;
+
+    s = natsConnection_Sign(nc, content, contentLen, &sig, &sigLen);
```
derekcollison: sigLen is fixed yes, 64 bytes?

src/conn.c
```diff
@@ -3797,3 +3914,38 @@ natsConn_signatureHandler(char **customErrTxt, unsigned char **psig, int *sigLen
...
+{
+    natsStatus      s       = NATS_OK;
+    userCreds       *uc     = (userCreds*) closure;
+
+    s = _sign(uc, (const unsigned char*) nonce, 0, psig, sigLen);
+    return NATS_UPDATE_ERR_STACK(s);
+}
+
+natsStatus
+natsConnection_Sign(natsConnection *nc, const unsigned char *payload, int payloadLen, unsigned char **sig, int *sigLen)
```
derekcollison: sigLen should be 64bytes yes? Should we check it?

src/nats.h
```diff
@@ -2403,6 +2403,28 @@ natsConnection_Drain(natsConnection *nc);
...
+ * @param message the byte array to sign.
+ * @param messageLen the length of the given byte array.
+ * @param sig the returned signature.
+ * @param sigLen the length of the returned signature.
+ */
+natsStatus
+natsConnection_Sign(natsConnection *nc,
+                    const unsigned char *message,
+                    int messageLen,
+                    unsigned char **sig,
```
derekcollison: Do they malloc this pre-call or does the call?

README.md
kozlovic: will fix

README.md
kozlovic: Yes, so should I remove that from function and document that returned signature is 64 bytes?

src/conn.c
kozlovic: The function allocates, but then yes, I could remove this param since the returned signature will be 64 bytes.

src/nats.h
kozlovic: The call does allocate (because I do that at low level). So siglen is not required (see above).
I was thinking of having a version that does sign in a buffer, but the low level thing I had done does allocation, so was too much of a change for now.

README.md
derekcollison: I think so.

src/conn.c
derekcollison: Agree, but why do we not pass in addr to a 64 byte char array? That way we can stick it on the stack if we want.

src/nats.h
derekcollison: I don't think we should malloc the space for the sig.

src/conn.c
kozlovic: Because low level crypto_sign() that I use has the signature followed by original input in same memory block. So I need to allocate (see natsKeys_Sign() and alloc of `sm`).
Now of course, I could still have the alloc and then make another copy to give byte array, but that's an extra memcopy..

src/nats.h
kozlovic: See above

src/conn.c
derekcollison: You can allocate but public function should take *char to a 64byte block that we fill in.

kozlovic: will do

PR #250, opened 2019-08-02
[ADDED] Ability to set IO read/write buffer size.

s-shin:
Using this option, we can change the internal temporary buffer size, which is used for read/write and was hard coded as `DEFAULT_BUF_SIZE` in conn.c.
Increase of the size can improve the performance of read/write operations in some heavy load situations.

I would like to change the value in my use case, so could you consider it?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ Compiled Object files, Static and Dynamic libs (Shared Objects)
 
 # Folders
 build*/
+cmake-build*/
 install/
 html/
 !doc/html/
@@ -21,6 +22,7 @@ html/
 .settings/
 
 .vscode/
+.idea/
 
 # Mac
 .DS_Store
--- a/src/conn.c
+++ b/src/conn.c
@@ -34,6 +34,7 @@
 #include "nkeys.h"
 
 #define DEFAULT_SCRATCH_SIZE    (512)
+#define MAX_INFO_MESSAGE_SIZE   (32768)
 
 #define NATS_EVENT_ACTION_ADD       (true)
 #define NATS_EVENT_ACTION_REMOVE    (false)
@@ -477,20 +478,13 @@ static natsStatus
 _readOp(natsConnection *nc, natsControl *control)
 {
     natsStatus  s = NATS_OK;
-    char        *buffer;
-
-    buffer = NATS_MALLOC(nc->opts->ioBufSize);
-    if (buffer == NULL) {
-        return NATS_UPDATE_ERR_STACK(NATS_NO_MEMORY);
-    }
+    char        buffer[MAX_INFO_MESSAGE_SIZE];
 
     buffer[0] = '\0';
 
-    s = natsSock_ReadLine(&(nc->sockCtx), buffer, nc->opts->ioBufSize);
+    s = natsSock_ReadLine(&(nc->sockCtx), buffer, sizeof(buffer));
     if (s == NATS_OK)
         s = nats_ParseControl(control, buffer);
-
-    NATS_FREE(buffer);
 
     return NATS_UPDATE_ERR_STACK(s);
 }
--- a/src/conn.c
+++ b/src/conn.c
@@ -2093,6 +2093,8 @@ _readLoop(void  *arg)
         natsConn_Lock(nc);
     }
 
+    NATS_FREE(buffer);
+
     natsSock_Close(fd);
     nc->sockCtx.fd       = NATS_SOCK_INVALID;
     nc->sockCtx.fdActive = false;
--- a/src/conn.c
+++ b/src/conn.c
@@ -34,8 +34,6 @@
 #include "nkeys.h"
 
 #define DEFAULT_SCRATCH_SIZE    (512)
-#define DEFAULT_BUF_SIZE        (32768)
-#define DEFAULT_PENDING_SIZE    (1024 * 1024)
 
 #define NATS_EVENT_ACTION_ADD       (true)
 #define NATS_EVENT_ACTION_REMOVE    (false)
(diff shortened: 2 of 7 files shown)


src/conn.c
```diff
@@ -34,8 +34,6 @@
 #include "nkeys.h"
 
 #define DEFAULT_SCRATCH_SIZE    (512)
-#define DEFAULT_BUF_SIZE        (32768)
-#define DEFAULT_PENDING_SIZE    (1024 * 1024)
```
s-shin: Remove this unused definition as well.

.gitignore
```diff
@@ -6,6 +6,7 @@ Compiled Object files, Static and Dynamic libs (Shared Objects)
```
s-shin: Ignore CLion files.

kozlovic:
@s-shin I like that idea. Will comment as soon as I can. As you can see there are some reported leaks since I presume the buffer is not freed at en of readLoop. We need to check also the possible value provided by user and consequences. Again, will have more thorough review later.

s-shin:
> As you can see there are some reported leaks since I presume the buffer is not freed at en of readLoop.

Exactly, this is just my mistake, and fixed in https://github.com/nats-io/nats.c/pull/250/commits/e2555a21893949de07db56e8721d4984e8a2d62a.

src/nats.h
```diff
@@ -1045,6 +1045,20 @@ natsOptions_SetPingInterval(natsOptions *opts, int64_t interval);
 NATS_EXTERN natsStatus
 natsOptions_SetMaxPingsOut(natsOptions *opts, int maxPingsOut);
 
+/** \brief Sets the size of the internal read buffer.
+ *
+ * Sets the size, in bytes, of the internal read buffer used for read data from
```
kozlovic: Maybe `used for reading data from` or `used to read data from`?

src/opts.c
```diff
@@ -597,6 +597,17 @@ natsOptions_SetMaxPingsOut(natsOptions *opts, int maxPignsOut)
     return NATS_OK;
 }
 
+natsStatus
+natsOptions_SetIOBufSize(natsOptions *opts, int ioBufSize)
+{
+  LOCK_AND_CHECK_OPTIONS(opts, (ioBufSize < 0));
+
+  opts->ioBufSize = ioBufSize;
```
kozlovic: I think you need to check for ioBufSize being 0 here, and if so, use the default value. The default value is set when creating an option object, but if user calls this with SetIOBufSize(opts, 0), then I think that would set it to zero. (see comment about the test, which I think if you write first should fail).

test/test.c
```diff
@@ -2390,6 +2391,10 @@ test_natsOptions(void)
         s = natsOptions_SetMaxPingsOut(opts, 10);
     testCond((s == NATS_OK) && (opts->maxPingsOut == 10));
 
+    test("Set IOBufSize: ");
+    s = natsOptions_SetIOBufSize(opts, 1024 * 1024);
+    testCond((s == NATS_OK) && (opts->ioBufSize == 1024 * 1024));
+
```
kozlovic: Here, you could have a test that checks negative value (and expect s == NATS_INVAID_ARG) and then set to 0 and expect the size to be set to NATS_OPTS_DEFAULT_IO_BUF_SIZE.

src/conn.c
```diff
@@ -2044,19 +2049,29 @@ static void
...
     natsSock    fd;
     int         n;
 
     natsConnection *nc = (natsConnection*) arg;
 
     natsConn_Lock(nc);
 
+    fd = nc->sockCtx.fd;
+
+    buffer = NATS_MALLOC(nc->opts->ioBufSize);
```
kozlovic: Maybe we could capture the option value in a local variable here so that we don't need to de-reference nc->opts in the natsSock_Read() call down there.

src/nats.h
s-shin: Correct, but I get the feeling that it would be better to describe write things.
How about this?

```
/** \brief Sets the size of the internal read/write buffers.
 *
 * Sets the size, in bytes, of the internal read/write buffers used for reading/writing data from
 * a socket.
 * If not specified, or the value is 0, the library will use a default value,
 * currently set to 32KB.
 *
 * @param opts the pointer to the #natsOptions object.
 * @param ioBufSize the size, in bytes, of the internal buffer for read/write
 * operations.
 */
```

src/opts.c
s-shin: >  if user calls this with SetIOBufSize(opts, 0), then I think that would set it to zero.

This is no problem because when `opts->ioBufSize` is set to zero, the default value will be set in `natsConn_create`.
https://github.com/nats-io/nats.c/pull/250/files#diff-152747a2399765ea9211ffbfd6ba74edR2978

Some other options (e.g. maxPingsOut) are implemented with this pattern as well.

(I also feel the pattern is a bit potentially dangerous though.)

s-shin: By the way, I will fix the indent here.

src/nats.h
s-shin: See 30984fd.

src/opts.c
s-shin: See 643cab1.

test/test.c
s-shin: See d086d8d.

src/conn.c
s-shin: See 9c8ad96.

src/conn.c
```diff
@@ -479,14 +477,21 @@ static natsStatus
 _readOp(natsConnection *nc, natsControl *control)
 {
     natsStatus  s = NATS_OK;
-    char        buffer[DEFAULT_BUF_SIZE];
+    char        *buffer;
```
kozlovic: Sorry that I did not spot that earlier: please revert changes to this function. This is used to read the INFO protocol from the server and we cannot be limited by a user setting of ioBufSize (this would cause the connection to fail).

s-shin: OK, I will revert the part of these changes.
I'm also thinking to make the name more meaningful such as `MAX_INFO_MESSAGE_SIZE`.

I just have one question: the size should be 32768, or can we shrink it?

kozlovic: Leave it at 32KB for now. The INFO from server is getting bigger and bigger. Is that size a problem in the context you are planning on using the library?

s-shin: > Is that size a problem in the context you are planning on using the library?

No problem at all! 

Now I fixed this at https://github.com/nats-io/nats.c/pull/250/commits/12c1efe8fcdb3f0c81036ea56048944f7985356b.

PR #253, opened 2019-08-08
[ADDED] natsMsg_Respond() API

### LINKED ISSUE #252: Add Respond() API
Similar to Go client:
https://github.com/nats-io/nats.go/pull/489
https://github.com/nats-io/nats.go/pull/472

kozlovic:
With the C client, this is not as convenient as it is for the Go client since the C message handler already passes the connection, so it is easy to use it to publish a response.
That being said, adding this new API to be closer to the Go client feature set.

Resolves #252


kozlovic:
@derekcollison I really question the usefulness of this for the C client since the message callback already has the `natsConnection` passed, but..

derekcollison:
No need then for C client. Your call.


________________________________

kozlovic:
Ok, then I will close and not add this API for now.

PR #268, opened 2019-09-25
[ADDED] natsOptions_SetWriteDeadline()

### LINKED ISSUE #215: Deadlock when reconnecting
After network glitches NATS fails to reconnect due to a deadlock. It seems that the flusher thread does a blocking write (!?😕) while the other threads are waiting for it. This deadlocks the NATS library.

```
Thread 4 (Thread 0x7f55813ff700 (LWP 97656)):
#0  0x00007f566520f4ed in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f5665211d12 in _L_cond_lock_938 () from /lib64/libpthread.so.0
#2  0x00007f5665211bdb in __pthread_mutex_cond_lock () from /lib64/libpthread.so.0
#3  0x00007f566520cde4 in pthread_cond_timedwait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#4  0x00000000008b53fc in natsCondition_AbsoluteTimedWait ()
#5  0x000000000089fc6c in _flushTimeout ()
#6  0x00000000008a4592 in natsConnection_Flush ()
#7  0x00000000008a4c2e in _doReconnect ()
#8  0x00000000008b5a7f in _threadStart ()
#9  0x00007f5665208dd5 in start_thread () from /lib64/libpthread.so.0
#10 0x00007f56630c3ead in clone () from /lib64/libc.so.6
Thread 3 (Thread 0x7f56247e0700 (LWP 97657)):
#0  0x00007f566520f4ed in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f5665211d12 in _L_cond_lock_938 () from /lib64/libpthread.so.0
#2  0x00007f5665211bdb in __pthread_mutex_cond_lock () from /lib64/libpthread.so.0
#3  0x00007f566520cde4 in pthread_cond_timedwait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#4  0x00000000008b53fc in natsCondition_AbsoluteTimedWait ()
#5  0x000000000089fc6c in _flushTimeout ()
#6  0x00000000008a030d in _close ()
#7  0x00000000008a2979 in natsConn_processErr ()
#8  0x00000000008ab3e6 in natsParser_Parse ()
#9  0x00000000008a0599 in _readLoop ()
#10 0x00000000008b5a7f in _threadStart ()
#11 0x00007f5665208dd5 in start_thread () from /lib64/libpthread.so.0
#12 0x00007f56630c3ead in clone () from /lib64/libc.so.6
Thread 2 (Thread 0x7f5624fe1700 (LWP 97658)):
#0  0x00007f566520fbfb in send () from /lib64/libpthread.so.0
#1  0x00007f5665c44fe9 in onload_send (fd=21, msg=0x7f559f2c7380, len=32998, flags=0) at ../../../../../src/lib/transport/unix/sockcall_intercept.c:944
#2  0x00000000008b380a in natsSock_WriteFully ()
#3  0x000000000089f69c in natsConn_bufferFlush ()
#4  0x000000000089f78b in _flusher ()
#5  0x00000000008b5a7f in _threadStart ()
#6  0x00007f5665208dd5 in start_thread () from /lib64/libpthread.so.0
#7  0x00007f56630c3ead in clone () from /lib64/libc.so.6
```

@altmind

### LINKED ISSUE #217: nats.c is using blocking sockets and select with infinite timeouts
See #215 for the concrete bug. This issue is a discussion on how to handle use sockets with unreliable networks in general.

I see in `natsSock_ConnectTcp` care was taken to handle connection timeout correctly by setting socket to non-blocking and then using `select` to wait with a timeout. That is all good but it's possible to block indefinitely by mistake if `ctx->deadline` was not set before calling `natsSock_ConnectTcp`; in which case `natsSock_WaitReady` will block forever. It would be a good idea to make it impossible to introduce a bug like that.

Later in `_processConnInit` the socket is set back to blocking mode again. I think that is misguided it allows future read and write operations to block. Instead `select` (or equivalent) should be used to make sure all operations will timeout.

kozlovic:
Adds a write deadline for all socket write operations. It means that
a publish call could fail (with NATS_TIMEOUT) if the server is not
able to consume data (either not fast enough or connection is in
a state where it is not reported as closed, but transmission is
not happening).

Resolves #217
Resolves #215

### DIFF
--- a/src/comsock.c
+++ b/src/comsock.c
@@ -1,4 +1,4 @@
-// Copyright 2015-2018 The NATS Authors
+// Copyright 2015-2019 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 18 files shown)


src/comsock.c
```diff
@@ -439,9 +440,11 @@ natsStatus
 natsSock_WriteFully(natsSockCtx *ctx, const char *data, int len)
 {
     natsStatus  s     = NATS_OK;
-    int         bytes = 0;
     int         n     = 0;
 
+    if (len == 0)
```
ColinSullivan1: How did a natsSock_Write of 0 behave previously?  return 0?

kozlovic: Fairly confident that I was never calling with len==0, but otherwise, connection would probably have been reported as closed.

PR #269, opened 2019-09-27
[ADDED] Options to load CA and client cert/key from memory.

### LINKED ISSUE #267: TLS  - Using cert content instead of file location
Hi All,

**This is not an issue. I am just trying to see if i there is any other approaches for TLS connection**

I am trying to use NATS C library to connect to a NATS cluster via TLS . I am able to successfully connect it using following method

```c
natsOptions_LoadCertificatesChain(opts,certChainFilePath,keyFilePath);
```
I want to know if there is anyway i can pass the cert and key file content as character array rather than passing the file path which is file system path. Reason being i am looking for such option is that for security reason i cant simply have the cert and key file on the host file system.

My plan is to have the cert and key file in some db(sqllite) and read the content from DB and pass it to natsOptions_LoadCertificatesChain() method.

Any help or suggestions ?

Thanks,
Arajit

kozlovic:
- natsOptions_SetCertificatesChain()
- natsOptions_SetCATrustedCertificates()

These options allow to pass the certificate and key from a memory
buffer as opposed to read it from a file.

Resolves #267

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -15868,7 +15868,202 @@ test_SSLVerify(void)
 
     nats_Sleep(100);
 
-    serverPid = _startServer("nats://127.0.0.1:4443", "-config tls.conf", true);
+    serverPid = _startServer("nats://127.0.0.1:4443", "-config tlsverify.conf", true);
+    CHECK_SERVER_STARTED(serverPid);
+
+    natsMutex_Lock(args.m);
+    while ((s == NATS_OK) && !(args.reconnected))
+        s = natsCondition_TimedWait(args.c, args.m, 2000);
+    natsMutex_Unlock(args.m);
+
+    if (s == NATS_OK)
+        s = natsConnection_PublishString(nc, "foo", "test");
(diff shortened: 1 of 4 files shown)


test/test.c
```diff
@@ -15868,7 +15868,202 @@ test_SSLVerify(void)
...
+    s = natsOptions_SetReconnectedCB(opts, _reconnectedCb, &args);
+    if (s == NATS_OK)
+        s = natsOptions_SetURL(opts, "nats://localhost:4443");
+    if (s == NATS_OK)
+        s = natsOptions_SetSecure(opts, true);
+    // For test purposes, we provide the CA trusted certs
+    if (s == NATS_OK)
+    {
+        s = natsOptions_SetCATrustedCertificates(opts, (const char*) natsBuf_Data(certBuf));
+        // Demonstrate that we can free the memory after this call..
```
sasbury: This comment was very helpful to me, i was about to comment and then realized i didn't have to and am anyway to say I like the comment

PR #276, opened 2019-10-24
Reset buffers when connecting. Fixes #275

### LINKED ISSUE #275: Not reconnecting after error
Using master from a week ago. Allowing infinite number of reconnects.

After nats.c was disconnected and then successfully reconnected, EPROTO error occurred somewhere in nats.c and caused it to stop reconnecting. I expect nats.c to try and reconnect again.

[2019-10-23, 20:05:35.817141] <warning> NATS: state changed to disconnected:
[2019-10-23, 20:05:35.967504] <debug> NATS state: CONNECTED
[2019-10-23, 20:05:35.967506] <debug> NATS stats: inMsgs=48578, outMsgs=5596336, inBytes=32383, outBytes=348927432, reconnects=1

[2019-10-23, 20:05:35.967507] <info> NATS: state changed to reconnected
[2019-10-23, 20:05:35.967508] <debug> NATS state: CONNECTED
[2019-10-23, 20:05:35.967508] <debug> NATS stats: inMsgs=48578, outMsgs=5596336, inBytes=32383, outBytes=348927432, reconnects=1

[2019-10-23, 20:05:35.967667] <warning> NATS: state changed to disconnected: Unknown Protocol Operation
[2019-10-23, 20:05:35.967675] <debug> NATS state: CLOSED
[2019-10-23, 20:05:35.967675] <debug> NATS stats: inMsgs=48578, outMsgs=5596336, inBytes=32383, outBytes=348927432, reconnects=1

[2019-10-23, 20:05:35.967676] <error> NATS: state changed to closed: Unknown Protocol Operation
[2019-10-23, 20:05:35.967678] <debug> NATS state: CLOSED
[2019-10-23, 20:05:35.967679] <debug> NATS stats: inMsgs=48578, outMsgs=5596336, inBytes=32383, outBytes=348927432, reconnects=1

rigtorp:
Reset buffers when connecting. This prevents sending of partial message frames.

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2661,11 +2661,7 @@ natsConn_processErr(natsConnection *nc, char *buf, int bufLen)
     }
     else
     {
-        close = true;
-        natsConn_Lock(nc);
-        nc->err = NATS_ERR;
-        snprintf(nc->errStr, sizeof(nc->errStr), "%s", error);
-        natsConn_Unlock(nc);
+        _processOpError(nc, NATS_ERR, false);
     }
     if (close)
         _close(nc, NATS_CONN_STATUS_CLOSED, false, true);
--- a/src/conn.c
+++ b/src/conn.c
@@ -398,13 +398,11 @@ _createConn(natsConnection *nc)
     {
         nc->sockCtx.fdActive = true;
 
-        if ((nc->pending != NULL) && (nc->bw != NULL)
-            && (natsBuf_Len(nc->bw) > 0))
-        {
-            // Move to pending buffer
-            s = natsConn_bufferWrite(nc, natsBuf_Data(nc->bw),
-                                     natsBuf_Len(nc->bw));
-        }
+        if (nc->pending != NULL)
+            natsBuf_Reset(nc->pending);
+
+        if (nc->bw != NULL)
+            natsBuf_Reset(nc->bw);
     }
 
     // Need to create the buffer even on failure in case we allow


src/conn.c
```diff
@@ -398,13 +398,11 @@ _createConn(natsConnection *nc)
...
         nc->sockCtx.fdActive = true;
 
-        if ((nc->pending != NULL) && (nc->bw != NULL)
-            && (natsBuf_Len(nc->bw) > 0))
-        {
-            // Move to pending buffer
-            s = natsConn_bufferWrite(nc, natsBuf_Data(nc->bw),
-                                     natsBuf_Len(nc->bw));
-        }
+        if (nc->pending != NULL)
```
kozlovic: No, this is missing the point of having pending buffers. The idea of this buffer is to collect all activity from the user into a buffer so that on reconnect data can be sent to the server. What needs to be done is avoid sending/storing partials in the first place.

rigtorp: Is this a quick fix for my current issue? At least it would be a temporary fix for me. I don't assume any delivery guarantees, so my system will recover. I thought non streaming NATS doesn't make any guarantees about message delivery so why all this business trying to resend pending messages? The primary goal is to have correct behavior under all error conditions, this pending stuff is only a optimization to use if it's implemented correctly.

There are better ways to handle pending messages. One would be to have a pool of message objects that are enqueued on a lock free queue to be sent. Only partially sent messages, inflight and TCP buffered messages would be lost. It would also avoid tripple buffering like how is done now. Another solution would be to store a queue of offsets in the pending buffer to message frame splits and discard the first partially sent message if any.

kozlovic: Yes, I think that would be a quick fix for the issue at hand. However, I cannot merge the PR as-is, but I will try to address the issue as soon as possible. So in the meantime, please work from your fork. Thanks again!

rigtorp: This fix did not help. I still get "Unknown protocol error" on reconnect. I think the parser state also needs to be reset.

[2019-11-01, 04:52:16.327355] <warning> NATS: state changed to disconnected: 
[2019-11-01, 04:52:16.327363] <debug> NATS state: CONNECTED
[2019-11-01, 04:52:16.327364] <debug> NATS stats: inMsgs=164552, outMsgs=19572204, inBytes=131162, outBytes=2047780352, reconnects=2

[2019-11-01, 04:52:16.327365] <info> NATS: state changed to reconnected
[2019-11-01, 04:52:16.327365] <debug> NATS state: CONNECTED
[2019-11-01, 04:52:16.327366] <debug> NATS stats: inMsgs=164552, outMsgs=19572204, inBytes=131162, outBytes=2047780352, reconnects=2

[2019-11-01, 04:52:16.346722] <warning> NATS: state changed to disconnected: Unknown Protocol Operation
[2019-11-01, 04:52:16.346729] <debug> NATS state: CLOSED
[2019-11-01, 04:52:16.346730] <debug> NATS stats: inMsgs=164552, outMsgs=19572204, inBytes=131162, outBytes=2047780352, reconnects=2

[2019-11-01, 04:52:16.346730] <error> NATS: state changed to closed: Unknown Protocol Operation
[2019-11-01, 04:52:16.346732] <debug> NATS state: CLOSED
[2019-11-01, 04:52:16.346732] <debug> NATS stats: inMsgs=164552, outMsgs=19572204, inBytes=131162, outBytes=2047780352, reconnects=2

kozlovic: The parser is destroyed at the end of the existing readloop and recreated when the new one start. Furthermore, the parser is used to read data, not to send, and if you are getting "Unknown Protocol Operation" it means that something was sent to the server that does not understand it, and returns this -ERR. 
So with the sequence above, it looks like it is during the reconnect that something bad was sent to the server, which then returned the -ERR that causes the close.

kozlovic:
I am truly sorry that I don't have the time at the moment to fix the underlying issue, so I will ask that you work on your own branch for now as I can't merge this PR. 
Your current approach *may* work for your specific issue, but understand that if the application were to do something that the server does not like, the client will keep reconnecting and again cause the server to send back the error, etc.. My goal, as soon as I find the time, will be to figure out where the partial comes from (I am surprised that you still got it with a reset of the pending).

kozlovic:
@rigtorp Check if #280 helps. Thanks!

kozlovic:
Closing this in favor of #280.

PR #277, opened 2019-10-24
Use MSG_NOSIGNAL and SO_NOSIGPIPE to suppress SIGPIPE signals

rigtorp:


rigtorp:
It will work for OpenBSD too. Some BSD needs SO_NOSIGPIPE. I can add that.

kozlovic:
On macOS:
```
Scanning dependencies of target nats
[  1%] Building C object src/CMakeFiles/nats.dir/comsock.c.o
/Users/ivan/dev/cnats/src/comsock.c:379:46: error: use of undeclared identifier 'MSG_NOSIGNAL'
            bytes = send(ctx->fd, data, len, MSG_NOSIGNAL);
                                             ^
1 error generated.
make[2]: *** [src/CMakeFiles/nats.dir/comsock.c.o] Error 1
make[1]: *** [src/CMakeFiles/nats.dir/all] Error 2
make: *** [all] Error 2
```
did not try Windows yet..

rigtorp:
New patch will work on macOs, BSD and Linux. You should setup CI for macOS and windows.

PR #303, opened 2020-03-23
[ADDED] natsSubscription_SetOnComplete API

### LINKED ISSUE #302: Add callback to be invoked when subscription is closed and handler returns
This will allow user to safely free resources that are used in the message handler.

Relates to #299 

/cc @pananton

kozlovic:
This allows users to set a callback that will be invoked when the
subscription is closed and message handler returns.

Resolves #302

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -11986,6 +11986,128 @@ test_NextMsgCallOnAsyncSub(void)
     _stopServer(serverPid);
 }
 
+static void
+testSubOnComplete(void *closure)
+{
+    struct threadArg *arg = (struct threadArg*) closure;
+
+    natsMutex_Lock(arg->m);
+    arg->status = (arg->control == 2 ? NATS_OK : NATS_ERR);
+    arg->done = true;
+    natsCondition_Signal(arg->c);
+    natsMutex_Unlock(arg->m);
+}
+
+static void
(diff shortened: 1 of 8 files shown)


test/test.c
```diff
@@ -11986,6 +11986,128 @@ test_NextMsgCallOnAsyncSub(void)
...
+    if (s == NATS_OK)
+    {
+        // Wait for message handler to be invoked
+        natsMutex_Lock(arg.m);
+        while ((s != NATS_TIMEOUT) && (arg.control != 1))
+            s = natsCondition_TimedWait(arg.c, arg.m, 2000);
+        natsMutex_Unlock(arg.m);
+    }
+    if (s == NATS_OK)
+        s = natsSubscription_Unsubscribe(sub);
```
matthiashanel: if you move if s == NATS_OK and natsSubscription_Unsubscribe(sub) into the lock, the callback won't need a sleep

kozlovic: The sleep is on purpose. I want the message handler to be in progress while I call Unsubscribe(). The idea is that this test will then make sure that the OnCompleteCB is invoked only after the message handler returns. I do so by setting something at the end of the message handler that the OnCompleteCB check as proof that it was invoked after it.

matthiashanel: yeah, never mind, you would need another condition and that is probably not worth it. LGTM

PR #312, opened 2020-04-23
[ADDED] natsConnection_GetClientIP()

kozlovic:
Returns the IP as determined by the server when accepting this
connection.

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -168,6 +168,7 @@ _clearServerInfo(natsServerInfo *si)
     NATS_FREE(si->connectURLs);
 
     NATS_FREE(si->nonce);
+    NATS_FREE(si->clientIP);
 
     memset(si, 0, sizeof(natsServerInfo));
 }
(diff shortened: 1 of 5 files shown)


src/conn.c
```diff
@@ -525,6 +526,8 @@ _processInfo(natsConnection *nc, char *info, int len)
         s = nats_JSONGetULong(json, "client_id", &(nc->info.CID));
     if (s == NATS_OK)
         s = nats_JSONGetStr(json, "nonce", &(nc->info.nonce));
+    if (s == NATS_OK)
+        s = nats_JSONGetStr(json, "client_ip", &(nc->info.clientIP));
 
 #if 0
```
derekcollison, line 532: You leave these in?

kozlovic: Sorry, not sure I understand the question

kozlovic: The #if 0 you mean?

derekcollison: Yes, common to leave those in for folks to easily debug stuff?

derekcollison: Been so long since I have done C ;)

kozlovic: That was on old vestige of debugging the parsing of the INFO. I will remove.

PR #319, opened 2020-05-06
[FIXED] Randomize discovered server URLs if allowed

kozlovic:
If randomize is allowed, discovered URLs will be randomly added
to the pool of urls instead of always be appended.

See nats-io/nats.go#565


kozlovic:
Need to update based on changes made to the Go client. Namely, need to ensure that we don't shuffle the whole pool but leave the first URL untouched.

PR #353, opened 2020-07-28
[ADDED] Support for no responders

kozlovic:
If connected to a NATS Server v2.2.0+ and a request is made while
there is no responder connected, the request will fail right away
instead of waiting for the given timeout interval.
This requires headers support.


src/pub.c
```diff
@@ -344,7 +374,14 @@ _oldRequestMsg(natsMsg **replyMsg, natsConnection *nc,
         s = natsConn_publish(nc, requestMsg, true);
     }
     if (s == NATS_OK)
+    {
         s = natsSubscription_NextMsg(replyMsg, sub, timeout);
+        // For servers that support it, we may receive an empty message
+        // with a 503 status header. If that is the case, return NULL
+        // message and NATS_TIMEOUT error.
```
derekcollison, line 381: Do we return timeout error in Go client or new type of client error for no responders?

kozlovic: In Go we return a new error:
```
ErrNoResponders           = errors.New("nats: no responders available for request")
```
but see my note [here](https://github.com/nats-io/nats.c/pull/353/files#diff-dd79a06628a6e43bbf6cfeac0d4d06ffR2926). I feel like a new status could break some applications. They could code such as if NATS_TIMEOUT then try again or assume there is no response, but if any other error code, something is really wrong and behave differently.

derekcollison: Hmm, I can see both cases. Still feel new functionality and should be a new error code. Its not really a timeout.

kozlovic: I would not have hesitated adding a new status error code if the feature was driven by an option, that is, CONNECT would send "no_responders" only if an option is set prior to create the connection, because then that means users would code knowing that this would possibly trigger say "NATS_NO_RESPONDERS" error code.
With this drop-in replacement where it is automatic as long as the app connects to a server v2.2.0, I think it is dangerous.

derekcollison: Well we should be consistent as we can across clients. I don't like when they drift apart too much behavior wise. So we need to keep working on a common goal.

When I did the Go client I remember us talking a bit and saying we will take a wait and see approach on if anyone has issues with this and runs into it etc.

kozlovic: Not too late to change in the Go client since it has not been released yet :-)
I would think that the "wait and see" approach would still be safer with using the timeout error than adding a new one. Maybe we should ask the community and the rest of the dev team to have their opinion?

derekcollison: Obviously we disagree about this one, I feel it definitely is not a timeout error and new applications should look to handle this error if they chose in some way like timeout. Most code I have seen does not distinguish, they just noticed that was an error and they did not get their response.

kozlovic: So, thinking more about it, yes, there could be an advantage of knowing that there is no responder vs the responder took too long to respond. But still, returning another status for applications that did not expect it, will break them. Then you said:

> Most code I have seen does not distinguish, they just noticed that was an error and they did not get their response.

Then it would not matter which status were return if nobody checks anyway.

But I disagree. For instance as soon as NATS Streaming server's dependencies are going to be upgraded to a NATS library that has the no responder feature (and connect to a NATS Server), the code will break:
https://github.com/nats-io/nats-streaming-server/blob/026e3a6ace642e09ea1bcb34440e5eefd692ecc7/server/partitions.go#L173
https://github.com/nats-io/nats-streaming-server/blob/026e3a6ace642e09ea1bcb34440e5eefd692ecc7/server/server.go#L2394

Now *I* know and I can fix that when upgrading NATS Streaming deps, but I am just saying that it may break bunch of user apps since this is not an opt-in feature. Maybe the solution is to make it so.

kozlovic: Actually agree that we should have a different status, because then you could write applications that if sending a request and get no responder can behave a certain way as opposed to not knowing if the request was received or not (due to disconnect somewhere) or the responder did not have enough time.
So again, I would really like this to be opt-in so that developers can code accordingly. But for now, if you prefer I can simply add the new status and maybe fix some of the tests that are expecting NATS_TIMEOUT instead of the new error code :-).

derekcollison: Could we approach like others in the past where we default to new behavior but if that causes issues the app developer can revert back to the old behavior?

kozlovic: Done! https://github.com/nats-io/nats.c/pull/353/commits/34916e21ed1a8e75f49e1a5ecdbfe74f21c29940

PR #361, opened 2020-09-15
Fix identical inner and outer 'if' condition

paoloteti:
This patch fixes the following `cppckeck` warnings:

```
msg.c:842:18: note: outer condition: replyLen>0
    if (replyLen > 0)
                 ^
msg.c:845:22: note: identical inner condition: replyLen>0
        if (replyLen > 0)
```

and it is obvious that `memcpy()` length parameter is already guarded by the outer condition.


kozlovic (approved):
LGTM. Thank you for the contribution!

PR #362, opened 2020-09-16
Fix possible memory leak due to an early return

paoloteti:
Cppcheck report a possible memory leak due to an early return after a memory allocation.

```
sub.c:350:9: error: Memory leak: sub [memleak]
        return nats_setError(NATS_INVALID_ARG, "Invalid bytes limit of %d", sub->bytesLimit);
```

Even if it cannot be triggered if options are setups using `natsOptions_Set*` helpers, the easy way to formally fix the issue is to validate the limit threshold (in bytes) before any memory allocation.


kozlovic (approved):
LGTM. Thanks again!

PR #363, opened 2020-09-20
[FIXED] Possible memory leak when mem alloc failed creating an URL

paoloteti:
GCC 10.0.1, when used with with `-fanalyzer`, reports:

`warning: leak of ‘url’ [CWE-401] [-Wanalyzer-malloc-leak]`

Issue is due to an early return. The easy way to fix it, without changing
too much, is to propagate the error code. In this way the function will
take care to free data structure calling `natsUrl_Destroy(url)`.

### DIFF
--- a/src/url.c
+++ b/src/url.c
@@ -179,14 +179,14 @@ natsUrl_Create(natsUrl **newUrl, const char *urlStr)
                 *pwd = '\0';
 
                 if (url->password == NULL)
-                    return nats_setDefaultError(NATS_NO_MEMORY);
+                    s = nats_setDefaultError(NATS_NO_MEMORY);
             }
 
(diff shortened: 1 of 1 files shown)


src/url.c
```diff
@@ -179,14 +179,14 @@ natsUrl_Create(natsUrl **newUrl, const char *urlStr)
                 *pwd = '\0';
 
                 if (url->password == NULL)
-                    return nats_setDefaultError(NATS_NO_MEMORY);
+                    s = nats_setDefaultError(NATS_NO_MEMORY);
```
kozlovic, line 182: Good catch. I broke that when introducing the stack trace updates I guess..

kozlovic (approved):
LGTM

PR #364, opened 2020-09-20
Fix GCC 10.x warning: on output truncated before terminating with nul

paoloteti:
GCC 10.x (also on TravisCI) raise the following warning:

```
In file included from /usr/include/string.h:494,
             from /home/travis/build/nats-io/nats.c/src/include/n-unix.h:37,
             from /home/travis/build/nats-io/nats.c/src/natsp.h:20,
             from /home/travis/build/nats-io/nats.c/src/msg.c:14:

In function ‘strncpy’,
inlined from ‘_canonicalKey.constprop’ at /home/travis/build/nats-io/nats.c/src/msg.c:217:9:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:106:10: warning: ‘__builtin_strncpy’ output
truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
  |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

/home/travis/build/nats-io/nats.c/src/msg.c: In function ‘_canonicalKey.constprop’:
/home/travis/build/nats-io/nats.c/src/msg.c:181:33: note: length computed here
181 |     int         lenKey  = (int) strlen(key);
```

This is due to the string been truncated after use strncpy().
Even if there isn't an issue in this specific case, this patch replace strncpy() and the explicit NUL termination with strlcpy() that is the preferred alternative to strncpy() that has a incomplete termination semantics.

strncpy() is banned in a lots of open-source code nowadays (git is an example)

By the way strlcpy() is not available in all platforms (it is part of muslc, but not glibc), so this patch add also a way to auto-detect if available and adds a private implementation if not.

### DIFF
--- a/src/msg.c
+++ b/src/msg.c
@@ -214,8 +214,7 @@ _canonicalKey(const char *key, char *keyStack, int keyStackLen, char **cKey, boo
     }
     else
     {
-        strncpy(keyStack, key, lenKey);
-        keyStack[lenKey] = '\0';
+        strlcpy(keyStack, key, lenKey + 1);
         k = keyStack;
     }
     // Move directly to character that we know needs change.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,9 @@ cmake_minimum_required(VERSION 2.8)
 project(cnats)
 include(CTest)
 include(FindPackageHandleStandardArgs)
+include(CheckFunctionExists)
+
+check_function_exists(strlcpy HAVE_STRLCPY)
 
 # Uncomment to have the build process verbose
 #set(CMAKE_VERBOSE_MAKEFILE TRUE)
--- a/src/natsp.h
+++ b/src/natsp.h
@@ -33,6 +33,10 @@
 #define NO_SSL_ERR  "The library was built without SSL support!"
 #endif
 
+#if !defined(HAVE_STRLCPY)
+size_t strlcpy(char *dst, const char *src, size_t siz);
+#endif
+
 #include "err.h"
 #include "nats.h"
 #include "buf.h"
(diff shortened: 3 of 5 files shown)


src/msg.c
```diff
@@ -214,8 +214,7 @@ _canonicalKey(const char *key, char *keyStack, int keyStackLen, char **cKey, boo
     }
     else
     {
-        strncpy(keyStack, key, lenKey);
-        keyStack[lenKey] = '\0';
+        strlcpy(keyStack, key, lenKey + 1);
```
kozlovic, line 217: First, let me say that I really appreciate your recent PRs since they improve the quality of the library.

For that one, however, I think that we can easily solve this without introducing all these changes.
The code here knows that the caller has a byte array that is big enough to contain the copy (+ terminating '\0'), otherwise, it would have gone to the preceding "if" statement and make a copy.

The compiler does not know that (unlike for the tests where the char [] is defined before the strncpy and compiler can "see" that the string copy clearly fit in the char [], hence you don't get the warnings for `test.c`).

So we could simply replace `strncpy` here with:
```
        memcpy((void*) keyStack, (const void*) key, (size_t) lenKey);
        keyStack[lenKey] = '\0';
        k = keyStack;
```

kozlovic (changes requested):
If you agree with my comment, you could do the change or I can close this PR and make the change myself.

paoloteti:
Yes, I agree that memcpy is enougth also because there is only one real occurrence of strncpy().
Yes close the PR and add a memcpy(). I'm answering from my phone and I don't have access to my laptop now. 

Thanks.

kozlovic:
Ok thanks! Closing in favor of #365

PR #365, opened 2020-09-21
Replaced strncpy with memcpy

kozlovic:
Some gcc compiler would report that strncpy could truncate. In
this specific instance, we know that it would not be happening,
but to remove the warning, replaced it with memcpy.

Related to #364


kozlovic:
@paoloteti This is the PR to replace strncpy with memcpy.

paoloteti:
LGTM

PR #367, opened 2020-10-03
[FIXED]: check for CLOCK_MONOTONIC, but use of CLOCK_REALTIME

paoloteti:
Both `nats_Now()` and `nats_NowInNanoSeconds()` check for `CLOCK_MONOTONIC`, but they use `clock_gettime()` with `CLOCK_REALTIME`.

### DIFF
--- a/src/natstime.c
+++ b/src/natstime.c
@@ -25,7 +25,7 @@ nats_Now(void)
     struct _timeb now;
     _ftime_s(&now);
     return (((int64_t)now.time) * 1000 + now.millitm);
-#elif defined CLOCK_MONOTONIC
+#elif defined CLOCK_REALTIME
     struct timespec ts;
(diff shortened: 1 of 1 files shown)


src/natstime.c
```diff
@@ -25,7 +25,7 @@ nats_Now(void)
     struct _timeb now;
     _ftime_s(&now);
     return (((int64_t)now.time) * 1000 + now.millitm);
-#elif defined CLOCK_MONOTONIC
+#elif defined CLOCK_REALTIME
```
kozlovic, line 28: Indeed. From PR: https://github.com/nats-io/nats.c/pull/63, I have this comment:
```
To be fair, this whole thing need revisiting because I used 
defined CLOCK_MONOTONIC but then only use CLOCK_REALTIME.
```

kozlovic (approved):
LGTM, thanks again for your contribution!

PR #372, opened 2020-10-14
[FIXED] Streaming: pings may not report failures

kozlovic:
If client is connected to a NATS server and the streaming server
is down, then the "no responder" feature would cause the server
to send an empty message in response to the PING request (which
was sent with PublishRequest()).
If message is empty, look for "no responder" status.


derekcollison (approved):
LGTM

PR #375, opened 2020-10-19
[FIXED] EventLoop: possible file descriptor leak

### LINKED ISSUE #374: socket fd leak when SSL handshake failed
nats.c: v2.2.0 release
libevent: v2.1.8 release

Socket fd may not be released when using event framework (eg. libevent) if any failure occurred in the time period after the socket is created and before read loop thread is created

In my case for example is SSL handshake failure, psuedo code basically looks like this:

```
while not connected {
    prepare natsOptions (eg. natsOptions_Create, natsOptions_SetSecure, natsOptions_SetEventLoop for libevent, ...)
    natsConnection_Connect
}
```

Because I misconfigured the certificate, I kept getting the following errors:
```
natsStatus_GetText[SSL Error], nats_GetLastError[(conn.c:712): SSL handshake error: (null)]
natsStatus_GetText[SSL Error], nats_GetLastError[(conn.c:712): SSL handshake error: (null)]
natsStatus_GetText[SSL Error], nats_GetLastError[(conn.c:712): SSL handshake error: (null)]
```

And soon I found the socket is leaking:
```
# lsof -p 7710
...
nats-daem 7710 root   26u     sock                0,8      0t0 57832077 protocol: TCP 
nats-daem 7710 root   27u     sock                0,8      0t0 57832658 protocol: TCP 
nats-daem 7710 root   28u     sock                0,8      0t0 57832728 protocol: TCP   
nats-daem 7710 root   29u     sock                0,8      0t0 57832841 protocol: TCP 
nats-daem 7710 root   30u     sock                0,8      0t0 57833356 protocol: TCP 
nats-daem 7710 root   31u     sock                0,8      0t0 57835198 protocol: TCP 
nats-daem 7710 root   32u     sock                0,8      0t0 57838069 protocol: TCP 
nats-daem 7710 root   33u     sock                0,8      0t0 57838165 protocol: TCP 
nats-daem 7710 root   34u     sock                0,8      0t0 57838282 protocol: TCP  
nats-daem 7710 root   35u     sock                0,8      0t0 57840396 protocol: TCP   
nats-daem 7710 root   36u     sock                0,8      0t0 57843879 protocol: TCP 
nats-daem 7710 root   37u     sock                0,8      0t0 57844195 protocol: TCP 
nats-daem 7710 root   38u     sock                0,8      0t0 57844266 protocol: TCP 
...
```

Here's my guess:
In natsConnection_Connect -> _connect, the socket is inited in _createConn, but got SSL error in _processConnInit, so the socket should be closed in _close.

```
static natsStatus
_connect(natsConnection *nc)
{
    s = _createConn(nc);
    if (s == NATS_OK)
    {
        s = _processConnInit(nc);
        if (s == NATS_OK)
        {
            ...
        }
        else
        {
            _close(nc, NATS_CONN_STATUS_DISCONNECTED, false, false);
        }
    }
}
```

Because this daemon uses libevent, nc->opts->evLoop won't be NULL. The socket will only be shutdown but not released in _close. Based on the comment, the socket should be released by _readLoop,
```
static void
_close(natsConnection *nc, natsConnStatus status, bool fromPublicClose, bool doCBs)
{
        // If there is no readLoop, then it is our responsibility to close
        // the socket. Otherwise, _readLoop is the one doing it.
        if ((ttj.readLoop == NULL) && (nc->opts->evLoop == NULL))
        {
            natsSock_Close(nc->sockCtx.fd);
        }
        else
        {
            // Shutdown the socket to stop any read/write operations.
            // The socket will be closed by the _readLoop thread.
            natsSock_Shutdown(nc->sockCtx.fd);
        }
}
```

However, in the time of SSL handshake failure, _readLoop thread hasn't started yet.

```
static natsStatus
_processConnInit(natsConnection *nc)
{
    // Where the SSL handshake happened
    s = _processExpectedInfo(nc);
    ...
    // Start the readLoop and flusher threads
    if (s == NATS_OK)
        s = _spinUpSocketWatchers(nc);
}
```

I think simply change the line here
```
-        if ((ttj.readLoop == NULL) && (nc->opts->evLoop == NULL))
+        if (ttj.readLoop == NULL)
```
should do the trick?

Haven't tried v2.3.0 but the code path is almost the same, so I think it still exist?

kozlovic:
Two issues: if using an event loop and SSL but the connection
fails before the event loop is attached, the library needs
to close the socket.
On normal socket shutdown, since there is no readloop in the
context of external event loop, ensure that the library closes
the socket properly.

Resolves #374

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -203,6 +203,7 @@ _freeConn(natsConnection *nc)
     natsMutex_Destroy(nc->subsMu);
     natsTimer_Destroy(nc->drainTimer);
     natsMutex_Destroy(nc->mu);
+    natsCondition_Destroy(nc->el.cond);
 
     NATS_FREE(nc);
 
@@ -1410,6 +1411,10 @@ _doReconnect(void *arg)
 
     natsConn_Lock(nc);
 
+    // For external event loop, wait for the socket to be closed.
+    while ((nc->opts->evLoop != NULL) && !nc->el.sockClosed)
+        natsCondition_Wait(nc->el.cond, nc->mu);
+
     // Kick out all calls to natsConnection_Flush[Timeout]().
     _clearPendingFlushRequests(nc);
 
@@ -1839,6 +1844,7 @@ _processConnInit(natsConnection *nc)
             // Set this first in case the event loop triggers the first READ
             // event just after this call returns.
             nc->sockCtx.useEventLoop = true;
+            _retain(nc);
 
             s = nc->opts->evCbs.attach(&(nc->el.data),
                                        nc->opts->evLoop,
@@ -1851,6 +1857,7 @@ _processConnInit(natsConnection *nc)
             else
             {
                 nc->sockCtx.useEventLoop = false;
+                _release(nc);
 
                 nats_setError(s,
                               "Error attaching to the event loop: %d - %s",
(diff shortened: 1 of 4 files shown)


src/conn.c
```diff
@@ -1410,6 +1411,10 @@ _doReconnect(void *arg)
 
     natsConn_Lock(nc);
 
+    // For external event loop, wait for the socket to be closed.
+    while ((nc->opts->evLoop != NULL) && !nc->el.sockClosed)
+        natsCondition_Wait(nc->el.cond, nc->mu);
```
derekcollison, line 1416: Should the other side to a broadcast and not just a directed signal?

kozlovic: We are in the same connection object and there can be only a reconnect thread started, so signal is enough.

derekcollison: ok just wanted to make sure we could not have a weird situation where multiple are in Wait().

kozlovic: Not multiple, but that forced me to take a closer look to make sure that we don't start multiple time the _doReconnect thread when I realize that there is the condition of "retry on failed connect" in which case _doReconnect should not wait to be notified that the socket is closed, because the socket was never "opened". So added a test for that and the fix. Thanks!

derekcollison (approved):
LGTM

PR #380, opened 2020-12-02
[CHANGED] Simplify compilation of event libraries examples

igor-sinelnikow:
Default install location of event libraries doesn't match default values of `LIBUV_DIR` and `LIBEVENT_DIR` variables. Typically headers go to `/usr/include` and libraries - to something like `/usr/lib/x86_64-linux-gnu` depending on the architecture (see [File list of package libevent-dev/bionic/amd64](https://packages.ubuntu.com/bionic/amd64/libevent-dev/filelist), [File list of package libuv1-dev/bionic/amd64](https://packages.ubuntu.com/bionic/amd64/libuv1-dev/filelist)).
So if I want to install libuv with apt-get and build Libuv example I would have to manually create a symlink to .so in `/usr/lib` and pass `-DLIBUV_DIR=/usr` to cmake, for instance.

I suggest to refuse from hard-coded paths: set empty `LIB*_DIR` by default and compose full paths to libraries and headers only when it is set explicitly. Otherwise, rely on plain library names and default include directories. This allows not to specify these variables in most cases.

I believe this change doesn't break backward compatibility:
- if the variables were set explicitly, then the behavior doesn't change at all;
- if the variables were not set, then `/usr/local/include` and `/usr/local/lib` were used. Compiler and linker will search in these directories anyway.


kozlovic (approved):
LGTM! I have checked that if I have an existing cmake build directory, your branch still works (uses the old lib*_dir references, so I still can build libevent/libuv examples). If I have a brand new cmake build directory with your branch, the lib*_dir are blank and building the libevent/libuv examples on my machine would fail (macOS) but then I set to /usr/local and it builds fine.

PR #381, opened 2020-12-05
[FIXED]: segmentation fault when ConnectTo urls parameter is NULL

paoloteti:
Check if `urls` list is `NULL`, before parsing it.

Internally the library will use `NATS_DEFAULT_URL` and therefore `NULL` become an an alias to `NATS_DEFAULT_URL`.

Patch include also a dedicated test case.

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3117,11 +3117,14 @@ _processUrlString(natsOptions *opts, const char *urls)
     char        *ptr         = NULL;
     int         len;
 
-    ptr = (char*) urls;
-    while ((ptr = strchr(ptr, ',')) != NULL)
+    if (urls != NULL)
     {
-        ptr++;
-        count++;
+        ptr = (char*) urls;
+        while ((ptr = strchr(ptr, ',')) != NULL)
+        {
+            ptr++;
+            count++;
+        }
     }
     if (count == 0)
         return natsOptions_SetURL(opts, urls);
--- a/src/nats.h
+++ b/src/nats.h
@@ -2545,6 +2545,8 @@ natsConnection_ProcessWriteEvent(natsConnection *nc);
  * @param nc the location where to store the pointer to the newly created
  * #natsConnection object.
  * @param urls the URL to connect to, or the list of URLs to chose from.
+ * If `NULL` this call is equivalent to #natsConnection_ConnectTo()
+ * with #NATS_DEFAULT_URL
  */
 NATS_EXTERN natsStatus
 natsConnection_ConnectTo(natsConnection **nc, const char *urls);
--- a/test/list.txt
+++ b/test/list.txt
@@ -58,6 +58,7 @@ IPResolutionOrder
 UseDefaultURLIfNoServerSpecified
 ConnectToWithMultipleURLs
 ConnectionWithNULLOptions
+ConnectionToWithNullURLs
 ConnectionStatus
 ConnClosedCB
 CloseDisconnectedCB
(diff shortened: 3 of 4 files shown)


src/conn.c
```diff
@@ -3117,6 +3117,9 @@ _processUrlString(natsOptions *opts, const char *urls)
     char        *ptr         = NULL;
     int         len;
 
+    if (urls == NULL)
```
kozlovic, line 3120: I agree that `natsConnection_To()` sounds like the user should provide the URL(s), so if NULL is passed, it is reasonable to fail. But what do you think of simply do processing of `urls` here if not NULL. If NULL, then `if (count == 0) will be true and we will fallback calling natsOptions_SetURL() with NULL, which is OK. The end-result would be that internally the library will use the DEFAULT_URL.

src/conn.c
```diff
@@ -3180,6 +3183,8 @@ natsConnection_ConnectTo(natsConnection **newConn, const char *url)
     s = natsOptions_Create(&opts);
     if (s == NATS_OK)
         s = _processUrlString(opts, url);
+    if (s != NATS_OK)
```
kozlovic, line 3186: If we make `_processUrlString()` fail, I would change to this to make it a bit more clear why we need to destroy the options in that case:
```
if (s == NATS_OK)
{
    s = _processUrlString(opts, url);
    // Until natsConn_create() takes ownership of the options, we are 
    // responsible for destroying it in case of error.
    if (s != NATS_OK)
        natsOptions_Destroy(opts);
}
```

src/conn.c
```diff
@@ -3117,6 +3117,9 @@ _processUrlString(natsOptions *opts, const char *urls)
     char        *ptr         = NULL;
     int         len;
 
+    if (urls == NULL)
+        return NATS_INVALID_ARG;
```
kozlovic, line 3121: If we think it is better to have this fail, then I would call:
```
return nats_setDefaultError(NATS_INVALID_ARG);
```
so that it updates the stack and we know where the error comes from, or even better:
```
return nats_setError(NATS_INVALID_ARG, "%s", "urls argument cannot be NULL");
```
which would produce this kind of trace:
```
Error: 16 - Invalid Argument - (conn.c:3121): urls argument cannot be NULL
Stack: (library version: 2.3.0)
  01 - _processUrlString
  02 - natsConnection_ConnectTo
```

kozlovic (changes requested):
Paolo, thank you again for your contribution. I wonder if we want this to fail or default to the default URL. If we want it to fail, I have suggested some changes. Thanks!

src/conn.c
paoloteti: I agree that it is common in C to use NULL mainly for optional, but also for default values. I got this issue because in my apps urls comes from a file and the parser returns NULL for options and urls. For options was ok, but not for urls.

I choose to have this fail because in the [documentantion ](https://docs.nats.io/developing-with-nats/connecting/default_server) and also in other places in the [code](https://github.com/nats-io/nats.c/blob/master/src/conn.c#L3088) default URL is always used as  `natsConnection_ConnectTo(&conn, NATS_DEFAULT_URL)`

BTW what is important here is just to guard `strchr(..)` against NULL values.

Let me know what do you think, and I'll change the code accordingly.

paoloteti: PR updated to default to DEFAULT_URL when urls=NULL

kozlovic (approved):
LGTM. Looks great! Thanks again for your contributions.

PR #382, opened 2020-12-06
Destroy options in case of failure before natsConn_create runs

kozlovic:
natsConn_create() is taking ownership of the options. If anything
fails before that point, we need to destroy the options that have
been created before that point.


kozlovic:
@paoloteti Based on your latest PR, I have taken care of destroying the options if something fails before the call to natsConn_create(), before that call takes ownership of the options.

paoloteti:
LGTM

PR #385, opened 2020-12-20
Improve integer values parsing

paoloteti:
After fuzzing my application I found some unexpected behavior when parsing port number as part of URLs.
This PR enhance integer parsing with a better error handling including overflow detection.

- Add `nats_ParseInt()` to replace `atoi()` that doesn't manage malformed
  input and overflow.
- Add support for overflow detection in `nats_ParseInt64()`

Semantics is not the same: `nats_ParseInt64()` doesn't allow white-spaces and it is mainly intended to be used to parse SIDs in messages headers.

### DIFF
--- a/src/url.c
+++ b/src/url.c
@@ -16,6 +16,7 @@
 #include <string.h>
 
 #include "mem.h"
+#include "util.h"
 
 void
 natsUrl_Destroy(natsUrl *url)
@@ -40,7 +41,10 @@ _parseHostAndPort(natsUrl *url, char *host, bool uInfo)
     if (sport != NULL)
     {
         if (sport[1] != '\0')
-            url->port = atoi(sport + 1);
+        {
+            if (!nats_ParseInt(sport + 1, &(url->port)))
+                return nats_setDefaultError(NATS_INVALID_ARG);
+        }
 
         *sport = '\0';
     }
--- a/src/util.c
+++ b/src/util.c
@@ -15,8 +15,10 @@
 
 #include <stdio.h>
 #include <stdarg.h>
+#include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
+#include <limits.h>
 
 #include "util.h"
 #include "mem.h"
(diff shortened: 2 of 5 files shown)


src/url.c
```diff
@@ -40,7 +41,10 @@ _parseHostAndPort(natsUrl *url, char *host, bool uInfo)
     if (sport != NULL)
     {
         if (sport[1] != '\0')
-            url->port = atoi(sport + 1);
+        {
+            if (!nats_ParseInt(sport + 1, &(url->port)))
```
kozlovic, line 45: I would have no problem not supporting white spaces (for instance the Go client would reject any white space or non numerical value). So we could simply here invoke nats_ParseInt64 (would have first to determine the length of the remaining string), and since the port is a `int` then fail if result is negative or greater than INT32_MAX. We can use `return nats_setError(NATS_INVALID_ARG, "invalid port: '%s'", sport+1);` to be more informative about the error.

src/util.c
```diff
@@ -82,12 +86,46 @@ nats_ParseInt64(const char *d, int dLen)
...
+        if (((n * 10) > (INT64_MAX - k)) || (n > (INT64_MAX / 10)))
+            return -1;
+
+        n = (n * 10) + k;
     }
 
     return n;
 }
 
+// nats_ParseInt expects decimal integer numbers.
```
kozlovic, line 101: I would not add this function. Note that I had to remove use of strtol in the past due to locale and security issues.

kozlovic (changes requested):
Thanks again for contributing. I would not address the issue this way. nats_ParseInt64 needs indeed fixing so that it returns -1 (and only -1) in case of overflow.
Parsing the port could make use of that function (since we already check that it has to be digits only) and add extra overflow check since the port is saved as `int` and not `int64_t`.
Note that I don't think that parsing of the port should remove prefix and suffix white spaces. But it should indeed fail and report that the port string is invalid when getting non digits and/or overflow.

src/url.c
paoloteti: I'll check better tomorrow (due to the timezone). BTW I started using `nats_ParseInt64()`, but if I'm not wrong the test-case `ConnectToWithMultipleURLs` has a sub-case with trailing spaces that `atoi()` instead manage, since it stops at the first non numerical value. So the length of the remaining string should not include spaces and, at the end, it is like using nats_ParseInt64() + discarding spaces + overflow check on int. This was the reason of `nats_ParseInt()`.

paoloteti:
Closed. Replaced by #387

PR #386, opened 2020-12-21
[CHANGED] Default CMAKE_MACOSX_RPATH to ON

### LINKED ISSUE #384: RPATHs disabled on OSX
Hi all,

It seems like RPATHs are explicitly disabled on OSX:
`CMakeLists.txt Line 170:     set(CMAKE_MACOSX_RPATH OFF)`

This makes it very difficult to use nats.c as part of another project on OSX, and makes it impossible to do so when the other project is Python based. This is because OSX System Integrity Protection (SIP) disables the use of DYLD_LIBRARY_PATH on Python and other executables contained in system locations, which makes it impossible for a Python-based project to load any shared libs that in turn load nats. This is a common use-case for Python interface libraries created using SWIG.

Besides, RPATHs are integral to properly designing relocatable apps on modern OSX, so CMAKE_MACOSX_RPATH should be left to the user and not explicitly set inside CMakeLists.txt.

I recommend removing that line and restoring the user's ability to set CMAKE_MACOSX_RPATH.

Thank you!
Ravi

kozlovic:
Made it a configurable option with `make edit_cache` to make it
easy for the user to change.

Resolves #384


kozlovic:
@ravidavi Can you please have a look?

ravidavi (approved):
This looks like a great solution, thanks!

PR #387, opened 2020-12-22
Rework of URLs parsing

kozlovic:
- URLs are now trimmed in natsOptions_SetURL and natsOptions_SetServers
- Invalid port reports an error
- Check for parse int overflow

Related to #385

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3117,7 +3117,6 @@ _processUrlString(natsOptions *opts, const char *urls)
     char        *urlsCopy    = NULL;
     char        *commaPos    = NULL;
     char        *ptr         = NULL;
-    int         len;
 
     if (urls != NULL)
     {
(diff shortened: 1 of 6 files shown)


kozlovic:
@paoloteti I thank you for your PR, but I feel like having a URL with say "nats://localhost:  4222  " should fail. I have no problem with that. So I have moved the trimming of URLs deep down so that all places where we pass URLs will benefit from the change. If there is an error in processing the port, you will get a more meaningful error message. For instance:
```
$ examples/nats-publisher -s "nats://localhost: 4222"
Sending 1000000 messages to subject 'foo'
Error: 16 - Invalid Argument
Error: 16 - Invalid Argument - (util.c:112): invalid port ' 4222'
```
I have also taken some of your additions to the parsing test and also check for parseint64 overflow. Thanks!

paoloteti:
LGTM
Tested locally. I'll close my PR.

kozlovic:
@paoloteti Thanks for the review.

PR #395, opened 2020-12-31
[FIXED] Subscription could not be closed while draining

### LINKED ISSUE #383: Question about natsSubscription_Drain
Hello. I use natsSubscription_Drain instead of natsSubscribption_Unsubscribe everywhere because it provides stronger guarantees that message dispatched to the subscription on the server-side will not be lost.

I wish I can use this function in the following way:
1) call natsSubscription_Drain - documentation states that it performs asynchonously
2) Wait for subscription closed callback to be called (set with natsSubscription_SetOnCompleteCB)
3) Call natsSubscription_WaitForDrainCompletion inside onCompleteCB using zero timeout to obtain information whether subscription was drained or not,

Can you please tell me, whether 3 wil work? Will natsSubscription_WaitForDrainCompletion return the correct value? Will it return immediately or block for some time? I do not want to call natsSubscription_WaitForDrainCompletion on a separate thread to avoid blocking...

Also I've found some strange behaviour in my tests:
1) Client opens connection to the server, creates subscription, receives some messages on it and then the server is stopped
2) Client is configured to reconnect indefinetely, so it starts reconnecting after it detects connection loss (onDisconnectedCB is called)
3) natsSubscription_Drain is called for subscription created in 1 and freezes for a pretty big amount of time (10 seconds in my test) and then returns NATS_TIMEOUT (code 26). Subscription closed CB is not called.

Note, that documenation states that this function is not blocking and I think 10 seconds is a pretty big value for non-blocking. Maybe a better solution will be to immediately return error when disconnected? Is there any sense at all to try to perform subscription drain when connection is lost? I mean, I thought that drain is needed to receive all messages buffered on the server-side - but the connection is lost, and susbcription is already removed on the server (if I understand correctly).

### LINKED ISSUE #394: Should be able to close subscription while draining
If user calls `natsSubscription_Drain(sub)`, then it is impossible to make the draining stop (except by calling natsConnection_Close()).
It should be possible to call `natsSubscription_Unsubscribe()` while the subscription is draining. This should make the draining stop after the current message callback returns.

kozlovic:
This addresses several issues related to drain:
- Subscriptions could not be closed (unsubscribe) by the user once
draining had started
- The natsConnection_Drain() and natsSubscription_Drain() were
possibly blocking since those functions would internally call
natsConnection_Flush().

The calls are now not blocking (the flush is done in a library
thread), the subscription can be closed while drain is in progress.

Two new APIs were added:
- natsSubscription_DrainTimeout(): this allows to specify a timeout.
When the timeout is reached and if the subscription is not fully
drained, the subscription will automatically closed.
- natsSubscription_DrainCompletionStatus(): once the user has verified
that the drain is complete, this function can be used to know if the
drain completed successfully or not. It returns NATS_OK if drain
was completed without error, but would return NATS_INVALID_SUBSCRIPTION
if the subscription was closed prematurely, or NATS_CONNECTION_CLOSED
if it was the connection.

Resolves #394
Resolves #383

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3424,8 +3424,7 @@ _iterateSubsAndInvokeFunc(natsStatus callerSts, natsConnection *nc, subIterFunc 
     {
         sub = (natsSubscription*) p;
         ls = (f)(callerSts, nc, sub);
-        if ((ls != NATS_OK) && (s == NATS_OK))
-            s = ls;
+        s = (s == NATS_OK ? ls : s);
     }
     natsHashIter_Done(&iter);
     natsMutex_Unlock(nc->subsMu);
@@ -3519,11 +3518,10 @@ _flushAndDrain(void *closure)
         s = NATS_OK;
 
         // Now wait for the number of subscriptions to go down to 0, or deadline is reached.
-        while (deadline - nats_Now() > 0)
+        while ((timeout == 0) || (deadline - nats_Now() > 0))
         {
             natsConn_Lock(nc);
-            closed = natsConn_isClosed(nc);
-            if (!closed)
+            if (!(closed = natsConn_isClosed(nc)))
             {
                 natsMutex_Lock(nc->subsMu);
                 subsDone = (natsHash_Count(nc->subs) == 0 ? true : false);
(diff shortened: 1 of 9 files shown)


kozlovic:
@pananton This (big) PR not only fixes the issue that the subscription could not be closed by the user after a drain was initiated, but it also makes sure that the drain are non blocking. I also added ability to drain the subscription with a timeout and added a function to get the status of the subscription drain (after it has completed). So I did not change the behavior of `natsSubscription_WaitForDrainCompletion()` as previously discussed.

I need to update this PR with a bit more tests since code coverage reduced a bit, but that will give you an idea of how the drain will now work and have a look at the new added APIs.

kozlovic:
@pananton From issue #383 you wrote:

> Looked into PR and it's great, thanks!

I made an update today. Would you have a chance to use the code from this PR to build and check that it works as expected in your project? If not, that's ok, I will merge. Thanks!

pananton:
Sorry for not answering earlier. Yes, I will update my code with new nats.c master branch. I also will check whether everything works as expected, but this will take some time.

PR #399, opened 2021-01-13
[FIXED] Receiving subject mapped replies with new style requests

kozlovic:
In somce cases (like JetStream) it is possible that the server
rewrites the subject of the reply sent to a requestor. When the
requestor is not using the old style requests (meaning that there
is a connection level subscription on say `_INBOX.<nuid>.*` and
replies are dispatched based on the last token, the requests would
always fail since the response token would not be found in the map.


derekcollison (approved):
LGTM

PR #400, opened 2021-01-26
Add a null check before freeaddrinfo()

vabalas:
Calling `freeaddrinfo()` with a `NULL` is undefined behavior.

Consider the [freeBSD manpage of freeaddrinfo(3)](https://www.freebsd.org/cgi/man.cgi?freeaddrinfo(3)):
```
IMPLEMENTATION NOTES
     The behavior of freeadrinfo(NULL) is left unspecified by both Version 4 of
     the Single UNIX Specification (“SUSv4”) and RFC 3493.  The current
     implementation ignores a NULL argument for compatibility with programs that
     rely on the implementation details of other operating systems.
```

Encountered this problem when targeting musl libc. See [the musl implementation here](https://git.musl-libc.org/cgit/musl/tree/src/network/freeaddrinfo.c). No-op is not guaranteed in the standard.

It is my first PR to `nats.c` so any feedback is very welcome.

### DIFF
--- a/src/comsock.c
+++ b/src/comsock.c
@@ -156,7 +156,8 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
     if (numIPs == 0)
     {
         for (i=0; i<numServInfo; i++)
-            freeaddrinfo(servInfos[i]);
+            if (servInfos[i])
+                freeaddrinfo(servInfos[i]);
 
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM. Thank you for your first contribution! My feedback would be that this could have been made an util function, say nats_freeAddrInfo() that would then take care of checking for NULL. But this is fine like that. Thank you again!

PR #401, opened 2021-01-27
Code cleanup

vabalas:
**This is a draft PR.** Following the helpful feedback from @kozlovic (See #400) I want to wrap the culprit `freeaddrinfo()` with a null check. 

Sorry for disturbing you because of a minor nuisance, but could you please indicate in which header should I put it (since `util.h` is not included everywhere)?

I appreciate your help!

### DIFF
--- a/src/comsock.c
+++ b/src/comsock.c
@@ -22,6 +22,7 @@
 #include "status.h"
 #include "comsock.h"
 #include "mem.h"
+#include "util.h"
 
 natsStatus
 natsSock_Init(natsSockCtx *ctx)
@@ -156,8 +157,7 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
     if (numIPs == 0)
     {
         for (i=0; i<numServInfo; i++)
-            if (servInfos[i])
-                freeaddrinfo(servInfos[i]);
+            nats_FreeAddrInfo(servInfos[i]);
 
         return NATS_UPDATE_ERR_STACK(NATS_NO_SERVER);
     }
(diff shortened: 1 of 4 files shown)


src/util.c
```diff
@@ -1373,6 +1373,13 @@ nats_ReadFile(natsBuffer **buffer, int initBufSize, const char *fn)
     return NATS_UPDATE_ERR_STACK(s);
 }
 
+void nats_FreeAddrInfo(struct addrinfo *res)
```
kozlovic, line 1376: To follow the code style and fix the issue reported on compile, I would write like this:
```
void
nats_FreeAddrInfo(struct addrinfo *res)
{
    // Calling freeaddrinfo(NULL) is undefined behaviour.
    if (res == NULL)
        return;

    freeaddrinfo(res);
}
```

src/comsock.c
```diff
@@ -156,8 +156,7 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
     if (numIPs == 0)
     {
         for (i=0; i<numServInfo; i++)
-            if (servInfos[i])
-                freeaddrinfo(servInfos[i]);
+            nats_FreeAddrInfo(servInfos[i]);
```
kozlovic, line 159: You need to add:
```
#include "util.h"
```
at top of this file.

kozlovic (changes requested):
Suggested some changes to fix compilation errors.

vabalas:
I appreciate your help very much! I added the amendments you requested.

kozlovic (approved):
LGTM

PR #403, opened 2021-02-25
[ADDED] natsConnection_GetLocalIPAndPort() function

### LINKED ISSUE #402: Expose Client Ip/Port From Socket
I would love to have observability into each `natsConnection`'s Ip and Port. I am aware of `natsConnection_GetClientIP`:

https://github.com/nats-io/nats.c/blob/6b6e4004975f2e35499d397c422ba61b9a6f5717/src/conn.c#L4131

but this IP comes from the streaming server itself. It also lacks a port number. As part of some end-to-end testing, I am working on simulating network issues in the connection from the NATS client to the server. However, I need observability in code into what client Ip/Port I am using (I have multiple NATS connections from the same host, so I need the information in the client API itself).

Does this exists already? It seems like it would require a code change.

If so, wouldn't it be possible to put something into the connection code itself to retrieve the Ip and Port client-side? Something like calling `getsockname` on the file descriptor here:

https://github.com/nats-io/nats.c/blob/dc6ce2c9d4722b5782de19eeece1e49083ce76fa/src/comsock.c#L207-L230

kozlovic:
This returns the connection's local IP and port.

Resolves #402

### DIFF
--- a/src/comsock.c
+++ b/src/comsock.c
@@ -543,3 +543,39 @@ natsSock_InitDeadline(natsSockCtx *ctx, int64_t timeout)
     natsDeadline_Init(&ctx->readDeadline, timeout);
     natsDeadline_Init(&ctx->writeDeadline, timeout);
 }
+
+void
+natsSock_GetLocalIPAndPort(natsSockCtx *ctx, char **ip, int *port)
+{
+    struct sockaddr_storage addr;
+    natsSockLen             addrLen = (natsSockLen) sizeof(addr);
+    char                    localIP[64];
+    void                    *laddr = NULL;
+    int                     fam;
+
+    *ip = NULL;
+    *port = 0;
+
+    if (getsockname(ctx->fd, (struct sockaddr*) &addr, &addrLen) != 0)
+        return;
+
+    if (((struct sockaddr*) &addr)->sa_family == AF_INET)
+    {
+        struct sockaddr_in *addr4 = (struct sockaddr_in*) &addr;
+
+        fam = AF_INET;
+        laddr = &(addr4->sin_addr);
+        *port = (int) ntohs(addr4->sin_port);
+    }
+    else
+    {
+        struct sockaddr_in6 *addr6 = (struct sockaddr_in6*) &addr;
+
+        fam = AF_INET6;
+        laddr = &(addr6->sin6_addr);
+        *port = (int) ntohs(addr6->sin6_port);
+    }
+
+    if (inet_ntop(fam, laddr, localIP, sizeof(localIP)) != NULL)
+        *ip = NATS_STRDUP(localIP);
+}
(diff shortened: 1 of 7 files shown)


src/comsock.c
```diff
@@ -543,3 +543,39 @@ natsSock_InitDeadline(natsSockCtx *ctx, int64_t timeout)
...
+
+    if (((struct sockaddr*) &addr)->sa_family == AF_INET)
+    {
+        struct sockaddr_in *addr4 = (struct sockaddr_in*) &addr;
+
+        fam = AF_INET;
+        laddr = &(addr4->sin_addr);
+        *port = (int) ntohs(addr4->sin_port);
+    }
+    else
```
samc1213, line 570: Maybe safer to check for AF_INET6 here, else consider an error?

src/comsock.c
```diff
@@ -543,3 +543,39 @@ natsSock_InitDeadline(natsSockCtx *ctx, int64_t timeout)
...
+    natsSockLen             addrLen = (natsSockLen) sizeof(addr);
+    char                    localIP[64];
+    void                    *laddr = NULL;
+    int                     fam;
+
+    *ip = NULL;
+    *port = 0;
+
+    if (getsockname(ctx->fd, (struct sockaddr*) &addr, &addrLen) != 0)
+        return;
```
samc1213, line 560: Should this method return a NATS_OK by default, or an error here?

src/nats.h
```diff
@@ -2878,6 +2878,26 @@ natsConnection_GetRTT(natsConnection *nc, int64_t *rtt);
 NATS_EXTERN natsStatus
 natsConnection_HasHeaderSupport(natsConnection *nc);
 
+/** \brief Returns the connection local IP and port.
+ *
+ * Unlike #natsConnection_GetClientIP, this function returns the
```
samc1213, line 2883: Might be helpful to point to this option in the natsConnection_GetClientIP documentation

samc1213 (commented):
I didn't pull this change into my project, but based on the implementation and tests, this seems great!

src/comsock.c
kozlovic: Given that the socket is connected with only any of those two, that should never happen, but will do.

src/comsock.c
kozlovic: At first I was doing that inside the connect and did not want to fail a connection just because of error getting the socket name, etc.. but since I have made that a separate call now, I will, thanks.

src/nats.h
kozlovic: Will do.

kozlovic:
@samc1213 I think I have addressed all your comments. Have a look when you get a chance. Thanks!

src/comsock.c
samc1213: I see, yeah this seems like a better place than the connect

samc1213 (approved):

kozlovic:
@samc1213 Thank you for the report and the review!

PR #408, opened 2021-03-05
Fix tabulation handling in parser

Enjection:


kozlovic (approved):
Good catch! Thank you for your contribution. I will add a test after merging your PR.

kozlovic:
@Enjection Thanks again for the contribution. I have added a test that would have caught the issue in PR #409. If you want to have a look.

PR #418, opened 2021-04-01
[FIXED] URL parsing returns "invalid port" if path after port

### LINKED ISSUE #417: Parsing URL fails with "invalid port" error if there is a path after the port
For instance, using this URL: `nats://localhost:4222/subject` will fail since v2.4.0 but was working before.
Note that NATS does not use anything after the port, but still, there is no reason to fail the parsing.

Reported by @RudeDude and @counterflow-ai.

kozlovic:
This was introduced in v2.4.0.

Resolves #417

### DIFF
--- a/src/url.c
+++ b/src/url.c
@@ -32,6 +32,21 @@ natsUrl_Destroy(natsUrl *url)
 }
 
 static natsStatus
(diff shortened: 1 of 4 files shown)


RudeDude:
Works for me!

PR #420, opened 2021-05-24
[FIXED] Slow consumer for subscriptions receiving messages with headers

### LINKED ISSUE #419: Slow Consumer Stops Response Messages
We’re running Nats Server 2.2.1. We have a cluster of 20 servers, spread across 2 sites (10 servers per site). There’s a Nats Server instance running on each server. Nats is configured as full mesh of gateways across the 20 servers. Each server runs a set of services each connecting to their local Nats Server instance. We’re using TLS for all Nats communication. We use the C, Go, and Java Nats client libraries.

We’ve observed some strange behaviour, and I’m struggling to find anything useful that might point me in the right direction. Twice now, one particular service (of which there’s an instance per server) stops receiving responses to requests. Assume we have services A, B, and C (A -> B -> C). Our problematic service is B. A sends a message to B, and B receives it. B sends a message to C, and C receives it. C responds to B’s request, but B never receives the response. After a timeout period, B responds to A, and A receives the response. Restarting B causes B to start receiving responses from C again.

We noticed that all of the instances of B on one site will stop receiving responses at roughly the same time. A while later, all of the instances of B on the other site will stop receiving responses at roughly the same time. We’ve observed this to happen on two occasions both on a Monday afternoon. So there maybe something external causing something to go wrong. The customer may have some kind of scheduled scanning taking place.

Service B is written in C and is using the nats.c 2.4.0 library. We create an asynchronous subscription to receive responses on a given subject. We use an asynchronous queue subscription for inbound requests. We’ve set the close, disconnect, reconnect, and error callbacks. Each of these callbacks simply prints an error message to the log file (unfortunately no further information about the connection or subscription is printed to the log file). At the time B stops receiving responses, the error callback is called with the error “Slow Consumer, messages dropped”. This is logged a few times, after which the service stops receiving responses. None of the other callbacks are called (their log statements don’t appear in the log file). The error never appears again after that. We have a pending limit of 1000 messages, but each instance of service B only receives requests at a rate of about 4 TPS, so it’s not particularly high. Service B usually responds in less that a second, and just before the problem occurs, response times are typical.

We have other services using the same framework receiving messages at much higher rates. We started using version 2.4.0 of the Nats C client library at the beginning of April, and this behaviour only showed itself last week (17th May 2021). Other than this, service B has been running for well over a year with Nats Server without incident.

While service B is in this state, if I check the subscriptions in Nats Server (http://127.0.0.1:8222/subsz?subs=1), I see the correct subscriptions. If I check the connection for service B in Nats Server (http://127.0.0.1:8222/connz) for the subscriptions, I see it there with the correct number of subscriptions. The connection uptime is 7+ days. We have a debug statement right in the Nats message callback handler within our framework, but I never see this being called for the responses.

At this point I’m a bit stumped. Are there any reasons why after getting “Slow Consumer, messages dropped” we’d stop receiving response messages? From the Nats Server perspective, it looks like everything is in place. I though that slow consumer would mean we might lose a few messages, but then message processing would resume, but it does not seem to be the case for some reason.

kozlovic:
The accounting for subscription's pending bytes was wrong because
the whole message buffer was added (counting headers) but then
only the data portion was removed once message was dispatched.

Over time, this would lead to a subscription's pending bytes
possibly higher than the pending limit which would cause the
connection to always drop messages for this subscription.

Resolves #419

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2518,6 +2518,7 @@ natsConn_processMsg(natsConnection *nc, char *buf, int bufLen)
     natsMsg          *msg = NULL;
     natsMsgDlvWorker *ldw = NULL;
     bool             sc   = false;
+    int              dl   = 0;
 
     natsMutex_Lock(nc->subsMu);
 
@@ -2537,6 +2538,10 @@ natsConn_processMsg(natsConnection *nc, char *buf, int bufLen)
     s = _createMsg(&msg, nc, buf, bufLen, nc->ps->ma.hdr);
     if (s != NATS_OK)
         return s;
+    // bufLen is the overall buffer len. In presence of headers, we need
+    // to capture the real message payload data length, which has been
+    // computed as the bufLen - header size.
+    dl = msg->dataLen;
 
     if ((ldw = sub->libDlvWorker) != NULL)
         natsMutex_Lock(ldw->lock);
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #423, opened 2021-05-26
[ADDED] natsMsg_IsNoResponders API

### LINKED ISSUE #421: Add support for new "no responders" error code for natsConnection_PublishRequest
Hello!
I'm using natsConnection_PublishRequest to publish requests and later parse replies in my own inbox subscription (which is passed in 'reply' function parameter). If there is no responder for the request (I'm using nats-server v2.2.5) then in my inbox I receive a reply for which natsMsg_GetDataLength returns 0 (so it is treated as empty message).
Can you please add some way to figure out that there were no responders for request for my use-case? Need it very much because natsConnection_Request which supports NATS_NO_RESPONDERS does not suite me at all due to it's sync nature.

Possible solutions:

Store natsStatus error code in natsMsg and:

1 - add a `natsStatus*` [out] parameter for natsMsg_GetReply to signal NATS_NO_RESPONDERS for replies
2 - add a function natsMsg_GetError which returns error code for a message

kozlovic:
This is used in message callbacks for subscriptions that have been
set on the reply subject for a PublishRequest calls. This will let
the application know that the request failed due to no-responders.

In other situations, such as natsConnection_Request() or using
the synchronous subscriptions, no message is returned and status
NATS_NO_RESPONDERS is returned.

This also fixes natsSubscription_NextMsg() that should have
returned NATS_NO_RESPONDERS in such situation but was not.

Resolves #421


derekcollison (approved):
LGTM - Note they can turn off that behavior if they want..

kozlovic:
> LGTM - Note they can turn off that behavior if they want..

Which behavior, the "no responders", how?

derekcollison:
https://github.com/nats-io/nats-server/blob/master/server/client.go#L529

kozlovic:
@derekcollison Right, but none of the client library (as far as I know) have an option to disable it (at least Go and this lib does not). That is, the current implementation is that if the server's INFO headers boolean is true, then the lib sends headers and no_responders both with value `true`.
But we can certainly add a connect option to disable no_responders. If I do so, I will do in separate PR.

PR #424, opened 2021-05-27
[ADDED] natsOptions_DisableNoResponders() API

kozlovic:
Added an option to disable the "no responders" feature when connecting
to NATS Servers v2.2.0+.

Also made use of natsMsg_IsNoResponders() in a place in STAN code,
and optimized _liftHeaders so that we don't create the headers map
if there is no need to lift (no header in incoming message wire)
and if call was not made from Set or Add APIs.


derekcollison (approved):
LGTM

PR #426, opened 2021-05-28
[FIXED] TLS fails to reconnect after some protocol error

### LINKED ISSUE #425: Client gets stuck reconnecting to TLS server discovered trough auto-discovery.
### Versions
Client: 2.4.1
Server: 2.2.3

### Steps to reproduce the issue
1. We configure our client to connect to one "root" server. We connect to it:
`const char *servers[] = {"tls://username:password@nats1.company.com:4222"};`
3. They start ping/pong-ing back and forth, everything works perfectly.
4. We stop the NATS server on server 1 (nats1.company.com).
5. Our client attempts to connect to server 2 (nats2.company.com) thanks to the auto-discovery mechanism.
6. Because we didn't include any auth details for nats2.company.com in `*servers[]` or trough `natsOptions_SetUserInfo` the server obviously returns a "Authorization Violation" (which should happen).
7. The client is now frozen / doesn't respond to anything anymore. Zero callbacks are fired, no reconnects are attempted back to nats1.company.com, nothing.
8. Turning off / restarting the servers does nothing, the client is permanently stuck in a frozen state.

### Notes:
- Weird (maybe related?) problem: `natsOptions_SetDiscoveredServersCB` is never fired (even though it connects to server 2 without a problem after stopping server 1 for the first time).
- `natsOptions_SetRetryOnFailedConnect` doesn't seem to have any effect on the problem.

### What's the expected result?
- The `natsOptions_SetClosedCB` callback is fired because the max reconnect attempts are reached and all servers have been tried.
- The client attempts to connect to server 1 again after x attempts on server 2 or vice versa? Not sure what NATS does exactly in a situation like this with clustering enabled.

### What's the actual result?
- Server 2 logs: `TLS handshake error: local error: tls: unexpected message` and `Client connection closed: TLS Handshake Failure` after the reconnect from server 1.
- Client gets stuck / frozen. Nothing happens anymore, even when I stop all servers.
- `natsOptions_SetDisconnectedCB` (from stopping server 1) is the last callback ever fired.
- Setting the `natsOptions_SetUserInfo` option or configuring server 2 trough  `*servers[]`  fixes everthing.

### Additional details
- Proof of concept code: https://pastebin.com/0jqniraD (build and run with `gcc main.c -levent -levent_pthreads -lnats -Wall -O2 -std=c99 -o client && ./client`
- Debug and trace logs from server 1 and server 2: https://pastebin.com/NDzBkC6P

kozlovic:
If a client is connected with TLS, then reconnects to a server
and the TLS hanshake is successful, but an error occurs as
part of the NATS connect handshake, then it would leave the
TLS connection in a bad state that would prevent to successfully
reconnect.

Resolves #425

### DIFF
--- a/src/comsock.c
+++ b/src/comsock.c
@@ -1,4 +1,4 @@
-// Copyright 2015-2019 The NATS Authors
+// Copyright 2015-2021 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -341,7 +341,6 @@ natsSock_Read(natsSockCtx *ctx, char *buffer, size_t maxBufferSize, int *n)
     natsStatus  s         = NATS_OK;
     int         readBytes = 0;
     bool        needRead  = true;
-    bool        wouldBlock= false;
 
     while (needRead)
     {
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

PR #429, opened 2021-06-09
[CHANGED] Headers keys are now case-sensitive

kozlovic:
Prior to v3.0.0, keys were changed to their canonical form, even
when being received from the server. This is no longer the case.
The case is now preserved and all header APIs are case sensitive.


derekcollison (approved):
LGTM

PR #430, opened 2021-06-15
WIP: [ADDED] Build option NATS_BUILD_LIB_SHARED_STATIC to build statically-linked shared library

tkielan:
Please review this PR.


tkielan:
This will add ability to build libnats.dll statically-linked with other mingw libraries on Windows and similar libnats.so linked with gcc libraries on Linux.

src/CMakeLists.txt
```diff
@@ -59,6 +59,10 @@ if(NATS_BUILD_LIB_SHARED)
   set_property(TARGET nats PROPERTY DEBUG_POSTFIX d)
   target_include_directories(nats PUBLIC
         $<INSTALL_INTERFACE:include>)
+  if(NATS_BUILD_LIB_SHARED_STATIC)
+    target_link_options(nats PRIVATE
```
kozlovic, line 63: Not an expert at all, so forgive the question but: why `PRIVATE` here?

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ option(NATS_BUILD_STREAMING "Build NATS Streaming" ON)
 option(NATS_BUILD_NO_PREFIX_CONNSTS "No prefix for connection status enum" OFF)
 option(NATS_BUILD_LIB_STATIC "Build static library" ON)
 option(NATS_BUILD_LIB_SHARED "Build shared library" ON)
+option(NATS_BUILD_LIB_SHARED_STATIC "Build statically-linked shared library" OFF)
```
kozlovic, line 30: Looks like this is available only for CMake 3.13+. Without bumping and if you set the option to ON, you would get something like:
```
CMake Error at src/CMakeLists.txt:63 (target_link_options):
  Unknown CMake command "target_link_options".
```
But since this is OFF by default, it will not break users existing setups, if we bump requirement to 3.13 we may (but requirement to 2.8 is kind of too old anyway).

kozlovic (commented):
Some comments. I have tried on my Windows VM (after updating CMake) and it worked. However, it failed on my Linux VM, maybe because my gcc there is too old (7.5):
```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/crtbeginT.o: relocation R_X86_64_32 against hidden symbol `__TMC_END__' can not be used when making a shared object
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
src/CMakeFiles/nats.dir/build.make:596: recipe for target 'src/libnatsd.so.2.5.0' failed
```

CMakeLists.txt
tkielan: Indeed adding target_link_option enforces cmake 3.13+. I haven't noticed this due to usage of recent cmake.
If you don't mind I would add cmake_minimum_required(VERSION 3.13).
PRIVATE here is what I use in modern cmake (verified, the keyword was not added later after 3.13) to keep -static private to the target and not to make it available for the other targets that use target_link_libraries(xxx PUBLIC -lnats)

src/CMakeLists.txt
tkielan: The PRIVATE here is to make this -static NOT to be inherited by the targets that decided to use nats as PUBLIC
e.g. target_link_libraries(my_3rd_party_target PUBLIC -lnats)

tkielan:
Travis has failed due to CMake 3.12. I don't know how to upgrade CMake to 3.13 on CI.

kozlovic:
@tkielan I have updated the repo to install a recent CMAKE (3.20) with PR: https://github.com/nats-io/nats.c/pull/431, could you please rebase your PR from the main branch?

kozlovic:
@tkielan Any comment on the fact that it failed for me with gcc?
```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/crtbeginT.o: relocation R_X86_64_32 against hidden symbol `__TMC_END__' can not be used when making a shared object
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
src/CMakeFiles/nats.dir/build.make:596: recipe for target 'src/libnatsd.so.2.5.0' failed
```
Do you think it is because my gcc is too old?

tkielan:
> @tkielan Any comment on the fact that it failed for me with gcc?
> 
> ```
> /usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/crtbeginT.o: relocation R_X86_64_32 against hidden symbol `__TMC_END__' can not be used when making a shared object
> /usr/bin/ld: final link failed: Nonrepresentable section on output
> collect2: error: ld returned 1 exit status
> src/CMakeFiles/nats.dir/build.make:596: recipe for target 'src/libnatsd.so.2.5.0' failed
> ```
> 
> Do you think it is because my gcc is too old?

I have local reproduction on Ubuntu with gcc 10.3. It is definitely not too old.
Haven't seen it on Windows. I will investigate.

tkielan:
Building on WSL Ubuntu 20.04 LTS with GCC 9.3.0 works like a charm!
Going back to Linux testing but after working hours. Will update

kozlovic:
@tkielan Have you had a chance to investigate? Seem that so far it works only on Windows or WSL?

tkielan:
This change requires more work than I've originally thought in order to make it work on Linux 64.
I'm closing this PR.

PR #432, opened 2021-06-16
Fixed errors/warnings from LGTM search

kozlovic:
Note that the 2 errors were actually false positive, since the
intent was really to assign, not to compare.
This was a trick to assign the waitMode based on the comparaison,
but to avoid confusion went to a more standard approach.

Added the LGTM badges.


derekcollison (approved):
LGTM

PR #433, opened 2021-06-21
test_SimultaneousRequest  fix

chucongqing:
use dynamic thread number and others

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -4969,9 +4969,11 @@ _recvTestString(natsConnection *nc, natsSubscription *sub, natsMsg *msg,
         }
         case 4:
         {
-            arg->status = natsConnection_PublishString(nc,
+            arg->status = natsConnection_Publish(nc,
                                                        natsMsg_GetReply(msg),
-                                                       arg->string);
+                                                       natsMsg_GetData(msg),
+                                                       natsMsg_GetDataLength(msg));
+            arg->sum++;
             if (arg->status == NATS_OK)
                 arg->status = natsConnection_Flush(nc);
             break;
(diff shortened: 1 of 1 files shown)


test/test.c
```diff
@@ -4969,9 +4969,11 @@ _recvTestString(natsConnection *nc, natsSubscription *sub, natsMsg *msg,
         }
         case 4:
         {
-            arg->status = natsConnection_PublishString(nc,
+            arg->status = natsConnection_Publish(nc,
                                                        natsMsg_GetReply(msg),
-                                                       arg->string);
+                                                       natsMsg_GetData(msg),
+                                                       natsMsg_GetDataLength(msg));
+            arg->sum++;
```
kozlovic, line 4976: No, this is wrong. The arg->sum is incremented by each thread doing the request (in _sendRequest) if the result is as expected.

test/test.c
```diff
@@ -11051,7 +11053,7 @@ _sendRequest(void *closure)
 
     nats_Sleep(250);
 
-    s = natsConnection_RequestString(&msg, arg->nc, "foo", "Help!", 2000);
+    s = natsConnection_RequestString(&msg, arg->nc, "foo", arg->string, 2000);
```
kozlovic, line 11056: Did not have to change. The point of this test is to test parallel calls to Request() APIs. The request test here does not matter.

test/test.c
```diff
@@ -4969,9 +4969,11 @@ _recvTestString(natsConnection *nc, natsSubscription *sub, natsMsg *msg,
         }
         case 4:
         {
-            arg->status = natsConnection_PublishString(nc,
+            arg->status = natsConnection_Publish(nc,
                                                        natsMsg_GetReply(msg),
-                                                       arg->string);
+                                                       natsMsg_GetData(msg),
```
kozlovic, line 4974: This has to be `arg->string` since the requestor in _sendRequest() checks that the reply content is as expected (arg->string).

kozlovic (commented):
I am not sure what issue you were trying to solve with those changes.

kozlovic:
Closing for now. Will re-open if needed.

PR #434, opened 2021-07-01
[FIXED] Request APIs could return error and msg, causing leak.

kozlovic:
In the race condition where a reply comes as the RequestX() call
had timeout, a NATS_TIMEOUT error would be returned along with
the message. Callers may not attempt to destroy the reply message
if the request call error'ed.


derekcollison (approved):
LGTM

PR #435, opened 2021-07-01
Added nats[Str]Hash_RemoveSingle to remove the sole element of a hash

kozlovic:
In _respHandler, there is a case when we know the hash has a single
element and we want to use that one. We had to do something like:

```
natsStrHashIter_Init(&iter, nc->respMap);
natsStrHashIter_Next(&iter, NULL, &value);
natsStrHashIter_RemoveCurrent(&iter);
natsStrHashIter_Done(&iter);
```
(which, by the way, the "done" was missing).

This is now replaced with a single call:
```
natsStrHash_RemoveSingle(nc->respMap, NULL, &value);
```


derekcollison (approved):
LGTM

PR #440, opened 2021-07-21
[CHANGED] Set default library delivery pool size to 1 instead of 2

### LINKED ISSUE #439: Why is it not possible to use one thread for all subscribers (nats_SetMessageDeliveryPoolSize(1))?
I want to use single thread mode for all subscribers with additional options: nats_SetMessageDeliveryPoolSize(1) and natsOptions_UseGlobalMessageDelivery(TRUE).
But it is not possible because by implementation minimum allowed threads num is 2.

In function nats_Open() maximum workers pool size is set to 2:
...
gLib.dlvWorkers.maxSize = 2;
gLib.dlvWorkers.workers = NATS_CALLOC(gLib.dlvWorkers.maxSize, sizeof(natsMsgDlvWorker*));
...

With function nats_SetMessageDeliveryPoolSize() we unable to reduce threads number (mnimum allowed threads count is 2):

...
natsStatus
nats_SetMessageDeliveryPoolSize(int max)
...
    // Do not error on max < workers->maxSize in case we allow shrinking
    // the pool in the future.
    if (max > workers->maxSize)
    {
...

I didn't find any way to change this behavior.

I want to clarify why predefined max threads count = 2 is used and why we can't use single thread? 

Thank you.

kozlovic:
The initial size was set to 2 and since lowering the pool max size
is a no-op, one would not be able to have a pool of size 1.
Changed the default size to 1.

Resolves #439

### DIFF
--- a/src/nats.c
+++ b/src/nats.c
@@ -1040,7 +1040,7 @@ nats_Open(int64_t lockSpinCount)
             gLib.libDefaultWriteDeadline = (int64_t) atol(defaultWriteDeadlineStr);
 
         gLib.libHandlingMsgDeliveryByDefault = (getenv("NATS_DEFAULT_TO_LIB_MSG_DELIVERY") != NULL ? true : false);
-        gLib.dlvWorkers.maxSize = 2;
+        gLib.dlvWorkers.maxSize = 1;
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #442, opened 2021-07-21
[CHANGED] Shuffling of IPs after hostname resolution

kozlovic:
The default behavior will now be to randomize the list of IPs that
we get after hostname resolution before attempting to connect.

The shuffling is disabled if the `natsOptions_SetNoRandomize` option
is enabled.

See https://github.com/nats-io/nats-architecture-and-design/issues/23


derekcollison (approved):
LGTM

Issue #450, opened 2021-08-10
CMAKE_BUILD_TYPE is always overwritten


marascio:
The top level `CMakeLists.txt` will always overwrite `CMAKE_BUILD_TYPE`, setting it to `Release`, or whatever has been specified in `NATS_BUILD_TYPE`. I'm not sure why `NATS_BUILD_TYPE` is used to drive the build type selection, but in the cases where nats is being included via `add_subdirectory()` this behavior is undesirable.

Preferably, `CMAKE_BUILD_TYPE` would be set only if it wasn't already set. This might have been the intention, but by using variable type `INTERNAL` it is being force set.

kozlovic:
Close by #451

PR #451, opened 2021-08-10
[UPDATED] Removed NATS_BUILD_TYPE for CMAKE_BUILD_TYPE (which defaults to Release if not set)

marascio:
Using type INTERNAL when setting CMAKE_BUILD_TYPE will always set
CMAKE_BUILD_TYPE to "Release". This is because INTERNAL implies FORCE.
The type should be STRING to match the type of the variable that is
created when specifying CMAKE_BUILD_TYPE via the command line using
-DCMAKE_BUILD_TYPE=Release, for example.


marascio:
Fixes #450

kozlovic:
> Using type INTERNAL when setting CMAKE_BUILD_TYPE will always set CMAKE_BUILD_TYPE to "Release"

No, it will set CMAKE_BUILD_TYPE to NATS_BUILD_TYPE. Meaning that if you change NATS_BUILD_TYPE to say "Debug", that's what CMAKE_BUILD_TYPE will be set to. The reason for doing the way I did is because at the time I did not know how to set a default CMAKE_BUILD_TYPE to "Release". I want users that would fork/clone the repo and build the library to be by default built in Release mode. If you know a way to do so, I would like to know.

marascio:
The change I've proposed still does what you want, but with the code as it currently is users can not set the build type in the typical way that you do with CMake, which is by running cmake as: `cmake -DCMAKE_BUILD_TYPE=Debug ..`. If I do this, then the build type is set to `Release` because that is the default value of `NATS_BUILD_TYPE` and you currently set `CMAKE_BUILD_TYPE` as type `INTERNAL` which implies `FORCE`.

Currently, only this will work: `cmake -DNATS_BUILD_TYPE=Debug ..`. With my change, this continues to work, but the more typical method also now works.

marascio:
I will also mention that KitWare has a blog on how to set a default build type. I've adapted and used this in my work but it is not required for what you're trying to accomplish. https://blog.kitware.com/cmake-and-the-default-build-type/

kozlovic:
@marascio I love that!

I removed `NATS_BUILD_TYPE` and put that near the top of the main CMakeLists.txt:
```
# Set a default build type if none was specified
set(default_build_type "Release")

if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
  message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
  set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
      STRING "Choose the type of build." FORCE)
  # Set the possible values of build type for cmake-gui
  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
    "Release" "Debug" "MinSizeRel" "RelWithDebInfo")
endif()
```
seem to work very well!
If you don't mind updating your PR to do that and also modify `.travis.yml` to remove/replace references to `-DNATS_BUILD_TYPE` with `-DCMAKE_BUILD_TYPE` that would be great!

marascio:
Updated.

.travis.yml
```diff
@@ -20,7 +20,7 @@ before_script:
   - export COVERALLS_SERVICE_NAME=travis-ci
 
 env:
-  - DO_COVERAGE="coverage" BUILD_OPT="-DNATS_COVERAGE=ON -DNATS_BUILD_ARCH=64 -DNATS_BUILD_TYPE=Debug -DNATS_COVERAGE_UPLOAD=ON -DNATS_BUILD_TLS_USE_OPENSSL_1_1_API=ON -DCMAKE_BUILD_TYPE=Debug"
+  - DO_COVERAGE="coverage" BUILD_OPT="-DNATS_COVERAGE=ON -DNATS_BUILD_ARCH=64 -DCMAKE_BUILD_TYPE=Debug -DNATS_COVERAGE_UPLOAD=ON -DNATS_BUILD_TLS_USE_OPENSSL_1_1_API=ON -DCMAKE_BUILD_TYPE=Debug"
```
kozlovic, line 23: I had both `-DNATS_BUILD_TYPE` and `-DCMAKE_BUILD_TYPE` on that line (proof that was clearly not the right approach, but I did not know how to do..). Could you remove say the instance at the end of the line? Thanks

kozlovic (changes requested):

.travis.yml
marascio: Whoops, missed that. Sorry. Fixed.

kozlovic (approved):
LGTM. Thank you for the contribution! This is something that I did not like but when I started the project (and not familiar with CMake), I did not know how to do it and went with that. Again, thanks!

PR #458, opened 2021-08-23
JetStream: lot of changes

kozlovic:
This is similar to PR that was done for the Go client and in
preparation of the next NATS Server release.

- Added ability to call nats_GetLastError() in an asynchronous error callback to get a more informative error text condition
- JetStream:
    - Added `Domain` in `jsPubAck` object
    - Consumer is now deleted, if it was created by the "Subscribe" call, on "Unsubscribe" or when "Drain" completes
    - Added `PushBound` boolean in `jsConsumerInfo`. This tells if the consumer's is active in that the server has registered a subscription interest on the deliver subject. This requires server v2.3.5+
    - Added `DeliverGroup` string in the `jsConsumerConfig` object. This is used to let the server know that the consumer should be created for a given queue group
    - Subscribe calls now reject mismatch between non-queue and queue subscriptions depending if the consumer was created with a `DeliverGroup` or not and if the consumer is already bound or not. This requires server v2.3.5+.
    - Queue subscription is rejected if the consumer is configured with flow control
    - Queue name is used as the consumer durable name if no consumer name nor durable is provided (otherwise, the queue subscribe call would create a single member for an ephemeral consumer, which was confusing)
    - Added `jsSequenceInfo` that is similar to `jsSequencePair` but also contains `Last` which is the last activity time (UTC)
    - Changed `ConsumerInfo`'s `Delivered` and `AckFloor` from `jsSequencePair` to `jsSequenceInfo`
    - Library now handles heartbeat and flow control messages, regardless if the user configured those
    - Library will send Flow Control response if server thinks the consumer is stalled, that is, an heartbeat control message will have the header "Nats-Consumer-Stalled" set with a flow control subject


derekcollison (approved):
LGTM

PR #459, opened 2021-08-25
[IMPROVED] JS: Subscribe returns error if consumer config does not match

kozlovic:
This is done to alert the user that consumer config is not changed
as one would expect.

Also fixed setting optional start seq/time needs to set the
proper deliver policy.


derekcollison (approved):
LGTM

PR #462, opened 2021-09-03
[FIXED] natsSubscription_Fetch() could return immediately with a timeout error

kozlovic:
This would happen when pull requests would have filled the waiting
queue in the JetStream consumer and a 408 status was returned.


derekcollison (approved):
LGTM

PR #468, opened 2021-09-29
[ADDED] JetStream Ordered Consumers

kozlovic:
The new `Ordered` option in `jsSubOptions` allows the jetstream subscription to be created as an ordered consumer.
It means that the subscription will be based on a fifo ephemeral consumer for in order delivery of messages.
There are no redeliveries and no acks. Flow control and heartbeats are required and set by default, but the heartbeats value can be overridden in the consumer configuration.

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -1000,8 +1000,8 @@ _connectProto(natsConnection *nc, char **proto)
     return s;
 }
 
-static natsStatus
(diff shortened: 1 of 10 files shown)


derekcollison (approved):
LGTM

Check that the flowcontrol aspect has same "fix" as Go client to avoid stalls with first messages being large and processing being slow.

derekcollison (approved):
LGTM

PR #486, opened 2021-11-11
[UPDATED] JetStream examples

kozlovic:
- Made the created stream use the `-subj` as subject instead of
the stream name.
- Create stream (if not present) in the js-sub tool, not just
the js-pub tool.
- Add 1sec HB if flow control (`-fc`) parameter is set.


aricart (approved):
LGTM!

PR #493, opened 2021-12-07
[FIXED] JetStream: auto-ack callback ack handling

### LINKED ISSUE #492: JetStream: auto ack callback still acks message even if user called natsMsg_Nak(), etc..
Suppose that user creates a subscription with auto-ack mode, but then inside the callback calls natsMsg_Nack(), which expects the server to redeliver the message. Currently, the library will still send the ACK after the callback returns.
This will ack the message in the server, which it should not.

kozlovic:
If a message is already ack'ed (except for natsMsg_InProgress()),
the library should not attempt to ack the message when the
callback returns.
It is even more important if the user called natsMsg_Nak() since
if the callback ack'ed the message, it would defeat the purpose
of the Nak.

Resolves #492

### DIFF
--- a/examples/js-sub.c
+++ b/examples/js-sub.c
@@ -38,7 +38,7 @@ onMsg(natsConnection *nc, natsSubscription *sub, natsMsg *msg, void *closure)
     if (++count == total)
         elapsed = nats_Now() - start;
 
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

PR #495, opened 2021-12-08
[FIXED] JetStream: do not send acks for js_AckNone subscriptions

### LINKED ISSUE #494: JetStream: with "ack none" subscriptions, library should never send [N]ACKs
Since the server will never redeliver nor do anything with an ACK, there is no point in the library sending ACKs to the server. The ACK calls do not need to fail, simply not send anything to the server.

kozlovic:
When a subscription is for a consumer with js_AckNone AckPolicy,
there is no point sending ACK messages to the server since these
would be ignored.

Resolves #494

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2633,6 +2633,9 @@ natsConn_processMsg(natsConnection *nc, char *buf, int bufLen)
         {
             bool signal= false;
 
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

PR #496, opened 2021-12-16
Add BUILD_INTERFACE to target_include_directories

joejustesen:
Using FetchContent to pull in the nats.c libraries does not work
as the location for include directories is not set to also look in
the source directories.

Since I am not installing nats.c locally, this changes will allow me to use the build artifacts in my project.


kozlovic (approved):

PR #497, opened 2021-12-27
[FIXED] GCC warning on format strings

paoloteti:
Fixes following series of GCC warnings using `-Wformat-signedness/-Wformat`

`warning: format ‘%d’ expects argument of type ‘int’, but argument XX has type ‘unsigned int’ [-Wformat=]`


kozlovic (approved):
LGTM!

PR #498, opened 2021-12-29
[IMPROVED] Use isxdigit() instead of an explicit in-range check

paoloteti:
isxdigit() is equivalent, more compact and a bit faster.


kozlovic (approved):
LGTM!

PR #499, opened 2022-01-03
Fixed warning about uninitialized variable

kozlovic:
There was a warning regarding a variable that may be uninitialized.
This was a false positive in that we would not call nats_Base64_DecodeInPlace()
if nats_Base64_DecodeLen() did not return NATS_OK. When returning
OK, `sl` would be initialized. Still, easy to fix that by just
setting `sl` to 0 during declaration.


aricart (approved):
LGTM

PR #501, opened 2022-01-09
[IMPROVED] crypto: effectively zeroize mem. content

paoloteti:
When `libsodim` is not used, memory content is zeroized using a simple `memset()` and de-allocated using `free()`, so compiler may decide to remove `memset()`.

This patch add `secure_memzero()`, similar to `sodium_zero()`, to be sure to zeroize memory content also when `libsodium `is not used.


kozlovic (approved):
LGTM. Thanks again Paolo for your contributions!

PR #503, opened 2022-01-15
[FIXED] JetStream subscription using ">"

### LINKED ISSUE #502: subject wildcard ">"  does not work
Firstly,we has a durable stream with a subject "test.user", "nats-js-sub -subj test.*  -stream ...     -print   -s ..." can consume events normally,but "nats-js-sub -subj test.>  -stream ...     -print   -s ..." cant consume any event. its likely because js_GetConsumerInfo() get filterSubject with "u003e" instead of ">" , cause “”subject ... does not match consumer filter subject ...".

kozlovic:
The issue was actually caused by JSON parsing, so that could manifest
in other situations.
But for the reported issue, if a consumer was created with a filter
subject that contained the ">" wildcard and the user creates a
subscription with that same subject, the check for filter subject
match would fail because "test.>" is different than "test.\u003e".

Resolves #502

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -456,62 +456,100 @@ _jsonTrimSpace(char *ptr)
 }
 
 static natsStatus
+_decodeUni(char **iPtr, char *val)
+{
+    int     res = 0;
+    char    *i  = *iPtr;
+    int     j;
+
+    if (strlen(i) < 5)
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #504, opened 2022-01-16
[FIXED] test: fix assign instead of compare

paoloteti:
Since it is a test condition the intent here is to compare and not to assign a value

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -390,8 +390,8 @@ static void test_natsBuffer(void)
             && (strcmp(backend, "abcd") == 0));
 
     test("Changing backend is reflected in buffer: ");
+    backend[1] = 'x';
     testCond((s == NATS_OK)
-             && (backend[1] = 'x')
              && (natsBuf_Data(buf)[1] == 'x'));
 
     test("Append less than capacity does not expand buffer: ");


test/test.c
```diff
@@ -391,7 +391,7 @@ static void test_natsBuffer(void)
 
     test("Changing backend is reflected in buffer: ");
     testCond((s == NATS_OK)
-             && (backend[1] = 'x')
+             && (backend[1] == 'x')
```
kozlovic, line 394: No, the intent was actually, as the text in `test()` above suggests, to change the backend content and verify that it is reflected in the natsBuffer object.
But I agree that some compilers may flag this as suspicious, so you could move setting `backend[1] = 'x';` above the `testCond()` to make that clearer.

PS: technically the `(s == NATS_OK)` is not required here. This is the remaining of original test code that would not exit the function on a failed `testCond()`. Since I do now, there are cases where checking for `s == NATS_OK` is not required, but it does not hurt, so I would not go over all test.c to remove the ones that are not needed :-).

kozlovic (changes requested):

kozlovic (approved):
LGTM. Thanks Paolo!

PR #505, opened 2022-01-16
[FIXED] nuid: fix wrong sizeof in memset

paoloteti:
Fixes following Coverity issue:

Passing argument &globalNUID of type natsLockedNUID * and
argument 32UL (sizeof (natsNUID)) to function memset is suspicious
because sizeof (natsLockedNUID) /*40*/ is expected.

### DIFF
--- a/src/nuid.c
+++ b/src/nuid.c
@@ -208,7 +208,7 @@ natsNUID_init(void)
     natsStatus      s;
     unsigned int    seed = (unsigned int) nats_NowInNanoSeconds();
 
-    memset(&globalNUID, 0, sizeof(natsNUID));
+    memset(&globalNUID, 0, sizeof(natsLockedNUID));
 
     srand(seed);
     _initCMWC(seed);


src/nuid.c
```diff
@@ -208,7 +208,7 @@ natsNUID_init(void)
     natsStatus      s;
     unsigned int    seed = (unsigned int) nats_NowInNanoSeconds();
 
-    memset(&globalNUID, 0, sizeof(natsNUID));
+    memset(&(globalNUID.nuid), 0, sizeof(globalNUID.nuid));
```
kozlovic, line 211: Good catch in that the memset was wrong, but it was the `sizeof()` that was wrong, not the address of the object. Here, I wanted to initialize `globalNUID`, which is a `natsLockedNUID` object, which itself, is a pointer to a mutex and the natsNUID structure. So the proper initialization is:
```
memset(&globalNUID, 0, sizeof(natsLockedNUID));
```

kozlovic (changes requested):

src/nuid.c
paoloteti: I’ll update, but mutex is initialized one line after, so I used memset() in this way.

kozlovic: But that was the point of the memset: if the initialization of the mutex fails, calling natsNUID_free() in case of error would then be dangerous because `globalNUID.mu` pointer may not have been initialized to NULL.

kozlovic (approved):
LGTM. Thanks again for your continuous contributions!

PR #507, opened 2022-01-22
Set return code on _connect() correctly

npes87184:
As s is not possible to be NAST_OK here.
Directly set s to NATS_CONNECTION_CLOSED.

Signed-off-by: yuchenlin <yuchenlin@synology.com>


kozlovic (approved):

PR #508, opened 2022-01-22
[FIXED] Coverity Scan unchecked return value

paoloteti:
Propagate error code to fix two unchecked return value issues reported by Coverity Scan.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -1967,9 +1967,10 @@ nats_ReadFile(natsBuffer **buffer, int initBufSize, const char *fn)
     else if (fclose(f) != 0)
         s = nats_setError(NATS_ERR, "error closing file '%s': '%s", fn, strerror(errno));
 
+    IFOK(s, natsBuf_AppendByte(buf, '\0'));
+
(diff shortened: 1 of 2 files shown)


src/util.c
```diff
@@ -1969,7 +1969,7 @@ nats_ReadFile(natsBuffer **buffer, int initBufSize, const char *fn)
 
     if (s == NATS_OK)
     {
-        natsBuf_AppendByte(buf, '\0');
+        s = natsBuf_AppendByte(buf, '\0');
```
kozlovic, line 1972: Thanks for the catch. However, if we do get the status, then we need to return buffer only if the last append is OK, otherwise, we need to destroy the buffer. So we could replace with (replacing the above `if (s == NATS_OK)`):
```
IFOK(s, natsBuf_AppendByte(buf, '\0'));

if (s == NATS_OK)
{
    *buffer = buf;
}
else if (buf != NULL)
{
...
```

kozlovic (changes requested):

PR #510, opened 2022-01-26
[FIXED] KV: Support for JetStream prefix

kozlovic:
When the JetStream context has a prefix, will use that for all
KV APIs as specified by ADR #19


aricart (approved):
LGTM

PR #511, opened 2022-01-27
[FIXED] JetStream: Fetch() could take longer than the given timeout

kozlovic:
The timeout was not re-evaluated when going back to low-level
natsSubscription_NextMsg().

Also fixed code and tests so that this works for all server versions
(2.6.6, 2.7.0 and current 2.7.1).


aricart (approved):
LGTM

PR #513, opened 2022-01-29
packaging: added rpm spec file

sergey-safarov:


packaging/rpm/libnats.spec
```diff
@@ -0,0 +1,55 @@
+%global source_name nats.c
+%define debug_package %{nil}
+
+Name:          libnats
+Version:       3.1.1
```
kozlovic, line 5: We are currently at version 3.2.0. Also, would be nice if that could be updated when generating version, otherwise it will need to be manually updated before I do a new release..

packaging/rpm/libnats.spec
```diff
@@ -0,0 +1,55 @@
...
+%{cmake_install}
+find %{buildroot} -name '*.a' -delete
+
+%files
+%{_libdir}/%{name}.so*
+
+%files devel
+%dir %{_includedir}/nats
+%dir %{_includedir}/nats/adapters
+%{_includedir}/nats.h
```
kozlovic, line 47: Note that we don't install nats.h in the root of include directory, instead it is installed under `nats` directory inside the include directory. We used to put it in top-level include directory, but we no longer do.
To be precise, we put under the include directory the file `deprnats.h` that is renamed `nats.h`, to indicate that users should use `nats/nats.h` instead.
So maybe we should not add it here? (unless I completely misunderstand what this is doing).

packaging/rpm/libnats.spec
```diff
@@ -0,0 +1,55 @@
...
+%{_includedir}/nats/*.h
+%{_includedir}/nats/adapters/*.h
+%{_libdir}/pkgconfig/%{name}.pc
+%{_prefix}/lib/cmake/cnats/*.cmake
+
+%changelog
+* Sat Nov 06 2021 Sergey Safarov <s.safarov@gmail.com> 3.1.1
```
kozlovic, line 54: Maybe refresh the date here and change the version?

kozlovic (changes requested):

PR #514, opened 2022-01-31
Better validate subject helper

kozlovic:
Also changed the custom inbox prefix option setter to make "." as
the last character invalid (to align with Go client).


aricart (approved):
LGTM

PR #517, opened 2022-02-03
[ADDED] JetStream: Ability to get subjects list from stream info

kozlovic:


kozlovic:
@aricart @derekcollison We discussed that this was not a priority, but I did start yesterday. We can keep the PR for a while if you think we should let that sit longer...

derekcollison (approved):
LGTM

aricart (approved):
LGTM

PR #518, opened 2022-02-10
Fixed InactiveThreshold consumer configuration handling

kozlovic:
The field was not (un)marshal'ed and the test that was supposed
to check the feature was actually checking the opposite, that is,
that the consumer was found instead of not found !!!

Related to #515


aricart (approved):
LGTM

PR #523, opened 2022-02-16
KV add kvStore_Update to the cross account test.

kozlovic:
This makes sure that "Update" also uses the prefix when needed.


aricart (approved):
LGTM

PR #524, opened 2022-02-16
[CHANGED] PurgeDeletes() will now keep markers that are less than 30min old

kozlovic:
There is a breaking change where kvStore_PurgeDeletes() accepts now
kvPurgeOptions instead of kvWatchOptions.

We needed from kvWatchOptions only the timeout, and as it standed, it was
bad since user could have passed watcher options that would have
affected the internal watcher that we make to collect delete markers.

Also, when invoking kvStore_PurgeDeletes(), the delete markers that
are older than a default of 30 minutes will be deleted, however more
recent ones will be kept. The data is always removed, even if a marker
is not.

The user can change the 30 minutes threshold using a new purge
option called kvPurgeOptions.DeleteMarkersOlderThan. If set to -1,
it restores the old behavior of deleting delete/purge markers,
regardless of their age.

### DIFF
--- a/src/kv.c
+++ b/src/kv.c
@@ -556,10 +556,12 @@ kvStore_UpdateString(uint64_t *rev, kvStore *kv, const char *key, const char *da
 }
 
 static natsStatus
-_delete(kvStore *kv, const char *key, bool purge)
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

PR #525, opened 2022-02-16
[FIXED] cmake files install path

linuxmaniac:
```
-rw-r--r-- root/root      1341 2022-02-16 09:16 ./usr/lib/x86_64-linux-gnu/cmake/cnats/cnats-config-none.cmake
-rw-r--r-- root/root      4356 2022-02-16 09:16 ./usr/lib/x86_64-linux-gnu/cmake/cnats/cnats-config.cmake
```


kozlovic (approved):

PR #527, opened 2022-02-25
[CHANGED] natsOptions_SetCertificatesChain() now not limited to RSA key

### LINKED ISSUE #526: natsOptions_SetCertificatesChain does only accept RSA keys
natsOptions_SetCertificatesChain only accept RSA keys. In a attempt at making things a bit safer, we are using EC keys.

The current workaround is to write those to files and load them with natsOptions_LoadCertificatesChain.

kozlovic:
This option used RSA specific OpenSSL APIs to load an RSA key.
This change will use any type of key.

Resolves #526

### DIFF
--- a/src/opts.c
+++ b/src/opts.c
@@ -552,24 +552,25 @@ natsOptions_SetCertificatesChain(natsOptions *opts, const char *certStr, const c
     }
     if (s == NATS_OK)
     {
-        RSA *rsa  = NULL;
-        BIO *bio  = NULL;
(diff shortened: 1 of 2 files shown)


thierryba:
I'm the reporter of #526 and this works just fine.

kozlovic:
@thierryba Thank you for confirming that this change fixes your issue. Merging now.

aricart (approved):
LGTM

thierryba:
Thanks a lot.

PR #528, opened 2022-03-01
[CHANGED] JetStream: do not set MaxAckPending if not user specified

kozlovic:
If user does not specify MaxAckPending, don't set any value and
let the server pick the default.
Ensure that once we get the info back, we set the subscription
pending limits to at least cover this number.


aricart (approved):
LGTM

PR #529, opened 2022-03-01
JetStream: auto-ack callback would do sync acks

kozlovic:
This is due to a change in PR #516 that is making any Ack call
that passes an option to transform the call to sync calls if
option has a Wait value set.
The auto-ack callback was calling ack with the jetstream's context
options, which has always a Wait by default of 5sec.


aricart (approved):
LGTM

PR #531, opened 2022-03-05
[FIXED] Connection fails if URL is missing port

### LINKED ISSUE #530: connect fail with domain name without port 4222
i try connect with my nast server with domain name with url:nats://user-id:user-password@nats.myinspurcloud.com, but get "No server available for connection", then i find i must specify the url:nats://user-id:user-password@nats.myinspurcloud.com:4222.
  use nats cli-took(nats-0.0.28-amd64.deb), i can connect this server directly with "-s nats://user-id:user-password@nats.myinspurcloud.com", i think the default port should be added into url parse.

kozlovic:
With missing port, we should default to 4222.

Resolves #530

### DIFF
--- a/src/natsp.h
+++ b/src/natsp.h
@@ -98,8 +98,6 @@
 #define WAIT_FOR_WRITE      (1)
 #define WAIT_FOR_CONNECT    (2)
 
-#define DEFAULT_PORT_STRING "4222"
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #537, opened 2022-04-04
[FIXED] JetStream: stream and/or consumer name checks

kozlovic:
Some of the APIs were not properly checking for missing/invalid
stream and/or consumer names (presence of ".") which could lead
to API calls timing out because the server had no subscription
interest on the formed subject.

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -1884,7 +1884,7 @@ natsStatus
 js_checkDurName(const char *dur)
 {
     if (strchr(dur, '.') != NULL)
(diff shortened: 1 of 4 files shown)


kozlovic:
@aricart @derekcollison Please wait for the review. I will minimize the changes (can have a special function for non durables), but most importantly, I think that there are other APIs that accept a consumer name that should check on the name (like delete). Will ping when PR has been updated and is ready for review.

kozlovic:
@aricart @derekcollison Ok, PR has been updated and is now ready for review.

derekcollison (approved):
LGTM

aricart:
LGTM

PR #543, opened 2022-05-20
[ADDED] JetStream: Support for consumer config MaxRequestMaxBytes

kozlovic:
This is to limit the size of a pull consumer request.

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

PR #546, opened 2022-05-25
[ADDED] Stream RePublish and some ConsumerConfig new fields

kozlovic:
Namely for pull consumers: ability to override the replica count or storage type:
```
// Generally inherited by parent stream and other markers, now can be configured directly.
Replicas int `json:"num_replicas"`
// Force memory storage.
MemoryStorage bool `json:"mem_storage,omitempty"`
```

For the stream, this new StreamConfig option:
```
// Allow republish of the message after being sequenced and stored.
RePublish *SubjectMapping `json:"republish,omitempty"`
```
Where SubjectMapping is:
```
// SubjectMapping allows a source subject to be mapped to a destination subject for republishing.
type SubjectMapping struct {
	Source      string `json:"src,omitempty"`
	Destination string `json:"dest"`
}
```

Note: Removed consumer config's MaxRequestMaxBytes that was added
in PR #543 until we settle how max_bytes in pull requests should behave
(both in client and server)

### DIFF
--- a/src/jsm.c
+++ b/src/jsm.c
@@ -97,6 +97,17 @@ _destroyStreamSource(jsStreamSource *source)
     NATS_FREE(source);
 }
 
+static void
+_destroySubjectMapping(jsSubjectMapping *sm)
+{
+    if (sm == NULL)
+        return;
+
+    NATS_FREE((char*) sm->Source);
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #551, opened 2022-07-06
[ADDED] jsPubAckHandler to handle "positive" publish acks

### LINKED ISSUE #538: Enhancements to js_PublishAsync
```
/** \brief Publishes data to JetStream but does not wait for a #jsPubAck.
 *
 * See #js_PublishMsgAsync for details.
 *
 * @param js the pointer to the #jsCtx object.
 * @param subj the subject the data is sent to.
 * @param data the data to be sent, can be `NULL`.
 * @param dataLen the length of the data to be sent.
 * @param opts the publish options, possibly `NULL`.
 */
NATS_EXTERN natsStatus
js_PublishAsync(jsCtx *js, const char *subj, const void *data, int dataLen,
                jsPubOptions *opts);
```
The current signature of js_PublishAsync does not provide a mechanism for a positive ack and does not call the  jsOpts.PublishAsync.ErrHandler in the following scenario
```
>> natsConnection_Connect
>> natsConnection_JetStream
>> sleep (shutdown the server from another shell)
>> js_PublishAsync return NATS_OK
>> jsOpts.PublishAsync.ErrHandler does NOT get called
```
This issue will track the changes needed to support the following requests:

1) Call jsOpts.PublishAsync.ErrHandler on timeout, the timeout can be set through jsPubOptions.MaxWait
2) Provide parity with Go PublishAsync API by supporting positive ack (along the lines of stanConnection_PublishAsync)

```
/* 
 * 's' can be NATS_OK, NATS_TIMEOUT or any other error due to async publish.
 * 'msg' handle, it is apps responsibility to free the msg
 */
typedef void (*jsPubAckHandler)(natsStatus s, natsMsg *msg, void *closure);
 
/* 
 * jsPubOptions.MaxWait can be used to control the timeout for the ack handler
 * 'ah' is called on success, async publish error or timeout
 */
NATS_EXTERN natsStatus
js_PublishAsyncAck(jsCtx *js, const char *subj, const void *data, int dataLen,
                jsPubOptions *opts, jsPubAckHandler ah, void *ahClosure);
```
Thanks for the help !!

kozlovic:
[FIXED] Ability to have async publish call timeout and invoke err handler

The new jsPubAckHandler takes precedence and is invoked even on
successful publish.

If the option MaxWait is specified on publish, then the library
will invoke the ack handler (or err handler) if a publish ack
is not received by then.

Resolves #538

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -81,6 +81,7 @@ _freeContext(jsCtx *js)
     NATS_FREE(js->rpre);
     natsCondition_Destroy(js->cond);
     natsMutex_Destroy(js->mu);
+    natsTimer_Destroy(js->pmtmr);
     nc = js->nc;
     NATS_FREE(js);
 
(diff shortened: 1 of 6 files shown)


aricart (approved):
LGTM

PR #556, opened 2022-07-12
fix armel build

linuxmaniac:
> Error: selected processor does not support `yield' in ARM mode

### DIFF
--- a/src/unix/mutex.c
+++ b/src/unix/mutex.c
@@ -78,7 +78,7 @@ natsMutex_Lock(natsMutex *m)
                 #if defined(__x86_64__) || \
                     defined(__mips__)
                     __asm__ __volatile__ ("pause" ::: "memory");
-                #elif defined(__arm__) || \
+                #elif (defined(__arm__) && __ARM_ARCH >=6) || \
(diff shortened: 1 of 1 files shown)


kozlovic:
@linuxmaniac Thank you for the contribution. I noticed from the build logs that there was a warning for a snprintf that I just fixed (https://github.com/nats-io/nats.c/pull/557). Also, notice from your log that you are using OpenSSL 1.1+ and so should add `-DNATS_BUILD_TLS_USE_OPENSSL_1_1_API=ON` in your cmake config (at one point, I should make this the default since OpenSSL 1.1 is now very common).

kozlovic (approved):
LGTM! Please also check my comment about OpenSSL and the fix for the snprintf warning.

linuxmaniac:
@kozlovic JFYI applied your patch for #557 and added ``-DNATS_BUILD_TLS_USE_OPENSSL_1_1_API=ON`` at the last version
https://buildd.debian.org/status/fetch.php?pkg=nats.c&arch=amd64&ver=3.3.0-4&stamp=1657760618&raw=0

PR #557, opened 2022-07-12
[FIXED] Compiler warning for JSON get time

kozlovic:
When formatting a date/time, the max lenght is 35 characters, so
the backing array should be 36 at least (for the terminal '\0').


aricart (approved):
LGTM

PR #559, opened 2022-07-15
[FIXED] KeyValue: set watcher pending limit to unlimited

kozlovic:
Since internal subscription is not currently exposed, setting
the subscription limit to unlimited.


aricart (approved):
Lgtm

PR #562, opened 2022-07-20
[ADDED] JetStream: Direct Get Message

kozlovic:
This starts with a new boolean in a stream configuration:
```
// Allow higher performance, direct access to get individual messages. E.g. KeyValue
bool  AllowDirect;
```

When a stream is created with that boolean, then it is possible
to use this API to get messages:
```
natsStatus
js_DirectGetMsg(natsMsg **msg, jsCtx *js, const char *stream, jsOptions *opts, jsDirectGetMsgOptions *dgOpts);
```

The possible options to get a message are:
```
typedef struct jsDirectGetMsgOptions
{
        uint64_t        Sequence;               ///< Get the message at this sequence
        const char      *NextBySubject;         ///< Get the next message (based on sequence) for that subject
        const char      *LastBySubject;         ///< Get the last message on that subject

} jsDirectGetMsgOptions;
```

### DIFF
--- a/src/js.h
+++ b/src/js.h
@@ -144,6 +144,9 @@ extern const int64_t    jsDefaultRequestWait;
 // jsApiMsgGetT is the endpoint to get a message, either by sequence or last per subject.
 #define jsApiMsgGetT "%.*s.STREAM.MSG.GET.%s"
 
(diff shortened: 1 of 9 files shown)


derekcollison (commented):
LGTM

PR #564, opened 2022-07-22
[UPDATED] JetStream: consumer configuration check

kozlovic:
New fields have been added in the consumer configuration, but the
function that checks that the existing consumer configuration matches
the configuration of a consumer that is looked-up duing a "subscribe"
call was not up-to-date.


derekcollison (approved):
LGTM

PR #565, opened 2022-07-26
[ADDED] natsSubscription_GetID() and natsSubscription_GetSubject()

hided:
This change adds the ability to get the subscription id / subject from a given natsSubscription instance.

This is done in a similar manner as the other "get functions" defined in nats.h (e.g. natsSubscription_GetStats which populates various fields such as pending bytes and dropped messages of a given subscription)

### DIFF
--- a/src/nats.h
+++ b/src/nats.h
@@ -4564,16 +4564,27 @@ natsSubscription_AutoUnsubscribe(natsSubscription *sub, int max);
 NATS_EXTERN natsStatus
 natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs);
 
-/** \brief Returns the subscription id and subject name.
+/** \brief Gets the subscription id.
  *
- * Returns the subscription id and subject name of this subscription.
+ * Returns the id of the given subscription.
  *
  * @param sub the pointer to the #natsSubscription object.
- * @param sid if not `NULL`, the memory location where to store the subscription id.
+ * @param id if not `NULL`, the memory location where to store the subscription id.
+ */
+NATS_EXTERN natsStatus
+natsSubscription_GetID(natsSubscription* sub, int64_t* id);
+
+/** \brief Gets the subject name.
+ *
+ * Returns the subject of the given subscription.
+ *
+ * \warning The string belongs to the subscription and must not be freed. Copy it if needed.
+ *
+ * @param sub the pointer to the #natsSubscription object.
  * @param subject if not `NULL`, the memory location where to store the subject name.
  */
 NATS_EXTERN natsStatus
-natsSubscription_GetInfo(natsSubscription* sub, int64_t* sid, char* subject);
+natsSubscription_GetSubject(natsSubscription* sub, const char* subject);
 
 /** \brief Sets the limit for pending messages and bytes.
  *
--- a/src/nats.h
+++ b/src/nats.h
@@ -4567,24 +4567,26 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs);
 /** \brief Gets the subscription id.
  *
  * Returns the id of the given subscription.
+ * 
+ * \note Invalid or closed subscriptions will cause a value of 0 to be returned.
  *
  * @param sub the pointer to the #natsSubscription object.
- * @param id if not `NULL`, the memory location where to store the subscription id.
  */
-NATS_EXTERN natsStatus
-natsSubscription_GetID(natsSubscription* sub, int64_t* id);
+NATS_EXTERN int64_t
+natsSubscription_GetID(natsSubscription* sub);
 
 /** \brief Gets the subject name.
  *
  * Returns the subject of the given subscription.
  *
+ * \note Invalid or closed subscriptions will cause a value of NULL to be returned.
+ *
  * \warning The string belongs to the subscription and must not be freed. Copy it if needed.
  *
  * @param sub the pointer to the #natsSubscription object.
- * @param subject if not `NULL`, the memory location where to store the subject name.
  */
-NATS_EXTERN natsStatus
-natsSubscription_GetSubject(natsSubscription* sub, const char* subject);
+NATS_EXTERN const char*
+natsSubscription_GetSubject(natsSubscription* sub);
 
 /** \brief Sets the limit for pending messages and bytes.
  *
--- a/src/nats.h
+++ b/src/nats.h
@@ -4564,6 +4564,17 @@ natsSubscription_AutoUnsubscribe(natsSubscription *sub, int max);
 NATS_EXTERN natsStatus
 natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs);
 
+/** \brief Returns the subscription id and subject name.
+ *
+ * Returns the subscription id and subject name of this subscription.
+ *
+ * @param sub the pointer to the #natsSubscription object.
+ * @param sid if not `NULL`, the memory location where to store the subscription id.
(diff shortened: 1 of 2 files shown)


src/nats.h
```diff
@@ -4564,6 +4564,17 @@ natsSubscription_AutoUnsubscribe(natsSubscription *sub, int max);
...
+/** \brief Returns the subscription id and subject name.
+ *
+ * Returns the subscription id and subject name of this subscription.
+ *
+ * @param sub the pointer to the #natsSubscription object.
+ * @param sid if not `NULL`, the memory location where to store the subscription id.
+ * @param subject if not `NULL`, the memory location where to store the subject name.
+ */
+NATS_EXTERN natsStatus
+natsSubscription_GetInfo(natsSubscription* sub, int64_t* sid, char* subject);
```
kozlovic, line 4576: Thank you for adding the doc! Since it looks like you are returning the subject from the subscription object, and not a copy, I usually make that a `const char*` and add a `\warning` in the doc to indicate that the returned string belongs to the subscription object and the user must not free this memory.
Here is an example for natsMsg_GetSubject: https://nats-io.github.io/nats.c/group__msg_group.html#gad7cad38946648e3047aa0665a3dd230a

src/sub.c
```diff
@@ -1124,6 +1124,35 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs)
     return s;
 }
 
+natsStatus
+natsSubscription_GetInfo(natsSubscription* sub, int64_t* sid, char* subject)
```
kozlovic, line 1128: Although I like reducing the number of APIs, what if some day we think that we should return some more information from the subscription? If we think that may happen, then we should here have a GetSubject() and GetID() right out of the get go.

kozlovic (changes requested):
Thank you for the contribution! Some remarks, let me know what you think. Adding a test also would be great, but it is a bit more involved and I will add one (if needed) after we merge.

src/nats.h
hided: Valid remark. Changed to const and added the warning in the docs.

src/sub.c
hided: I was also debating whether to split this into 2 separate API's or not. And it is indeed a bit random that these 2 unrelated fields are returned together.

I split them as you suggested.

hided:
> Adding a test also would be great, but it is a bit more involved and I will add one (if needed) after we merge.

I did have a look at the test file but as you might be able to tell i'm quite unfamiliar with C. I don't have a clue how to actually run/debug the tests so I hope you don't mind I didn't provide you with a test.

src/nats.h
```diff
@@ -4564,6 +4564,28 @@ natsSubscription_AutoUnsubscribe(natsSubscription *sub, int max);
 NATS_EXTERN natsStatus
 natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs);
 
+/** \brief Gets the subscription id.
+ *
+ * Returns the id of the given subscription.
+ *
+ * @param sub the pointer to the #natsSubscription object.
+ * @param id if not `NULL`, the memory location where to store the subscription id.
```
kozlovic, line 4572: But now that we make individual getters, it is a bit silly to support `NULL` for the return location. That is, why would one call this function without wanting to retrieve the id?

src/nats.h
```diff
@@ -4564,6 +4564,28 @@ natsSubscription_AutoUnsubscribe(natsSubscription *sub, int max);
...
+natsSubscription_GetID(natsSubscription* sub, int64_t* id);
+
+/** \brief Gets the subject name.
+ *
+ * Returns the subject of the given subscription.
+ *
+ * \warning The string belongs to the subscription and must not be freed. Copy it if needed.
+ *
+ * @param sub the pointer to the #natsSubscription object.
+ * @param subject if not `NULL`, the memory location where to store the subject name.
```
kozlovic, line 4584: Same comment that for the ID. Now, as a convention, we could say that those getter return "invalid" values that the user would have to interpret as the subscription being closed/invalid. For instance, natsSubscription_GetID() could be returning a int64_t (instead of status), and assume that 0 means invalid/closed. And natsSubscription_GetSubject() could return NULL to indicate invalid/closed subscription (since empty subject is not valid).

src/nats.h
```diff
@@ -4564,6 +4564,28 @@ natsSubscription_AutoUnsubscribe(natsSubscription *sub, int max);
...
+ */
+NATS_EXTERN natsStatus
+natsSubscription_GetSubject(natsSubscription* sub, const char* subject);
```
kozlovic, line 4587: If we keep return status and return value through "out" parameter, then you need to have `const char **subject` here.

src/sub.c
```diff
@@ -1124,6 +1124,58 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs)
...
+
+    SUB_DLV_WORKER_UNLOCK(sub);
+
+    natsSub_Unlock(sub);
+
+    return NATS_OK;
+}
+
+natsStatus
+natsSubscription_GetSubject(natsSubscription* sub, const char* subject)
```
kozlovic, line 1154: See prior comment.

src/sub.c
```diff
@@ -1124,6 +1124,58 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs)
...
+    if (sub->closed)
+    {
+        natsSub_Unlock(sub);
+        return nats_setDefaultError(NATS_INVALID_SUBSCRIPTION);
+    }
+
+    SUB_DLV_WORKER_LOCK(sub);
+
+    if (subject != NULL)
+        subject = (const char*)sub->subject;
```
kozlovic, line 1170: Would be `*subject = (const char*) sub->subject;`

kozlovic (changes requested):
I can add a test later. Check my comments about what we return. Not sure which approach is best, but for instance natsMsg_GetSubject() returns NULL is message is NULL, we could do the same.

src/sub.c
```diff
@@ -1124,6 +1124,54 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs)
...
+
+    natsSub_Lock(sub);
+
+    if (sub->closed)
+    {
+        natsSub_Unlock(sub);
+        return 0;
+    }
+
+    int64_t id = 0;
```
kozlovic, line 1141: I prefer that we declare this at the top of the function (to be consistent with rest of the code).

src/sub.c
```diff
@@ -1124,6 +1124,54 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs)
...
+    SUB_DLV_WORKER_LOCK(sub);
```
kozlovic, line 1142: I missed that in earlier PRs, but for both these APIs, you don't need to call SUB_DLV_WORKER_LOCK, so let's remove.

src/sub.c
```diff
@@ -1124,6 +1124,54 @@ natsSubscription_QueuedMsgs(natsSubscription *sub, uint64_t *queuedMsgs)
...
+
+    natsSub_Lock(sub);
+
+    if (sub->closed)
+    {
+        natsSub_Unlock(sub);
+        return NULL;
+    }
+
+    const char* subject = NULL;
```
kozlovic, line 1165: Same, please declare at top, and remove SUB_DLV_WORKER_LOCK/SUB_DLV_WORKER_UNLOCK

kozlovic (changes requested):
Almost there! Sorry for all those back and forth.

kozlovic (approved):

PR #567, opened 2022-07-29
[CHANGED] `const int` to `int` for `Count` field in lists

### LINKED ISSUE #545: C++: natsMsgList default constructor is a deleted function
The following line of code fails to compile using VS2022 (Win SDK 8.1)

```cplusplus
natsMsgList list;
```
with error: 
natsMsgList::natsMsgList(void)': attempting to reference a deleted function

'natsMsgList::natsMsgList(void)': function was implicitly deleted because 'natsMsgList' has an uninitialized const-qualified data member 'natsMsgList::Count'

Any thoughts on how to fix this?

kozlovic:
This was causing compilation errors in C++ and was not really
adding any protection. It is stated in the doc that this value
should not be changed by the user prior to calling the destroy
function for those various lists.

Resolves #545

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -1147,7 +1147,7 @@ js_PublishAsyncGetPendingList(natsMsgList *pending, jsCtx *js)
             if (js->pmcount > 0)
                 js->pmcount--;
         }
-        *(int*)&(pending->Count) = count;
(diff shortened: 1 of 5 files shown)


aricart (approved):
LGTM

PR #571, opened 2022-08-11
Adjust to server direct get message time formatting

kozlovic:
The direct get message time formatting was different that the regular
message get, but now they are the same so removed code that was handling
that special case.


aricart (approved):
LGTM

PR #578, opened 2022-08-29
[ADDED] JetStream: support for ephemeral pull subscription.

kozlovic:
If the durable name is not specified, the library will create an
ephemeral pull subscription.


aricart (approved):
LGTM

PR #580, opened 2022-09-13
[ADDED] `Name` in consumer configuration

kozlovic:
When this is used, the library will send request to a different API subject space where the stream and consumer names are in the subject, along with a filter subject if one is specified.

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -2128,10 +2128,11 @@ _processConsInfo(const char **dlvSubject, jsConsumerInfo *info, jsConsumerConfig
 }
 
 natsStatus
-js_checkDurName(const char *dur)
+js_checkConsName(const char *cons, bool isDurable)
 {
-    if (strchr(dur, '.') != NULL)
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

kozlovic:
Related to https://github.com/nats-io/nats-architecture-and-design/issues/148

@derekcollison I went with the approach that as long as the new consumer config param "Name" is specified, then the library uses the new subject space, that is, I don't check on server version. The worst case is that the request will timeout. Do you think that checking the server version is required? (of course, always the question: what if client connects to server that is pre 2.9.0 while the server that would handle the request is 2.9.0+ and would be able, but I know that we should have NATS network at the same version level, so I could make the check if you think this is required)

derekcollison:
I don't think version check required imo..

aricart (approved):
LGTM - the thing I notice, is for new API clients need to validate the durable name to be something acceptable for embedding, because otherwise it will collide with the filter part of the subject.

PR #582, opened 2022-09-16
[FIXED] JetStream: natsSubscription_Destroy should not delete JS consumer

kozlovic:
Added comments in the js_Subscribe() calls that explains the behavior of auto created/deleted JS consumers.
It was the intent that auto-created JS consumers would be deleted only when explicitly calling unsubscribe/drain, similar to Go client.

However, in C, the user needs to call natsSubscription_Destroy to free memory. This call is calling internally natsSubscription_Unsubscribe (if the call was not made explicitly), but then we need to disable the deletion of the JS consumer in that case.

### DIFF
--- a/src/nats.h
+++ b/src/nats.h
@@ -5884,6 +5884,37 @@ jsSubOptions_Init(jsSubOptions *opts);
 
 /** \brief Create an asynchronous subscription.
  *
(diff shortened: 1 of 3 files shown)


aricart (approved):
LGTM

PR #583, opened 2022-09-17
[CHANGED] Count message headers toward subscription limits

kozlovic:
When headers were introduced, the whole buffer (so data + hdr) was added as pending limits, but when messages were removed from the subscription, only the data part was removed, which caused an issue that was address in PR #420 by just accounting for the data size (not header).

Headers becoming more and more prevalent, it could be that a big part of message content is the headers, so we should probably count the headers size along with the data toward the subscription's bytes limit.


derekcollison (approved):
LGTM

PR #585, opened 2022-09-19
[FIXED] Durable/Consumer name checks

### LINKED ISSUE #581: Consumer/Durable name validation
Should exclude ".", " ", "*" and ">".

kozlovic:
We need to reject names that contain '.', ' ', '*' or '>'.

Resolves #581


aricart (approved):
LGTM

aricart:
Scott had said that providing other kinds of characters were rejected by the server, but I have not verified the functionality beyond the `/` being rejected.

kozlovic:
@aricart This is fine, the check here is so that the requests are not malformed (or not received by the server due to space, etc..). The actual validation can still be made by the server.

PR #586, opened 2022-09-19
[UPDATED] JetStream: Handle Tier limits in jsAccountInfo

kozlovic:
Related to https://github.com/nats-io/nats-architecture-and-design/issues/120

### DIFF
--- a/src/jsm.c
+++ b/src/jsm.c
@@ -1718,6 +1718,75 @@ js_EraseMsg(jsCtx *js, const char *stream, uint64_t seq, jsOptions *opts, jsErrC
 // Account related functions
 //
 
+static natsStatus
+_unmarshalAccLimits(nats_JSON *json, jsAccountLimits *limits)
+{
+    natsStatus  s;
+    nats_JSON   *obj = NULL;
+
+    s = nats_JSONGetObject(json, "limits", &obj);
+    if (obj == NULL)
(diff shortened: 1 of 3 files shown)


src/jsm.c
```diff
@@ -1718,6 +1718,75 @@ js_EraseMsg(jsCtx *js, const char *stream, uint64_t seq, jsOptions *opts, jsErrC
...
+    nats_JSON       *json   = f->value.vobj;
+
+    t = (jsTier*) NATS_CALLOC(1, sizeof(jsTier));
+    if (t == NULL)
+        return nats_setDefaultError(NATS_NO_MEMORY);
+
+    ai->Tiers[ai->TiersLen++] = t;
+
+    DUP_STRING(s, t->Name, subject);
+    IFOK(s, nats_JSONGetULong(json, "memory", &(t->Memory)));
```
aricart, line 1758: For these you can get all the same values from `_unmarshalAccLimits`

kozlovic: No, `_unmarshalAccLimits` deals with "max" limits. The "memory", etc.. fields are also parsed in top-level of js_unmarshalAccountInfo() and so could have made that a function, but not worth it.

Note that in this function (bottom) I do call `_unmarshalAccLimits` for the account limits.

aricart: That is what I thought but wanted to verify.

PR #587, opened 2022-09-19
[CHANGED] JetStream: Pull consumer can now use AckNone and AckAll

kozlovic:
Since NATS Server v2.9.0, it is possible to create a pull consumer with an js_AckNone or js_AckAll acknowledgment mode, so removing the restriction at the API level.

Related to https://github.com/nats-io/nats-architecture-and-design/issues/156


aricart (approved):
LGTM

PR #588, opened 2022-09-20
[CHANGED] JetStream: js_GetStreamInfo() now returns all subjects

kozlovic:
When the option jsOptions.Stream.Info.SubjectsFilter is specified, the js_GetStreamInfo() would collect the subjects for the stream and return them in si->State.Subjects. However, the server had a 100,000 limit. Since server v2.9.0, pagination is supported therefore this call is going to under the cover get the info until all pages have been collected.

Related to https://github.com/nats-io/nats-architecture-and-design/issues/153

### DIFF
--- a/src/jsm.c
+++ b/src/jsm.c
@@ -24,6 +24,14 @@ typedef enum
 
 } jsStreamAction;
 
+typedef struct apiPaged
+{
+    int64_t total;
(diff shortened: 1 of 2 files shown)


aricart (approved):
LGTM

PR #596, opened 2022-10-01
[FIXED] JetStream: Ordered consumers may not be reset in some cases

kozlovic:
If the recreation of the ordered consumer were to fail, the library would invalidate the subscription and not try again.

This PR also changes the ordered consumer to use memory storage and replicas of 1. It also ensures that the ordered consumer is reset on missed heartbeats.


aricart (approved):
LGTM

PR #599, opened 2022-10-05
[FIXED] JetStream: `jsConsumerConfig.Name` was not unmarshal'ed

kozlovic:
The new configuration `Name` in `jsConsumerConfig` could be set when calling `js_AddConsumer`, but when getting the `jsConsumerInfo` response (from that call or any `js_GetConsumerInfo`, the `Name` was present in `jsConsumerInfo`, but was not in the `Config` field (a `jsConsumerConfig`). Although I am not sure this is needed, making sure that we unmarshal properly.

### DIFF
--- a/src/jsm.c
+++ b/src/jsm.c
@@ -2660,6 +2660,7 @@ _destroyConsumerConfig(jsConsumerConfig *cc)
     if (cc == NULL)
         return;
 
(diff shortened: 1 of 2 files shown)


aricart (approved):
LGTM

PR #600, opened 2022-10-05
[FIXED] JetStream: Ordered consumer issues

kozlovic:
- It would reset when receiving hearbeats due to a bug on how I was checking for sequence mismatch (overriding ldseq/sseq/etc..)
- On reset, the consumer configuration would not be the same than the original. That was because I was not making a copy of the original consumer config.

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -1246,7 +1246,7 @@ jsSub_free(jsSub *jsi)
     NATS_FREE(jsi->nxtMsgSubj);
     NATS_FREE(jsi->cmeta);
     NATS_FREE(jsi->fcReply);
-    NATS_FREE(jsi->fsubj);
+    js_destroyConsumerConfig(jsi->ocCfg);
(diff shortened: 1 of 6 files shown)


src/js.c
```diff
@@ -1538,6 +1538,7 @@ jsSub_processSequenceMismatch(natsSubscription *sub, natsMsg *msg, bool *sm)
     jsSub       *jsi   = sub->jsi;
     const char  *str   = NULL;
     int64_t     val    = 0;
+    struct mismatch *m = &jsi->mismatch;
```
ColinSullivan1, line 1541: Can m be null?

kozlovic: No, this is a pointer to an internal structure (take the address of). I could have name those fields say "mldseq", etc.. but decided to make them part of a struct.

ColinSullivan1 (approved):
LGTM.

PR #601, opened 2022-10-07
[FIXED] JetStream: Ordered Consumer could stop receiving messages

kozlovic:
Resetting some of the JS subscription needs to be done "in-place" as opposed to in the thread that attempts to recreate it.

Also speed up the reset on reconnect (as opposed to wait for missed hearbeats). Of course, it could be that the reset is for nothing if nothing was missing.


derekcollison (approved):
LGTM

PR #602, opened 2022-10-07
More fixes to ordered consumer

kozlovic:
- Ensure that we set the jsi->consumer only on when the consumer's delivery subject matches the subscription's current subject. This is really because of tests with very low HB.
- Copy the consumer name in natsSubscription_GetInfo() in case the name would be freed during a reconnect.


derekcollison (approved):
LGTM

PR #603, opened 2022-10-10
[FIXED] Reconnecting to implicit URL would not use known user info

kozlovic:
If the user provides an URL with a username and password but the library discovers new servers and a disconnect occurs, the library would reconnect without the username and password. The option `natsOptions_SetUserInfo()` is normally recommended to solve this issue.

However, to be closer to other client libraries, this library will now save the username/password of an explicit URL and use those for implicit URLs (unless the aforementioned option is used).


derekcollison (approved):
LGTM

PR #607, opened 2022-10-26
[CHANGED] KeyValue: library will no longer try to update the stream

kozlovic:
Until now, the js_CreateKeyValue() call would possibly attempt to update the underlying stream to set some properties that were added from release to release. This had an unexpected side effect when a stream in a newer server has new properties that the application linked to an older library was not aware of and would not unmarshal. Even if a stream comparison (minus the property we would want to update) would return OK, it could be wrong because we would disregard new properties not known by the old client library.


aricart (approved):
LGTM

PR #608, opened 2022-10-27
[ADDED] KeyValue: Support for Mirror and Sources

kozlovic:
Also updated state for proper Put() and Get() semantics with mirrors and across domains, e.g. leafnodes.

In practice when a mirror is across a domain, should be named the same as origin. That allows an app to run anywhere without anything special in terms of domains when binding to the KV itself.

(Removed dead-code following previous PR #607)

### DIFF
--- a/src/js.h
+++ b/src/js.h
@@ -103,6 +103,9 @@ extern const int64_t    jsDefaultRequestWait;
 #define jsAckInProgress     "+WPI"
 #define jsAckTerm           "+TERM"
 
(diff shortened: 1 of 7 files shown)


aricart (approved):
LGTM

kozlovic:
@aricart If you could have a look at the 2nd commit. It fixes the issue with delete/purge_deletes/watch.

aricart (approved):
LGTM!

PR #610, opened 2022-10-31
Fixed test for server v2.9.4

kozlovic:
Since 2.9.4 sends a "409 Server Shutdown" on shutdown for r1 pull consumers, I am tweaking a test to drop HB messages instead of stopping the server to detect missing heartbeats.


aricart (approved):
LGTM

PR #611, opened 2022-11-04
[ADDED] Stream alternates in `jsStreamInfo`

kozlovic:
`jsStreamInfo` will now possibly have a list of stream alternates represented by a mirror:
```
typedef struct jsStreamInfo
{
        (...)
        jsStreamAlternate       **Alternates;
        int                     AlternatesLen;

} jsStreamInfo;
```
The definition of a stream alternate is:
```
typedef struct jsStreamAlternate
{
        const char              *Name;
        const char              *Domain;
        const char              *Cluster;

} jsStreamAlternate;
```

Also added some of the new JS specific error codes.

### DIFF
--- a/src/jsm.c
+++ b/src/jsm.c
@@ -211,6 +211,18 @@ js_cleanStreamState(jsStreamState *state)
     _destroyStreamStateSubjects(state->Subjects);
 }
 
+static void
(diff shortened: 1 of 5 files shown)


aricart (approved):
LGTM

PR #620, opened 2022-12-09
[ADDED] KeyValue: `RateLimit` option in `kvWatchOptions`

### LINKED ISSUE #619: Rate Limit of Push Consumers for KV Watches
Is it possible to apply RateLimit using nats-c to the underlying push consumer when watching a KV ?

Users would like to throttle the initial reply when watching all.

kozlovic:
This is expressed in bits per second and allow the server to rate limit updates sent to the watcher. This applies to both initial and new updates.

Resolves #619


derekcollison (approved):
LGTM - Still want to see if this is the correct answer to the problem posed by Fineco.

kozlovic:
Closing without merging since user reported that there was no need for this PR.

PR #624, opened 2023-01-18
[ADDED] Support padding when allocating buffer for incoming messages

dmitrmax:
When json is used as messages' payload and simdjson library is used to parse them it is strictly desirable to have SIMDJSON_PADDING bytes in the end of payload buffer. Otherwise parsing can't be done in zero copy manner - simdjson will reallocate it and copy all the payload into a temporary buffer.

I would like to introduce a new NATS option which will provide ability to set such a custom padding in message delivery path.

### DIFF
--- a/src/msg.c
+++ b/src/msg.c
@@ -744,16 +744,18 @@ natsMsg_GetTime(natsMsg *msg)
 }
 
 natsStatus
-natsMsg_create(natsMsg **newMsg,
+natsMsg_createWithPadding(natsMsg **newMsg,
                const char *subject, int subjLen,
                const char *reply, int replyLen,
-               const char *buf, int bufLen, int hdrLen)
+               const char *buf, int bufLen, int bufPaddingSize, int hdrLen)
 {
     natsMsg     *msg      = NULL;
     char        *ptr      = NULL;
     int         bufSize   = 0;
     int         dataLen   = bufLen;
     bool        hasHdrs   = (hdrLen > 0 ? true : false);
+    // Make payload a null-terminated string and add at least one zero byte to the end
+    int         padLen    = (bufPaddingSize > 0 ? bufPaddingSize : 1);
 
     bufSize  = subjLen;
     bufSize += 1;
@@ -763,7 +765,7 @@ natsMsg_create(natsMsg **newMsg,
         bufSize += 1;
     }
     bufSize += bufLen;
-    bufSize += 1;
+    bufSize += padLen;
     if (hasHdrs)
         bufSize++;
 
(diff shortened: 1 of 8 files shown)


src/nats.h
```diff
@@ -3099,6 +3099,21 @@ natsOptions_DisableNoResponders(natsOptions *opts, bool disabled);
...
 
+/** \brief Sets a custom padding when allocating buffer for incoming messages
+ *
+ * By default library allocates natsMsg with payload buffer size
+ * equal to payload size. Sometimes it can be useful to add some
+ * padding to the end of the buffer which can be tweaked using
+ * this option.
+ *
+ * To clear the custom message buffer padding, call this function with 0.
+ *
```
kozlovic, line 3110: I would like to add a note here that says that changing the option has no effect on existing NATS connections.

kozlovic (changes requested):
Thank you for this pull request! Since this is an addition, I would like you to update your PR to be against the `dev` branch (instead of `main`). Also, would you be able to add a test (see https://github.com/nats-io/nats.c#testing for instructions). If you can't that's ok, I can come up with one and validate that your changes are ok before merge, then will submit a PR myself to add testing.

src/nats.h
dmitrmax: Done

src/msg.c
```diff
@@ -744,16 +744,18 @@ natsMsg_GetTime(natsMsg *msg)
...
-               const char *buf, int bufLen, int hdrLen)
+               const char *buf, int bufLen, int bufPaddingSize, int hdrLen)
 {
     natsMsg     *msg      = NULL;
     char        *ptr      = NULL;
     int         bufSize   = 0;
     int         dataLen   = bufLen;
     bool        hasHdrs   = (hdrLen > 0 ? true : false);
+    // Make payload a null-terminated string and add at least one zero byte to the end
+    int         padLen    = (bufPaddingSize > 0 ? bufPaddingSize : 1);
```
kozlovic, line 758: As you have seen, say the payload is 5 bytes and corresponds to `hello`, the library allocates 6 bytes and stores the `\0` at the end, so we end up with payload `hello\0`, which is convenient to get the message payload as a string (if content is a string). With your changes, the padding size incudes the `\0`, is that what you wanted? That is, suppose that for same payload, you set the option of padding size of 5, then the result would be `hello\0\0\0\0\0`, which means 5 zeros, not 5+1. It is inconsequential for normal use (if the payload is a string), and the message data length API will still return 5, but it makes the padding case be a bit different than no padding in that non padded means payload + 1, while padded means payload+padding (and not +1). If you think this is what you need for your use case, then I am fine with this change.

kozlovic (commented):
Thank you for adding the test, I appreciate it! I have posted a comment. If you think that it is ok to keep it this way (suits your use-case), then I will approve and merge.

src/msg.c
dmitrmax: Actually what was done before (adding just a single `\0` to the end) can be seen just as a special case for a generic padding implementation which is done in this PR. Now by default we have one `\0` but with the new option we can increase the number of `\0` to any arbitrary number. And yes, if I set 5 in this option then `hello\0\0\0\0\0` is exactly what I'm expecting!

dmitrmax:
@kozlovic thank you for such a turbo review and accepting the PR )

PR #625, opened 2023-01-23
[CHANGED] Default connection error handler.

### LINKED ISSUE #622: Default Error Listener
Go example here:  https://github.com/nats-io/nats.go/blob/b142f6053bcff1167f6b6c5fc1a56df120fd059f/nats.go#L1379

kozlovic:
If no error handler is specified by the user, the library will use a default error handler that prints some error information to stderr.

If the user did not specify an error handler but does not wish to have anything printed to stderr, the user should set a dummy error handler that does not do anything.

Resolves #622


aricart (approved):
LGTM

PR #632, opened 2023-02-01
[FIXED] Bytes accounting while reading the INFO protocol

dmitrmax:
This is potential buffer overflow vulnerability that can be exploited if server is controlled by an intruder.

If we detect that the buffer contains data then we have to reduce available space for the subsequent read operation. Otherwise read can overwrite memory past the buffer.

P.S.: Just discovered that this function is called only from one place and supplied buffer is always empty there. So actualy not a vulnerabilty but as it is written should be fixed in case anyone in future will use this function with non-empty buffer.


kozlovic (approved):
LGTM! I have updated the title of the PR. Thank you for you contribution.

PR #636, opened 2023-02-08
[FIXED] natsConnection_GetConnectedUrl() not working during reconnect

### LINKED ISSUE #635: Signature callback does not know the server URL
Hi,
I'm using `natsOptions_SetNKey` and my custom callback to extract a NKey seed from a secure storage (e.g. Windows credential manager or KWallet). However, in case I have multiple NATS servers, I don't know with which one I'm trying to 'shake hands', because `natsConnection_GetConnectedUrl` still returns NULL during my callback. 

Would it be possible to make `natsConnection_GetConnectedUrl` return a value already after receiving INFO, while the handshake is still in progress?

Thanks in advance

kozlovic:
This function, and natsConnection_GetConnectedServerId() would work only when the connection was fully connected. There is no reason to not make it work while the connection is reconnecting.

Resolves #635

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3881,7 +3881,8 @@ natsConnection_GetConnectedUrl(natsConnection *nc, char *buffer, size_t bufferSi
 
     buffer[0] = '\0';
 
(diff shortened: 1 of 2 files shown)


aricart (approved):
LGMT

PR #638, opened 2023-02-16
[FIXED] Crash processing inbound message for destroyed subscription

### LINKED ISSUE #637: Random crash in natsConn_processMsg
We are getting this sporadic errors, they seem to be correlated with a subscription timing out but looking at the code I can't figure out why would it crash when trying to lock the connection mutex. Any ideas what could cause this? it looks like a race condition between the readloop and the lifetime of the connection.

```
/usr/lib/x86_64-linux-gnu/libpthread.so.0(+0x14420) [0x7fedfaf84420]
/usr/lib/x86_64-linux-gnu/libpthread.so.0(pthread_mutex_trylock+0x17) [0x7fedfaf7b257]
./app(natsMutex_Lock+0x34) [0x564abde25964]
./app(natsConn_processMsg+0x1ae) [0x564abddf56ee]
./app(natsParser_Parse+0x419) [0x564abde19989]
./app(+0x2093e93) [0x564abddf4e93]
./app(+0x20c4ca4) [0x564abde25ca4]
/usr/lib/x86_64-linux-gnu/libpthread.so.0(+0x8609) [0x7fedfaf78609]
/usr/lib/x86_64-linux-gnu/libc.so.6(clone+0x43) [0x7fedf538f133]
```

kozlovic:
A race could cause the read loop to crash when processing a message for a subscription that is being removed in another thread.

Resolves #637

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2617,7 +2617,6 @@ natsConn_processMsg(natsConnection *nc, char *buf, int bufLen)
     bool             sc   = false;
     bool             sm   = false;
     nats_MsgList     *list = NULL;
-    natsMutex        *mu   = NULL;
     natsCondition    *cond = NULL;
     // For JetStream cases
     jsSub            *jsi    = NULL;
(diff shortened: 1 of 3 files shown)


aricart (approved):
LGTM

PR #643, opened 2023-03-08
Fixed some flappers

kozlovic:
Signed-off-by: Ivan Kozlovic <ijkozlovic@gmail.com>


levb (approved):
LGTM. Curious: why 900?

kozlovic:
> LGTM. Curious: why 900?

In the past, experimentation with Travis showed that it was enough. I try to have tests finish as fast as possible but don't want to have flappers, so that's how I ended up with 900.

PR #645, opened 2023-03-09
[FIXED] natsConnection_RequestMsg() incorrectly returns NATS_NO_SERVER_SUPPORT

### LINKED ISSUE #644: natsConnection_RequestMsg returns NATS_NO_SERVER_SUPPORT instead of NATS_TIMEOUT when there is no connection
This happens only when sending messages with headers, if no headers are sent it behaves normally.

```c++
        natsMsg *msg = nullptr;
        auto status = natsMsg_Create(&msg, subject.c_str(), nullptr, (const char*) message->data(), (int)message->size());
        if (NATS_OK != status){
            throw std::runtime_error("Issue creating requesting msg" + std::string(nats_GetLastError(&status)));
        }

        // If this line is removed we get NATS_TIMEOUT instead of NATS_NO_SERVER_SUPPORT
        if (natsMsgHeader_Set(msg, "some_header", "some_value") != NATS_OK) {
            throw std::runtime_error("Couldn't set the header");
        }

        natsMsg *replyMessage = nullptr;
        status = natsConnection_RequestMsg(&replyMessage, connection_handle, msg, timeout);
        natsMsg_Destroy(msg);
        if(NATS_TIMEOUT == status || NATS_NO_RESPONDERS == status){
            return nullptr;
        }

        // We shouldn't get here if there is no server running
        if(NATS_OK != status){
            throw std::runtime_error("Issue requesting: " + std::string(nats_GetLastError(&status)));
        }
```

kozlovic:
When using `natsOptions_SetRetryOnFailedConnect` option to connect and the library is not yet connected, calling `natsConnection_RequestMsg` with a message with headers would incorrectly return `NATS_NO_SERVER_SUPPORT` instead of `NATS_TIMEOUT`.

Resolves #644

### DIFF
--- a/src/pub.c
+++ b/src/pub.c
@@ -104,6 +104,10 @@ natsConn_publish(natsConnection *nc, natsMsg *msg, const char *reply, bool direc
     // accessing the headers. It should still be considered having headers.
     if ((msg->headers != NULL) || natsMsg_needsLift(msg))
     {
+        // Do the check for server's headers support only after we have completed
+        // the initial connect (we could be here with initc true - that is, initial
+        // connect in progress - when using natsOptions_SetRetryOnFailedConnect
+        // option).
         if (!nc->initc && !nc->info.headers)
         {
             natsConn_Unlock(nc);
--- a/src/pub.c
+++ b/src/pub.c
@@ -104,7 +104,7 @@ natsConn_publish(natsConnection *nc, natsMsg *msg, const char *reply, bool direc
     // accessing the headers. It should still be considered having headers.
     if ((msg->headers != NULL) || natsMsg_needsLift(msg))
     {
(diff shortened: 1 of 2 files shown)


src/pub.c
```diff
@@ -104,7 +104,7 @@ natsConn_publish(natsConnection *nc, natsMsg *msg, const char *reply, bool direc
     // accessing the headers. It should still be considered having headers.
     if ((msg->headers != NULL) || natsMsg_needsLift(msg))
     {
-        if (!nc->info.headers)
```
levb, line 107: nit: 0/5 update the comment with the comment from this PR to explain why `initc` is excluded?

levb (approved):

PR #647, opened 2023-04-06
added `natsConn_(get|set)(Error|Closed)Callback`

levb:
To use in the microservices implementation, to wrap connection handlers on an active connection.

(still need to add a test for `setClosedCallback`)

### DIFF
--- a/src/nats.c
+++ b/src/nats.c
@@ -726,11 +726,14 @@ _timerThread(void *arg)
 static void
 _asyncCbsThread(void *arg)
 {
-    natsLibAsyncCbs *asyncCbs = &(gLib.asyncCbs);
-    natsAsyncCbInfo *cb       = NULL;
-    natsConnection  *nc       = NULL;
+    natsLibAsyncCbs         *asyncCbs   = &(gLib.asyncCbs);
+    natsAsyncCbInfo         *cb         = NULL;
+    natsConnection          *nc         = NULL;
+    natsConnectionHandler   cbHandler   = NULL;
+    natsErrHandler          errHandler  = NULL;
+    void                    *cbClosure  = NULL;
 #if defined(NATS_HAS_STREAMING)
-    stanConnection  *sc       = NULL;
+    stanConnection          *sc         = NULL;
 #endif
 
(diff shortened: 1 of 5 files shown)


test/test.c
```diff
@@ -34440,6 +34513,7 @@ static testInfo allTests[] =
     {"SyncSubscriptionPending",         test_SyncSubscriptionPending},
     {"SyncSubscriptionPendingDrain",    test_SyncSubscriptionPendingDrain},
     {"AsyncErrHandler",                 test_AsyncErrHandler},
+    {"AsyncSetErrHandler",              test_AsyncSetErrHandler},
```
kozlovic, line 34516: This is not enough to add a test to the test suite. Follow instructions [here](https://github.com/nats-io/nats.c#testing).

kozlovic (changes requested):
Tests are failing (does not seem to be flapping since the same tests fail in various matrix runs). It could be because of the extra `break` that I pointed out, but not sure.

src/nats.c
```diff
@@ -760,41 +763,62 @@ _asyncCbsThread(void *arg)
         sc = cb->sc;
 #endif
 
+        // callback handlers can be updated on a live connection, so we need to
+        // lock.
+        natsMutex_Lock(nc->mu);
```
kozlovic, line 768: You would also need to reset cbHandler/Closure and errHandler/closure in each loop iteration!

levb:
Oops, @kozlovic sorry I tagged you too early; I pushed to run the tests, still finishing this up. Thanks for jumping onto it!

kozlovic (approved):
LGTM. Just a note, since recently, it is assumed that the async error callback will always be set (if not set by the user, we use a default handler that prints to stderr, see `natsConn_defaultErrHandler`). I see that your new setter does not allow NULL, so there is no risk at the moment that you would remove the error handler. If you do in the future, check invocations of `natsAsyncCb_PostErrHandler` that again assume that a callback is always set.

levb:
#640 has an improved implementation, closing this PR.

PR #650, opened 2023-05-02
natsConnection_Buffered: fix returned value in case of error

kolomenkin:
This function returns number of bytes in the buffer.
It is incorrect to return value from `natsStatus` enum.

We could keep setting internal error code.
But in this case we should set error code for other cases too where we return `-1`

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3811,7 +3811,7 @@ natsConnection_Buffered(natsConnection *nc)
     int buffered = -1;
 
     if (nc == NULL)
-        return nats_setDefaultError(NATS_INVALID_ARG);
+        return buffered;
 
     natsConn_Lock(nc);
 


kolomenkin:
@kozlovic , sorry, my mistake. I knew I should create PR against dev, but forgot to ensure this.
I updated PR's target branch few moments ago.

kozlovic:
At this point, from your fork, make sure that you are up-to-date with the upstream `main` branch and create a PR with that one line change. There should not be 7 commits like there is in this PR.

kolomenkin:
@kozlovic , sorry, my first PR fix was incorrect.
Now I rebased my commit and branch on top of latest master branch.
So now there is only 1 commit in PR.

kozlovic (approved):

PR #656, opened 2023-05-19
Lists of callbacks for natsOptions: cbClosed, cbAsyncErr

levb:
Addresses https://github.com/nats-io/nats.c/pull/640#discussion_r1192493551

Added support for multiple (chained) handlers for ConnectionClosed and AsyncError callback events.

- Added a `nats_CallbackList` internal type, use it for closed and error cases.
- Removed prior code for `natsConn_get/setXXXCallback`.
- Cleaned up access to other callbacks in conn.c (lock `opts`).

### DIFF
--- a/src/opts.c
+++ b/src/opts.c
@@ -1045,6 +1045,8 @@ natsStatus
 natsOptions_SetErrorHandler(natsOptions *opts, natsErrHandler errHandler,
                             void *closure)
 {
+    nats_CallbackList *replaced = NULL;
+
     if (errHandler == NULL)
     {
         errHandler = natsConn_defaultErrHandler;
@@ -1058,6 +1060,8 @@ natsOptions_SetErrorHandler(natsOptions *opts, natsErrHandler errHandler,
         UNLOCK_OPTS(opts);
         return natsOptions_addErrorCallback(opts, errHandler, closure);
     }
+    
+    replaced = opts->asyncErrCb;
     if (opts->asyncErrCb->next != NULL)
     {
         // can't allow overriding a list of callbacks with a single one.
@@ -1071,6 +1075,7 @@ natsOptions_SetErrorHandler(natsOptions *opts, natsErrHandler errHandler,
 
     UNLOCK_OPTS(opts);
 
+    NATS_FREE(replaced);
     return NATS_OK;
 }
 
@@ -1101,6 +1106,7 @@ natsOptions_SetClosedCB(natsOptions *opts, natsConnectionHandler closedCb,
             UNLOCK_OPTS(opts);
             return natsOptions_addConnectionClosedCallback(opts, closedCb, closure);
         }
+        to_free = opts->closedCb;
         opts->closedCb->type = CALLBACK_TYPE_CONN;
         opts->closedCb->f.conn = closedCb;
         opts->closedCb->closure = closure;
--- a/src/opts.c
+++ b/src/opts.c
@@ -1011,6 +1011,7 @@ natsOptions_addErrorCallback(natsOptions *opts, natsErrHandler f, void *closure)
 
     UNLOCK_OPTS(opts);
 
+    NATS_FREE(replaced);
     return s;
 }
 
--- a/src/opts.c
+++ b/src/opts.c
@@ -1045,8 +1045,6 @@ natsStatus
 natsOptions_SetErrorHandler(natsOptions *opts, natsErrHandler errHandler,
                             void *closure)
 {
-    nats_CallbackList *replaced = NULL;
-
     if (errHandler == NULL)
     {
         errHandler = natsConn_defaultErrHandler;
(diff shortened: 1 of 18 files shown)


src/opts.c
```diff
@@ -883,17 +883,190 @@ natsOptions_SetMaxPendingMsgs(natsOptions *opts, int maxPending)
...
+    if (cb == NULL)
+        return nats_setDefaultError(NATS_NO_MEMORY);
+    cb->type = CALLBACK_TYPE_ERROR;
+    cb->f.err = f;
+    cb->closure = closure;
+
+    LOCK_AND_CHECK_OPTIONS(opts, 0);
+
+    // unlink if already in the list.
+    natsOptions_unlinkCallback(&replaced, &opts->asyncErrCb, (void(*)(void))f, closure);
```
kozlovic, line 1006: You are missing a free for `replaced`

src/opts.c
```diff
@@ -883,17 +883,190 @@ natsOptions_SetMaxPendingMsgs(natsOptions *opts, int maxPending)
...
+
+void natsOptions_removeConnectionClosedCallback(natsOptions *opts, natsConnectionHandler f, void *closure)
+{
+    nats_CallbackList *removed = NULL;
+
+    natsMutex_Lock(opts->mu);
+
+    natsOptions_unlinkCallback(&removed, &opts->closedCb, (void(*)(void))f, closure);
+
+    UNLOCK_OPTS(opts);
```
kozlovic, line 1025: nit: but since you used explicitly `natsMutex_Lock(opts->mu)` above, I would use `natsMutex_Unlock()` here, or use the macro to lock - just for consistency sake.

src/opts.c
```diff
@@ -883,17 +883,190 @@ natsOptions_SetMaxPendingMsgs(natsOptions *opts, int maxPending)
...
+}
+
 natsStatus
 natsOptions_SetErrorHandler(natsOptions *opts, natsErrHandler errHandler,
                             void *closure)
 {
-    LOCK_AND_CHECK_OPTIONS(opts, 0);
+    if (errHandler == NULL)
+    {
+        errHandler = natsConn_defaultErrHandler;
```
kozlovic, line 1049: But in this case, should you not keep track if there was an "old" one and if so free it like you do for `SetClosedCB`? In other words, for error CB we always set it to something, but that does not mean that we may be replacing something that would then need to be freed?

src/opts.c
```diff
@@ -934,7 +1131,7 @@ natsOptions_SetReconnectedCB(natsOptions *opts,
                              natsConnectionHandler reconnectedCb,
                              void *closure)
 {
-    LOCK_AND_CHECK_OPTIONS(opts, 0);
+LOCK_AND_CHECK_OPTIONS(opts, 0);
```
kozlovic, line 1134: Check indentation?

kozlovic (changes requested):

src/opts.c
levb: oops, thx! fixed. 0/5 if it's worth the time adding tests for these new methods. Perhaps if they get another use-case and/or get publicly exposed?

src/opts.c
levb: Actually, no. Here we only allow overriding a "single" callback, and hijack the struct that is already in place. `SetClosedCB` allows to set to `NULL` in which case there is something to free.

Sorry I overreacted and pushed a commit to "fix" this, now

src/opts.c
```diff
@@ -883,17 +883,190 @@ natsOptions_SetMaxPendingMsgs(natsOptions *opts, int maxPending)
...
+
+void natsOptions_removeErrorCallback(natsOptions *opts, natsErrHandler f, void *closure)
+{
+    nats_CallbackList *removed = NULL;
+
+    natsMutex_Lock(opts->mu);
+
+    natsOptions_unlinkCallback(&removed, &opts->asyncErrCb, (void(*)(void))f, closure);
+
+    UNLOCK_OPTS(opts);
```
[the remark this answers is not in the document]
levb: added a `natsOptions_lock/unlock` and used them everywhere, will makes it easier to grep for them.

src/opts.h
```diff
@@ -16,12 +16,22 @@
...
+static inline void natsOptions_unlock(natsOptions *opts)
+{
+    natsMutex_Unlock(opts->mu);
+}
+
 #define LOCK_AND_CHECK_OPTIONS(o, c) \
     if (((o) == NULL) || ((c))) \
         return nats_setDefaultError(NATS_INVALID_ARG); \
-    natsMutex_Lock((o)->mu);
+    natsOptions_lock(o);
```
kozlovic, line 32: You need to undo this change or make a change to `sopts.c` because it fails for Stan options.

kozlovic (changes requested):
Also getting a warning from micro-stats that you did not change in this PR. I am assuming that this will be handled in the micro service PR itself.

kozlovic (approved):
LGTM!

PR #657, opened 2023-05-22
Removed the never-implemented schema

levb:
I never finished the schema implementation, but thankfully it's been removed, see https://github.com/nats-io/nats-architecture-and-design/issues/220 and https://github.com/nats-io/nats.go/pull/1270.

This PR removes it from the C implementation.


piotrpio (approved):
LGTM

PR #658, opened 2023-05-26
Fixed error handling and final cleanup

levb:
Haven't really fixed anything yet.

WAS:
_- Fixed `micro` error callback mapping to endpoints & the test.
- Fixed the Stop/Destroy cleanup - ended up non-trivial, and potentially overly complex. @kozlovic I'd really appreciate any tips on how this can be simpler. Cases this handles:
  - A "normal" shutdown. the main/user thread calling `microService_Destroy`, draining the subs, and leaving `nc` intact,  `on_connection_closed` is never called.
  - The connection is closed - closes the subs before calling `on_connection_closed`, server stops
  - An error - the subs need to be drained._

### DIFF
--- a/src/micro.c
+++ b/src/micro.c
@@ -41,11 +41,6 @@ micro_AddService(microService **new_m, natsConnection *nc, microServiceConfig *c
 
     MICRO_CALL(err, micro_clone_service_config(&m->cfg, cfg));
 
-    if (err == NULL)
-    {
-        m->is_running = true;
-    }
-
     // Wrap the connection callbacks before we subscribe to anything.
     MICRO_CALL(err, wrap_connection_event_callbacks(m))
     MICRO_CALL(err, micro_init_monitoring(m));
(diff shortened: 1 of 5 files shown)


levb:
@kozlovic I am kinda struggling with a clean shutdown on errors. The stream of async errors for a connection continues after a microservice is stopped, and there appear no way to "drain" all pending error messages to properly refcount the `m` pointer (closure) in the callback.

One solution that I seem coming back to is a global (or connection-level?) higher-level list of active microservices, with its own lock, that can be checked against. Initially it seemed an overkill, but (a) it should work, and (b) it would just have single error/closed callbacks for all microservices, and all that "CallbackList" ugliness can be disposed of. Thoughts? 

PS. I think this was a self-inflicted wound of trying to replicate the Go client's logic; I now really appreciate your tip that designing "clean" for the C implementation may be more appropriate; and then dressing up to be consistent, at the API level.

levb:
A different PR is coming.

PR #659, opened 2023-06-01
[FIXED] Possible crash if error handler if subscription was destroyed

kozlovic:
Since error handler is invoked asynchronously, if the user had destroyed the subscription, it was possible that the error handler would be presented with a subscription that was already freed.

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -14374,6 +14374,130 @@ test_AsyncErrHandler(void)
 }
 
 static void
+_asyncErrBlockingCb(natsConnection *nc, natsSubscription *sub, natsStatus err, void* closure)
+{
+    struct threadArg    *arg = (struct threadArg*) closure;
+
+    natsMutex_Lock(arg->m);
+
+    arg->sum++;
+
+    while ((arg->sum == 1) && !arg->closed)
+        natsCondition_Wait(arg->c, arg->m);
+
+    if (sub != arg->sub)
+        arg->status = NATS_ERR;
+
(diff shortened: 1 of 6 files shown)


test/test.c
```diff
@@ -14373,6 +14373,130 @@ test_AsyncErrHandler(void)
...
+    if (sub != arg->sub)
+        arg->status = NATS_ERR;
+
+    if ((arg->status == NATS_OK) && (err != NATS_SLOW_CONSUMER))
+        arg->status = NATS_ERR;
+
+    if (arg->status == NATS_OK)
+    {
+        // Call some subscription API to make sure that the pointer has not been freed.
+        arg->current = natsSubscription_IsValid(sub);
```
levb, line 14397: 👍

test/test.c
```diff
@@ -14373,6 +14373,130 @@ test_AsyncErrHandler(void)
...
+    s = natsConnection_PublishString(nc, "foo", "msg3");
+    IFOK(s, natsConnection_PublishString(nc, "foo", "msg4"));
+    testCond(s == NATS_OK);
+
+    test("Wait a bit for async error to be posted: ");
+    nats_Sleep(200);
+    testCond(true);
+
+    test("Destroy subscription: ");
+    natsSubscription_Destroy(sub);
```
levb, line 14475: 👍

levb (approved):
LGTM! Let me try this with `micro`, see if it helps resolve the order of execution on shutdown issues there. It's a better fix than what I was trying to do just wrapping the invocation itself.

PR #660, opened 2023-06-03
Fixed microservice cleanup and tests

levb:
Fixes race issues on cleanup, and a few small bugs.

NOTE:
- `Test_MicroServiceStopsOnClosedConn` and `Test_MicroServiceStopsWhenServerStops` still appear to take too long, but I am not sure what's causing the wait/timeout there.
- I've seen `Test_MicroAsyncErrorHandler` hang in CI once, was not able to reproduce nor debug the problem.

For non-`micro` changes (`nats.[ch]`, `opts.[ch]`, it may be easier to view against `dev`, https://github.com/nats-io/nats.c/compare/dev...lev-tmp2 since this PR mostly rolls back prior changes.

- Added a global list of running services, use it for all connection callbacks.
- Reworked service/endpoint locking and refcounting.
- Improved synchronization in tests
- Fixed `micro_ErrorFromStatus`
- Renamed some static methods for consistency (`_` prefix)


src/micro.c
```diff
@@ -75,11 +99,20 @@ micro_add_endpoint(microEndpoint **new_ep, microService *m, const char *prefix,
     if (cfg == NULL)
         return NULL;
 
+    // retain `m` before the endpoint uses it for its on_complete callback.
```
levb, line 87: The service now does its own refcounting for endpoints to avoid a race on cleanup

src/micro.c
```diff
@@ -143,127 +170,124 @@ micro_add_endpoint(microEndpoint **new_ep, microService *m, const char *prefix,
...
 
-    err = microService_Stop(m);
-    if (err != NULL)
-        return err;
+    _lock_service(m);
 
-    micro_release_service(m);
-    return NULL;
-}
+    // Release the service reference for the completed endpoint. It can not be
```
levb, line 272: matches `retain` in `micro_add_endpoint`

test/test.c
```diff
@@ -33135,21 +33055,27 @@ test_MicroServiceStopsOnClosedConn(void)
     test("Close the connection: ");
     testCond(NATS_OK == natsConnection_Drain(nc));
     natsConnection_Close(nc);
+
+    test("Wait for it: ");
+    _waitForConnClosed(&arg);
```
kozlovic, line 33060: Replace this with:
```
testCond(_waitForConnClosed(&arg) == NATS_OK);
```

test/test.c
```diff
@@ -33190,38 +33115,43 @@ test_MicroServiceStopsWhenServerStops(void)
...
-
+    _startMicroservice(&m, nc, &cfg, &arg);
+    
     test("Test microservice is running: ");
     testCond(!microService_IsStopped(m))
 
     test("Stop the server: ");
     testCond((_stopServer(serverPid), true));
 
+    nats_Sleep(1000);
```
kozlovic, line 33126: Not sure you need the sleep here and that may be the reason why you think that this takes longer than needed.

test/test.c
```diff
@@ -33135,21 +33055,27 @@ test_MicroServiceStopsOnClosedConn(void)
     test("Close the connection: ");
     testCond(NATS_OK == natsConnection_Drain(nc));
     natsConnection_Close(nc);
+
+    test("Wait for it: ");
```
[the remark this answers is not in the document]
levb: 👍 it got lost in my cut&pasting I think, and that's why the test was taking so long - was timing out. Thx!!

test/test.c
levb: 🤦

levb:
@kozlovic Thx for suggesting natsLib as the placeholder for the globals. I like it and will definitely keep it. `nc` is already retained/released by `_new/_free_service`. I actually considered adding these (hash, etc) to connections but that seemed too intrusive/presumptuous even if services bind to single connections ATM. LMK what else (if anything) you think needs to be done for the proper cleanup there?

I think I fixed (or explained) all the other PR feedback, thank you! So silly that I forgot about that `nats_Sleep` and `_onClosed` and was sweating why it was taking too long!

2 questions:
1. What environment do you use to get detailed valgrind info? It doesn't seem visible in CI, and I'd love to fix the CI to display it.
2. What code formatter do you use? (re: whitespace misses). I use VSCode, and the default C formatter is not 100% compatible with the existing code, so I didn't use it thoroughly on my changes to shared files. I am 0/5 on the choice, can just use the same as you do, if possible.

src/micro.c
```diff
@@ -383,78 +481,205 @@ void micro_free_cloned_service_config(microServiceConfig *cfg)
...
-    microService *m = (microService *)closure;
+    natsStatus s = NATS_OK;
+
     if (m == NULL)
-        return;
+        return micro_ErrorInvalidArg;
+
+    if (service_callback_mu == NULL)
+    {
+        IFOK(s, natsMutex_Create(&service_callback_mu));
```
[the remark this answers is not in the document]
levb: (see my comment at the end, tl;dr thx, keep)

kozlovic:
> What environment do you use to get detailed valgrind info? It doesn't seem visible in CI, and I'd love to fix the CI to display it.

If you run with `ctest -T memcheck`, it will use the default valgrind options that I define in the top-root `CMakeLists.txt` file: `"--leak-check=full --track-fds=yes --show-reachable=yes --num-callers=50"` (that you can see if you do a `make edit_cache`). There is currently no matrix in Travis to run those (instead I run the sanitize thread/address). I usually run a manual `ctest -T memcheck` on my Linux VM.

Keep in mind that if you have a real leak, then the current runs on Travis would find it (with sanitize address). What is not reported is the still reachable, for instance the global variables you added at one point.

> What code formatter do you use? (re: whitespace misses). I use VSCode, and the default C formatter is not 100% compatible with the existing code, so I didn't use it thoroughly on my changes to shared files. I am 0/5 on the choice, can just use the same as you do, if possible.

I use the same, but you may want to check Settings and look for `trim` in the search, you should have "Text Editor/Files" and then you should check the box "Files: Trim Trailing Whitespaces" which says "When enabled, will trim trailing whitespace when saving a file".

src/micro_error.c
```diff
@@ -107,7 +107,7 @@ micro_ErrorFromStatus(natsStatus s)
     err = NATS_CALLOC(1, sizeof(microError) + message_len + 1);
     if (err == NULL)
         return &_errorOutOfMemory;
-
+    err->message = (char *)(err + 1);
```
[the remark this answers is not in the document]
kozlovic: As of now, it is declared as `char *message` (at least in this PR). That is actually what I do (single alloc of struct + some content) with `struct __natsMsg`. 

I would declare `const char* message`, to indicate that this should not be freed individually, and add a comment before that field to say that it has to be the last in the struct (in case you add something later). So you should have something like this:
```
char *ptr = NULL;
...
ptr = (char*) (((char*) &(err->message)) + sizeof(err->message));
memcpy(ptr, message, message_len + 1);
err->message = (const char*) ptr;
```

kozlovic (changes requested):
I would recommend making the change I recommend about error's message. Note that if you do so, you may need to cast ("`(char*) err->message`") at line 68.

src/micro_error.c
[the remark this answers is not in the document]
levb: @kozlovic I changed to `const`, 👍 it's cleaner. I decided on `ptr=(char*)err + sizeof(*err)` as the least controversial way to calculate `ptr`. Please take a look.

kozlovic (changes requested):
Just some little changes and I think we are good to go.

kozlovic (approved):
LGTM!

PR #664, opened 2023-06-14
Merge dev into main

levb:


kozlovic (approved):
LGTM

levb:
@kozlovic I intend to squash and merge, but I think historically you used a "squash commit", keeping all ^^^ commits in the `main`'s log. Do you have a (strong?) preference going forward? Do you see tradeoffs of importance? I like squash since it drops all the cross-merge history.

kozlovic:
Sorry, the comment above was related more to the "micro" PR. In case of merging `dev` to `main`, you actually don't want to do a squash because it would then be difficult to track what were the contributions that need to go in the release notes. The "micro" PR should have been squashed, but a merge from dev to main should not.

PR #669, opened 2023-07-10
[ADDED] Build with github actions

levb:
Moved the build system from Travis to GitHub actions.

There are 2 main jobs: 
- `debug` runs test coverage and sanitize with gcc, in debug mode; it also runs debug-compiled tests on clang and latest/LTS Ubuntu versions.
- `release` compiles in release mode, and tests various server versions, and packaging options (no TLS, no streaming)

PRs run `debug` by default. If you want to run the release build on a PR before merging it, set the `build-release` on it.

- Moved test coverage to codecov
- Added testing with Ubuntu (latest, 20.04), several versions of gcc, clang, latest and `dev`/`main` NATS server
- https://github.com/nats-io/nats.c/pull/671 checks race detection (`-fsanitize=thread`)
- https://github.com/nats-io/nats.c/pull/670 checks leak and access-after-free detection (`-fsanitize=address`)
- https://github.com/nats-io/nats.c/pull/673 turns on verbose cmake to compare compile flags
  - note that cmake stopped adding `-fPIC` to gcc, 2/5 it's ok
- cleaned out the use of `NATS_BUILD_TLS_USE_OPENSSL_1_1_API` in CI since it's always on

NOTE that since I added a "catch all" for thread sanitizer warnings, many tests have been failing in that mode both on travis and github workflows, so I filed https://github.com/nats-io/nats.c/issues/672 and skipped them in the `sanitize=thread` mode.


mtmk (approved):
LGTM

PR #679, opened 2023-09-12
[ADDED] Support for (multiple) ConsumerConfig.FilterSubjects

levb:
@mtmk @piotrpio @Jarema I am not sure how much client logic there needs to be if both `FilterSubject` and `FilterSubjects` are set; I tried to match the `subscription<>consumer filter` logic, otherwise the changes are minimal, just added the fields. LMK if you think more needs to be done here.

### DIFF
--- a/src/jsm.c
+++ b/src/jsm.c
@@ -2756,6 +2756,22 @@ _marshalConsumerCreateReq(natsBuffer **new_buf, const char *stream, jsConsumerCo
         IFOK(s, natsBuf_Append(buf, cfg->FilterSubject, -1));
         IFOK(s, natsBuf_AppendByte(buf, '"'));
     }
+    if ((s == NATS_OK) && (cfg->FilterSubjectsLen > 0))
+    {
+        int i;
+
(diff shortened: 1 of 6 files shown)


src/jsm.c
```diff
@@ -3083,7 +3105,7 @@ js_AddConsumer(jsConsumerInfo **new_ci, jsCtx *js,
         {
             // No subject filter, use <stream>.<consumer name>
             // otherwise, the filter subject goes at the end.
-            if (nats_IsStringEmpty(cfg->FilterSubject))
+            if (nats_IsStringEmpty(cfg->FilterSubject) || (cfg->FilterSubjectsLen > 0))
```
mtmk, line 3108: Probably an unnecessary check. When both `filter_subject` and `filter_subjects` are set server seems to be returning `consumer cannot have both` error whether the API subject was appended with `filter_subject` or not.

levb: took it out

mtmk (approved):
LGTM

levb:
`Test_ServerPoolUpdatedOnClusterUpdate` seems to be flapping a lot (after the OS change?), investigating.

PR #680, opened 2023-09-13
[CHANGED] Travis build changes

levb:
This PR prepares for testing with nats server v2.10 changes. It also fixes an issue with `sanitize=thread` that seemed to have appeared due to Travis `bionic` image changes.

- Use Ubuntu `focal` v20.04
- Added testing against nats-server `latest` release, and against the `dev` branch
- Changed gcc `sanitize=address` builds to Debug to get line numbers in the output
- Added `NATS_TEST_VALGRIND=yes` to `sanitize=thread` to reduce repeat counts and avoid timeouts

### DIFF
--- a/buildOnTravis.sh
+++ b/buildOnTravis.sh
@@ -17,7 +17,7 @@ if [ "$NATS_TEST_SERVER_VERSION" != "" ]; then
     rel=$(curl -s https://api.github.com/repos/nats-io/nats-server/releases/latest | jq -r '.tag_name')
   fi
 
-  if wget https://github.com/nats-io/nats-server/releases/download/$rel/nats-server-$rel-linux-amd64.tar.gz; then
+  if [ "$rel" != "${rel#v}" ] && wget https://github.com/nats-io/nats-server/releases/download/$rel/nats-server-$rel-linux-amd64.tar.gz; then
     tar -xzf nats-server-$rel-linux-amd64.tar.gz
     mv nats-server-$rel-linux-amd64 $HOME/nats-server-$rel
   else
--- a/buildOnTravis.sh
+++ b/buildOnTravis.sh
@@ -10,6 +10,23 @@ echo "coverage   = " $2
 echo "build opts = " $3
 echo "test opts  = " $4
 
+if [ "$NATS_TEST_SERVER_VERSION" != "" ]; then
+  rel=$NATS_TEST_SERVER_VERSION
+  mkdir -p $HOME/nats-server-$rel
+  if [ "$rel" = "latest" ]; then
+    rel=$(curl -s https://api.github.com/repos/nats-io/nats-server/releases/latest | jq -r '.tag_name')
+  fi
+
+  if wget https://github.com/nats-io/nats-server/releases/download/$rel/nats-server-$rel-linux-amd64.tar.gz; then
+    tar -xzf nats-server-$rel-linux-amd64.tar.gz
+    mv nats-server-$rel-linux-amd64 $HOME/nats-server-$rel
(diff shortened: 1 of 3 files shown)


buildOnTravis.sh
```diff
@@ -10,6 +10,23 @@ echo "coverage   = " $2
...
 echo "test opts  = " $4
 
+if [ "$NATS_TEST_SERVER_VERSION" != "" ]; then
+  rel=$NATS_TEST_SERVER_VERSION
+  mkdir -p $HOME/nats-server-$rel
+  if [ "$rel" = "latest" ]; then
+    rel=$(curl -s https://api.github.com/repos/nats-io/nats-server/releases/latest | jq -r '.tag_name')
+  fi
+
+  if wget https://github.com/nats-io/nats-server/releases/download/$rel/nats-server-$rel-linux-amd64.tar.gz; then
```
mtmk, line 20:
```suggestion
  if [ "$rel" != "${rel#v}" ] && wget https://github.com/nats-io/nats-server/releases/download/$rel/nats-server-$rel-linux-amd64.tar.gz; then
```
All GitHub releases seems to start with `v`.

mtmk (approved):
Minor suggestion otherwise LGTM

test/test.c
```diff
@@ -16311,6 +16311,11 @@ test_ServerPoolUpdatedOnClusterUpdate(void)
         const char *urls[] = {"127.0.0.1:4222", "127.0.0.1:4223", "127.0.0.1:4224"};
         test("Check pool: ");
         s = _checkPool(conn, (char**)urls, (int)(sizeof(urls)/sizeof(char*)));
+        if (s != NATS_OK)
+        {
+            nats_Sleep(100);
```
levb, line 16316: @kozlovic if you don't mind a quick look here to confirm.

I think the the (connected) server was sending 2 INFO packets, the first was coming with obsolete list of URL, and then the second one would correct it. Not sure if there's a better way to wait for the second info packet to arrive in the test, but this seems to have fixed the flapping.

kozlovic: I am not able to reproduce, but _checkPool() is already sleeping and waiting until it gets the expected array, but I realize that we don't retry for the expected size (just the expected content). We could improve _checkPool instead of adding the sleep here...

levb: @kozlovic Thanks! please https://github.com/nats-io/nats.c/pull/680/commits/209e7ac27c9ef0e98f7b53b972fad2f046c1c532

test/test.c
```diff
@@ -7430,14 +7430,15 @@ _checkPool(natsConnection *nc, char **expectedURLs, int expectedURLsCount)
     bool    ok;
 
     natsMutex_Lock(nc->mu);
-    if (nc->srvPool->size != expectedURLsCount)
```
kozlovic, line 7433: You could still have this "if/print error" statement at the end (before releasing the mutex) so that we have this printed in case it helps.
```
    ...
    if (nc->srvPool->size != expectedURLsCount)
        printf("Expected pool size to be %d, got %d\n", expectedURLsCount, nc->srvPool->size);

    natsMutex_Unlock(nc->mu);
    return NATS_ERR;
}
```

kozlovic (approved):
You could still maintain the error message by moving it at the end of function. But LGTM.

test/test.c
levb: @kozlovic this seemed like a 1-liner to add, but upon further reading the function, I don't think the retry logic was quite correct, so I refactored the function.

kozlovic (approved):
LGTM

PR #694, opened 2023-11-11
tls_available option

biot2:
1. Added `tls_available` parameter understanding of `INFO` structure.
2. Fixed wrongful security option check causing incorrect error code `NATS_SECURE_CONNECTION_WANTED`


biot2:
I'm new to nats.c library, but it seems failed test has nothing to do with this pull request. Am I right? Should I do anything now?

vahid-nasehi:
@kozlovic

biot2:
@kozlovic

kozlovic (approved):
LGTM. It would have been nice to add the test that showed the failure to connect prior to the fix and verify that it passes now. But we can add that later. Thank you for your contribution.

biot2:
> LGTM. It would have been nice to add the test that showed the failure to connect prior to the fix and verify that it passes now. But we can add that later. Thank you for your contribution.

You're welcome. Actually the test module is so big and a bit complicated. I will try to add the test case and make sure everything else is as it should be.

kozlovic:
@biot2 I meant that maintainers could add the test, you don't have to especially if you are new to the project. I have added the test myself, you can see PR #697

PR #705, opened 2024-01-10
[FIXED] Random authentication error due to missing signature

tyler92:
Sometimes there is an authentication error due to a missing signature when `natsOptions_SetUserCredentialsFromMemory` is used:

```
[1] 2023/12/29 13:23:44.880932 [ERR] 172.18.0.1:33898 - cid:36 - authentication error
```

The problem is that _base64Encode doesn't do anything if the first byte is zero. Since the function works with strings and bytes, it makes sense to check `srcLen` instead of the first byte.


kozlovic (approved):
LGTM. But will wait for PR https://github.com/nats-io/nats.c/pull/708 to be merged before merging this one so that build on main branch does not fail.

PR #711, opened 2024-01-16
Fixed download of NATS Server for latest/main

kozlovic:
The "latest" was not working and would always use the one from nats.c.deps.
As for "main", I noticed lately that sometimes, we would get a binary of size 0, which would make all tests using the nats-server to fail. So make sure that we can run the server and get a valid version string. If not, repeat the test up to 5 times, otherwise fail the run.

### DIFF
--- a/buildOnTravis.sh
+++ b/buildOnTravis.sh
@@ -11,6 +11,7 @@ echo "build opts = " $3
 echo "test opts  = " $4
 
 if [ "$NATS_TEST_SERVER_VERSION" != "" ]; then
+  rm -fr $HOME/nats-server*
   rel=$NATS_TEST_SERVER_VERSION
   mkdir -p $HOME/nats-server-$rel
   if [ "$rel" = "latest" ]; then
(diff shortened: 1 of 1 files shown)


kozlovic:
@levb If you get a chance tomorrow (01/16/2024), could you review this one? Without this, the run that uses `main` will fail more often than not because the download of the binary fails, so I added few tries.

buildOnTravis.sh
```diff
@@ -11,6 +11,7 @@ echo "build opts = " $3
 echo "test opts  = " $4
 
 if [ "$NATS_TEST_SERVER_VERSION" != "" ]; then
+  rm -fr $HOME/nats-server*
```
levb, line 14: 👍

buildOnTravis.sh
```diff
@@ -19,14 +20,33 @@ if [ "$NATS_TEST_SERVER_VERSION" != "" ]; then
 
   if [ "$rel" != "${rel#v}" ] && wget https://github.com/nats-io/nats-server/releases/download/$rel/nats-server-$rel-linux-amd64.tar.gz; then
     tar -xzf nats-server-$rel-linux-amd64.tar.gz
-    mv nats-server-$rel-linux-amd64 $HOME/nats-server-$rel
+    mv nats-server-$rel-linux-amd64 nats-server
```
levb, line 23: @wallyqs should the retry logic be added to your script?

levb (approved):

PR #713, opened 2024-01-22
Modify JetStream examples.

pch-blog:
I have written this content by translating it using ChatGPT 3.5.

Developed and tested a prototype C/C++ program using the JetStream examples, resolved issues encountered during development, and added the necessary modifications to the examples.

- When directly converting the 'Subjects' variable, declared as 'char []', to 'const char **' and passing it to the 'js_AddStream' function, a SIGSEGV occurred, leading to program termination.
- 'SubjectsLen' is named in a way that suggests it can invoke the 'strlen' function, so I clarified in the comments that it represents the number of elements.
- It is necessary to set 'MaxBytes' to limit the memory size of the stream. 
   (In my test environment, when not setting the memory for the stream, the NATS.io SERVER's memory keeps increasing, causing the container to become unresponsive. 53 streams, each with one subject per stream.)

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -30,16 +30,17 @@ jobs:
           rm -rf ./html/*
           doxygen DoxyFile.NATS.Client
           git diff --quiet || echo "CHANGES=true" >> $GITHUB_OUTPUT
+          echo "SOURCE_TITLE=$(git log -1 --pretty=%s)" >> $GITHUB_OUTPUT
+          echo "SOURCE_SHA=$(git log -1 --pretty=%h)" >> $GITHUB_OUTPUT
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v6
         if: steps.update.outputs.CHANGES == 'true'
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          commit-message: Update documentation
-          branch: update-docs
-          title: Auto-generated documentation update [skip ci]
+          commit-message: "[skip ci] Update docs: ${{ steps.update.outputs.SOURCE_SHA }}: ${{ steps.update.outputs.SOURCE_TITLE }}"
+          branch: update-docs-${{ github.ref_name }}
+          title: Auto-generated documentation update for ${{ github.ref_name }}
           body: Documentation changes were detected. Please review and merge.
           labels: documentation
-          branch-suffix: timestamp
           delete-branch: true


examples/js-pub.c
```diff
@@ -86,9 +86,12 @@ int main(int argc, char **argv)
             // Initialize the configuration structure.
             jsStreamConfig_Init(&cfg);
             cfg.Name = stream;
-            // Set the subject
+            // Set the subject (Two-dimensional array)
             cfg.Subjects = (const char*[1]){subj};
+            // Set the subject count
```
levb, line 91: 👍

examples/js-pub.c
```diff
@@ -86,9 +86,12 @@ int main(int argc, char **argv)
...
             cfg.SubjectsLen = 1;
+            // Set the stream memory size
+            cfg.MaxBytes = (1024 * 1024) * 64;
```
levb, line 94: This example is minimal. There are many ways to restrict the lifetime of messages in a stream, I do not think we should be opinionated in the simple example.

1/5 maybe add a separate example on setting up the stream limits, and remove this line?

examples/js-sub.c
```diff
@@ -109,9 +109,12 @@ int main(int argc, char **argv)
             // Initialize the configuration structure.
             jsStreamConfig_Init(&cfg);
             cfg.Name = stream;
-            // Set the subject
+            // Set the subject (Two-dimensional array)
             cfg.Subjects = (const char*[1]){subj};
+            // Set the subject count
             cfg.SubjectsLen = 1;
+            // Set the stream memory size
+            cfg.MaxBytes = (1024 * 1024) * 64;
```
levb, line 117: same as above

levb (changes requested):
@pch-blog Thank you for the contribution. 

> * When directly converting the 'Subjects' variable, declared as 'char []', to 'const char **' and passing it to the 'js_AddStream' function, a SIGSEGV occurred, leading to program termination.

`Subjects` is indeed an array of C strings, each (0-terminated). Setting to a single string still requires an array of 1 pointer in it.

> * It is necessary to set 'MaxBytes' to limit the memory size of the stream.
>   (In my test environment, when not setting the memory for the stream, the NATS.io SERVER's memory keeps increasing, causing the container to become unresponsive. 53 streams, each with one subject per stream.)

This example illustrates a minimal stream configuration. There are several ways to restrict the message lifetime in a stream, I think it would be best to add a separate example for that.

examples/js-pub.c
pch-blog: thank you for the feedback!!

Upon reflection, I agree with the statement "I do not think we should be opinionated in the simple example."
However, if using 'js_MemoryStorage' for storage configuration, it seems necessary to add a comment explaining the message lifespan.

To provide further explanation, in my test environment, although I restricted the memory in the nats-server configuration, not configuring the message lifespan the same way as the current C client example resulted in the nats-server exceeding the configured memory. Additionally, it went beyond the container's total memory limit.

In Korea, there isn't much information on the NATS.io C client, and most of it is in English. So, it took a bit of time to code by referring to the [NATS Docs](https://docs.nats.io/) and [developer documentation](https://nats-io.github.io/nats.c/). I wanted to share the challenging parts I encountered.

Anyway, thank you once again for the feedback.

PR #720, opened 2024-03-06
[ADDED] GitHub action to update docs

levb:
@kozlovic @Jarema this is part of retiring `dev` and switching to releases from release branches. I am editing the release instructions now, will send the PR shortly for that. This GH action eliminates the need to re-generate the docs upon release.

This PR automatically updates the docs in `main` when needed (via a new PR). See https://github.com/levb/nats.c/pull/69 as an example, contains numerous changes caused by using a newer doxygen version.

Note that it will only fire when pushed to `main`, I debugged in my fork without the branch filter.

Example:

image

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -26,6 +26,7 @@ jobs:
         id: update
         run: |
           cd doc
+          rm -rf ./html/*
           doxygen DoxyFile.NATS.Client
           git diff --quiet || echo "CHANGES=true" >> $GITHUB_OUTPUT
 
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,0 +1,52 @@
+name: Update Docs
+
+on:
+  push:
+    branches:
+        - main
+
+jobs:
+  update-docs:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,44 @@
+name: Update Docs
+
+on:
+  push:
+    branches:
+        - main
```
kozlovic, line 6: Do we really need to regenerate doc for every push to main? Should it be done only for tags? I understand that the PR is submitted only if some changes are detected, so I guess this would still be fine. Maybe then should you "pin" the oxygen version so that this limits the changes simply due to the doxygen version change?

levb: Right, I thought about it but it'd be nice to use the newer doxygen - I think it's just a matter of approving the PR when it happens, and then we go back to quiet. I was a little concerned that there would be timestamps or other "random" changes, but so far so good.

How about if it becomes an issue, I pin it?

kozlovic (approved):
Sounds good. LGTM.

kozlovic (approved):
LGTM

PR #726, opened 2024-03-12
[ADDED] js_PauseConsumer support (also changed jsConsumerConfig, jsCo…

levb:
…nsumerInfo)

Fixes https://github.com/nats-io/nats.c/issues/717


test/test.c
```diff
@@ -24242,6 +24242,38 @@ test_JetStreamMgtConsumers(void)
...
+
+    if (serverVersionAtLeast(2, 11, 0))
+    {
+        test("Add consumer (non durable, paused): ");
+        cfg.PauseUntil = TIME_20350101;
+        s = js_AddConsumer(&ci, js, "MY_STREAM", &cfg, NULL, &jerr);
+        testCond((s == NATS_ERR) && (jerr == JSStreamNotFoundErr) && (ci == NULL));
+        nats_clearLastError();
+
+        test("Verify config: ");
```
kozlovic, line 24256: This is a bit confusing. You try to add a paused consumer above, but the check is that the operation should fail. And here you are checking for the config? Is it that you just want to check the marshaling and see that pause_until is properly set?

test/test.c
```diff
@@ -24242,6 +24242,38 @@ test_JetStreamMgtConsumers(void)
...
+        s = natsSubscription_NextMsg(&resp, sub, 1000);
+        testCond((s == NATS_OK) && (resp != NULL) && (strncmp(natsMsg_GetData(resp), "{\"stream_name\":\"MY_STREAM\","
```
kozlovic, line 24258: Assuming that what I mentioned above about checking consumer add request is correct: since we control the marshaling, it is likely that it will be in this order, etc.. but this test will have to be updated next time you add something to the consumer config. Would it not be enough to use `strstr` instead and look for `\"pause_until\":..`?

test/test.c
```diff
@@ -24333,6 +24365,35 @@ test_JetStreamMgtConsumers(void)
...
+    if (serverVersionAtLeast(2, 11, 0))
+    {
+        test("Pause consumer: ");
+        jsConsumerPauseResponse *cpr = NULL;
+        s = js_PauseConsumer(&cpr, js, "MY_STREAM", "my_name", TIME_20350101, NULL, &jerr);
+        testCond((s == NATS_OK) && (jerr == 0) && (cpr != NULL)
+            && cpr->Paused
+            && (cpr->PauseUntil == TIME_20350101)
+            && (cpr->PauseRemaining > 0));
+        NATS_FREE(cpr);
```
kozlovic, line 24377: Would use a `jsConsumerPauseResponse_Destroy` here.. will be more portable if one day some other field is present that would need to be freed.

test/test.c
```diff
@@ -24333,6 +24365,35 @@ test_JetStreamMgtConsumers(void)
...
+        cpr = NULL;
+
+        test("Verify consumer paused with GetInfo: ");
+        s = js_GetConsumerInfo(&ci, js, "MY_STREAM", "my_name", NULL, &jerr);
+        testCond((s == NATS_OK) && (jerr == 0) && (ci != NULL)
+            && ci->Paused
+            && (ci->PauseRemaining > 0));
+        NATS_FREE(ci);
```
kozlovic, line 24385: For sure here you need to use jsConsumerInfo_Destroy(ci). This is what is causing your memory leaks in tests.

test/test.c
```diff
@@ -24242,6 +24242,38 @@ test_JetStreamMgtConsumers(void)
...
+                                                                                     "\"pause_until\":" TIME_20350101_STR ","
+                                                                                     "\"replay_policy\":\"instant\",\"rate_limit_bps\":400,"
+                                                                                     "\"sample_freq\":\"60%%\",\"max_waiting\":500,\"max_ack_pending\":600,"
+                                                                                     "\"flow_control\":true,\"idle_heartbeat\":700,"
+                                                                                     "\"num_replicas\":1,\"mem_storage\":true}}",
+                                                              natsMsg_GetDataLength(resp)) == 0));
+        natsMsg_Destroy(resp);
+        cfg.PauseUntil = 0;
+        resp = NULL;
+    }
```
kozlovic, line 24275: I think you may also want to have a test that verifies that a consumer CAN be created paused. I think that it is possible.

src/jsm.c
```diff
@@ -3446,6 +3452,140 @@ js_DeleteConsumer(jsCtx *js, const char *stream, const char *consumer,
...
+    return NATS_UPDATE_ERR_STACK(s);
+}
+
+static natsStatus
+_marshalConsumerPauseReq(natsBuffer **new_buf, uint64_t pauseUntil)
+{
+    natsStatus s = NATS_OK;
+    natsBuffer *buf = NULL;
+
+
```
kozlovic, line 3516: Extra line here...

src/jsm.c
```diff
@@ -3446,6 +3452,140 @@ js_DeleteConsumer(jsCtx *js, const char *stream, const char *consumer,
...
+    if (s == NATS_OK)
+        *new_buf = buf;
+    else
+        natsBuf_Destroy(buf);
+
+    return NATS_UPDATE_ERR_STACK(s);
+}
+
+natsStatus
+js_PauseConsumer(jsConsumerPauseResponse **new_cpr, jsCtx *js,
```
kozlovic, line 3533: I would add a jsConsumerPauseResponse_Destroy(), which currently would be a simple `NATS_FREE` of the pointer, but will be more portable.

kozlovic (changes requested):

.travis.yml
```diff
@@ -63,6 +63,7 @@ jobs:
           packages:
             - g++-9
       env:
+        - NATS_TEST_SERVER_VERSION=main
```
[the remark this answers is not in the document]
levb: Yes, I was thinking about doing that, and maybe testing with a few older server releases too. Separate PR? Here, I just wanted to make sure that my new tests are running with checks in CI.

test/test.c
levb: Cut&pasting prior art FWIW, yes, that appears the intent. I'll move this to after the stream is created, and check the info, as you suggested elsewhere.

src/jsm.c
levb: done.

test/test.c
levb: Moved, create for real now

test/test.c
levb: Sort of not applicable after I moved the code, resolving

src/jsm.c
```diff
@@ -3446,6 +3452,149 @@ js_DeleteConsumer(jsCtx *js, const char *stream, const char *consumer,
...
+
+    IFOK(s, _marshalConsumerPauseReq(&buf, pauseUntil));
+
+    // Send the request
+    IFOK_JSR(s, natsConnection_Request(&resp, nc, subj, natsBuf_Data(buf), natsBuf_Len(buf), o.Wait));
+
+    // If we got a response, check for error or return the consumer info result.
+    IFOK(s, _unmarshalConsumerPauseResp(new_cpr, resp, errCode));
+
+    NATS_FREE(subj);
```
kozlovic, line 3586: You are missing a `natsBuf_Destroy(buf);` here.

kozlovic (changes requested):

kozlovic (approved):
LGTM!

PR #734, opened 2024-03-18
Modify JetStream examples (Add comments to SubjectsLen)

pch-blog:
'SubjectsLen' is named in a way that suggests it can invoke the 'strlen' function, so I clarified in the comments that it represents the number of elements.


levb (approved):
LGTM, and thank you! 🎉

Issue #64, opened 2017-04-10
What about Android


rFronteddu:
Hello I was wondering if you tried compiling cnats for android.

kozlovic:
I did not. Not sure if anybody in the community has.

rFronteddu:
I'll give it a try this weekend.

rFronteddu:
I made it, **I had to add #include <netinet/in.h> in n-unix** since without that include the compiler wasn't able to find the IPTCP macro in setops. I followed the ndk-build path since I am more familiar with it. It shouldn't be hard at all to make a cmake option for one more experienced with cmake.

I just compiled a simple main executable which uses the natsconnect method.  I still haven't tested the executable but I'll do that soon.

Android.mk -> https://pastebin.com/eLEC49DL
Application.mk -> https://pastebin.com/TRVs8WuH

psantiagol:
Hi, I am compiling nats for Android with cmake. I managed to do so, but I had to manually comment this line in CMakeLists.txt set(NATS_EXTRA_LIB "rt -lpthread") as this lib is not available in Android. It would be nice if cmake workflow had into account this casuistic.

kozlovic:
@psantiagol Could you try with removing just the `-lpthread` part? This was recently added to solve a CentOS build issue in this PR: https://github.com/nats-io/nats.c/pull/710.

psantiagol:
Hi! in fact what is causing the failure is the rt one, as in Android environment isn't found such library.

kozlovic:
@psantiagol Were you referring just to `NATS_EXTRA_LIB` or even `NATS_USE_PTHREAD` (the line above)?

Could you try something like:
```
else(APPLE)
   set(NATS_OS "LINUX")
   set(NATS_USE_PTHREAD "-pthread")
   if(NOT ANDROID)
     set(NATS_EXTRA_LIB "rt -lpthread")
   endif()
endif(APPLE)
```
and see if that works? If so, I will submit a PR. If you meant to also include the `NATS_USE_PTHREAD` line, please move the `if(NOT ANDROID)` a line above and let me know.

psantiagol:
@kozlovic I tested both. In both cases compiled and ran correctly, so I assume set(NATS_USE_PTHREAD "-pthread") line is not necessary neither.

kozlovic:
[...]

PR #739, opened 2024-03-26
[FIXED] Build: issue on Android with `NATS_EXTRA_LIB`

kozlovic:
A user (@psantiagol) reported an issue (in #64) building on Android
due to the use of `NATS_EXTRA_LIB`. This PR skip the definition of
this variable if building on Android. Should other users then report
that this is breaking their build on Android, we will have to make
this a conditional based on a boolean build option.

Also, fixed test JetStreamBackOffRedeliveries.
With PR https://github.com/nats-io/nats-server/pull/5242 in server
code, not providing a MaxDeliver is considered unlimited and should
not error out. So fixing test to set MaxDeliver <= to the backoff
list to get the expected error.
    

levb (approved):
LGTM

levb:
Cherry-picked for `v3.8.1`

Biswa96:
> Should other users then report
> that this is breaking their build on Android, we will have to make
> this a conditional based on a boolean build option.

IMO, this is the correct condition. termux-packages was carrying a patch for this issue for 2 years https://github.com/termux/termux-packages/blob/019cdbab029f468c8a370f78b8cdacc1b5dbd24e/packages/libnats-c/CMakeLists.txt.patch

Issue #736, opened 2024-03-22
Compile error on Linux


KonstantinWinkel:
### Observed behavior

```[ 30%] Linking C shared library libnats.so
/usr/bin/ld: CMakeFiles/nats.dir/buf.c.o:(.bss+0x0): multiple definition of `micro_ErrorInvalidArg'; CMakeFiles/nats.dir/asynccb.c.o:(.bss+0x0): first defined here
/usr/bin/ld: CMakeFiles/nats.dir/buf.c.o:(.bss+0x8): multiple definition of `micro_ErrorOutOfMemory'; CMakeFiles/nats.dir/asynccb.c.o:(.bss+0x8): first defined here
```

this error comes up for every file that at some point includes "nats.h"

### Expected behavior

No error, just compilation

### Server and client version

Latest git version ( commit 49c7f32cb9f307f909aa69d35a6f88d2bc884af4 )

### Host environment

Pop!_OS 22.04 LTS x86_64

### Steps to reproduce

```
git clone https://github.com/nats-io/nats.c
cd nats.c
mkdir build && cd build
cmake .. -DNATS_BUILD_STREAMING=OFF -DNATS_BUILD_WITH_TLS=OFF
sudo make
```

*PROPOSED FIX*: 
```nats.h line 49:   #define NATS_EXTERN extern```

PR #731, opened 2024-03-14
Moved micro_args to examples

levb:
Fixes https://github.com/nats-io/nats.c/issues/701

PR #743, opened 2024-04-11
[FIXED] GH-736: added extern to micro_Errors

levb:
In https://github.com/nats-io/nats.c/pull/731 I wrongly assumed that NATS_EXTERN was a superset of `extern` and it is not.

Unfortunately, our current CI set up does not catch these, I will revive the GH actions PR that used the more recent compiler versions.

Fixes https://github.com/nats-io/nats.c/issues/736


kozlovic (approved):
LGTM

PR #750, opened 2024-04-25
[ADDED] kvStore_WatchMulti, js_Subscribe[Sync]Multi

levb:
Added `kvStore_WatchMulti`, `js_SubscribeSyncMulti`, `js_SubscribeMulti` APIs to use multi-subject consumers.

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -2790,6 +2790,23 @@ js_Subscribe(natsSubscription **sub, jsCtx *js, const char *subject,
 }
 
 natsStatus
+js_SubscribeMulti(natsSubscription **sub, jsCtx *js, const char **subjects, int numSubjects,
+                  natsMsgHandler cb, void *cbClosure,
+                  jsOptions *jsOpts, jsSubOptions *opts, jsErrCode *errCode)
+{
+    natsStatus s;
+
+    if (errCode != NULL)
+        *errCode = 0;
+
+    if (cb == NULL)
+        return nats_setDefaultError(NATS_INVALID_ARG);
+
+    s = _subscribeMulti(sub, js, subjects, numSubjects, NULL, cb, cbClosure, false, jsOpts, opts, errCode);
+    return NATS_UPDATE_ERR_STACK(s);
+}
+
+natsStatus
 js_SubscribeSync(natsSubscription **sub, jsCtx *js, const char *subject,
                  jsOptions *jsOpts, jsSubOptions *opts, jsErrCode *errCode)
 {
--- a/src/js.c
+++ b/src/js.c
@@ -2480,7 +2480,7 @@ _subscribeMulti(natsSubscription **new_sub, jsCtx *js, const char **subjects, in
     // that is, if user did not provide a `Stream` name through options).
     if (nats_IsStringEmpty(stream) && numSubjects > 0)
     {
-        // Ise the first subject to find the stream.
+        // Use the first subject to find the stream.
         s = _lookupStreamBySubject(&stream, nc, subjects[0], &jo, errCode);
         if (s != NATS_OK)
             goto END;
--- a/src/js.c
+++ b/src/js.c
@@ -2598,15 +2598,7 @@ PROCESS_INFO:
                     s = nats_setDefaultError(NATS_NO_MEMORY);
             }
             IF_OK_DUP_STRING(s, jsi->stream, stream);
-            if ((s == NATS_OK) && numSubjects > 0)
-            {
-                int l = nats_printStringArray(NULL, 0, subjects, numSubjects);
-                jsi->psubj = NATS_CALLOC(l, 1); // '\0' included
-                if (jsi->psubj == NULL)
-                    s = nats_setDefaultError(NATS_NO_MEMORY);
-                if (s == NATS_OK)
-                    nats_printStringArray(jsi->psubj, l, subjects, numSubjects);
-            }
+            IFOK(s, nats_formatStringArray(&jsi->psubj, subjects, numSubjects));
             if (s == NATS_OK)
             {
                 jsi->js     = js;
--- a/src/js.c
+++ b/src/js.c
@@ -2195,38 +2195,68 @@ _checkConfig(jsConsumerConfig *s, jsConsumerConfig *u)
 
 static natsStatus
 _processConsInfo(const char **dlvSubject, jsConsumerInfo *info, jsConsumerConfig *userCfg,
-                 bool isPullMode, const char *subj, const char *queue)
+                 bool isPullMode, const char **subjects, int numSubjects, const char *queue)
 {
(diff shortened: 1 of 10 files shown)


src/js.c
```diff
@@ -2197,38 +2197,68 @@ _checkConfig(jsConsumerConfig *s, jsConsumerConfig *u)
...
+        subjects = NULL;
+        numSubjects = 0;
+    }
 
-    // Make sure this new subject matches or is a subset.
-    if (!nats_IsStringEmpty(subj))
+    // Match the subjects against the consumer's filter subjects.
+    if (numSubjects > 0 && filterSubjectsLen > 0)
     {
-        if (nats_IsStringEmpty(ccfg->FilterSubject) && (ccfg->FilterSubjectsLen == 0))
```
levb, line 2214: This was done incorrectly before this change, now restore the prior logic of matching exactly only if both subjects and filter subject(s) are set.

src/util.c
```diff
@@ -2555,3 +2555,56 @@ nats_freeMetadata(natsMetadata *md)
     md->List = NULL;
     md->Count = 0;
 }
+
+
+int nats_printStringArray(char *out, int bufLen, const char **strings, int count)
```
levb, line 2560: TODO: add a test

src/js.c
```diff
@@ -2559,8 +2598,15 @@ _subscribe(natsSubscription **new_sub, jsCtx *js, const char *subject, const cha
                     s = nats_setDefaultError(NATS_NO_MEMORY);
             }
             IF_OK_DUP_STRING(s, jsi->stream, stream);
-            if ((s == NATS_OK) && !nats_IsStringEmpty(subject))
-                DUP_STRING(s, jsi->psubj, subject);
+            if ((s == NATS_OK) && numSubjects > 0)
+            {
+                int l = nats_printStringArray(NULL, 0, subjects, numSubjects);
```
kozlovic, line 2603: First I think the `print` is a bit confusing, but why not go with a simple `natsBuf` and just call append with each string/byte for the comma, the same way we do when marshaling?

src/js.c
```diff
@@ -2764,12 +2764,12 @@ _subscribe(natsSubscription **new_sub, jsCtx *js, const char *subject, const cha
...
 {
+    natsStatus s = NATS_OK;
     const char *subjects[] = {subject};
+    int numSubjects = nats_IsStringEmpty(subject) ? 0 : 1;
 
-    if (nats_IsStringEmpty(subject))
-        return _subscribeMulti(new_sub, js, NULL, 0, pullDurable, usrCB, usrCBClosure, isPullMode, jsOpts, opts, errCode);
-
-    return _subscribeMulti(new_sub, js, subjects, 1, pullDurable, usrCB, usrCBClosure, isPullMode, jsOpts, opts, errCode);
+    s = _subscribeMulti(new_sub, js, subjects, numSubjects, pullDurable, usrCB, usrCBClosure, isPullMode, jsOpts, opts, errCode);
```
kozlovic, line 2771: Not sure... looks like there is a case in _subscribeMulti where if numSubjects == 0 we would still set cfg->FilterSubjects to subjects (which in this case would not be NULL). But cfg->FilterSubjectsLen will be correctly set to 0, which I believe all other code use the len before accessing the FilterSubjects field. But to be safe, I think if numSubjects == 0, we should have subjects set to NULL. If that is not "clean" to do, then revert to check of subject being empty and invoking with different values.

kozlovic (changes requested):
I assume that you are going to update PR with replacement for "printStringArray", but wanted to start commenting.

src/js.c
levb: Yes, that was my understanding as well, but cleared it out now as you suggested.

src/util.c
```diff
@@ -2555,3 +2555,61 @@ nats_freeMetadata(natsMetadata *md)
...
+        }
+        IFOK(s, natsBuf_AppendByte(&buf, '"'));
+    }
+
+    IFOK(s, natsBuf_AppendByte(&buf, ']'));
+    IFOK(s, natsBuf_AppendByte(&buf, '\0'));
+    
+    if (s != NATS_OK)
+    {
+        return s;
```
kozlovic, line 2610: You should still call `natsBuf_Cleanup(&buf);` here in case allocation occurred.

test/test.c
```diff
@@ -5466,6 +5466,31 @@ test_natsSrvVersionAtLeast(void)
...
+{
+    natsStatus s;
+    char *str = NULL;
+
+    test("Check empty: ");
+    s = nats_formatStringArray(&str, NULL, 0);
+    testCond((s == NATS_OK) && (str != NULL) && (strcmp(str, "[]") == 0));
+    NATS_FREE(str);
+    str = NULL;
+
```
kozlovic, line 5480: Since you account for `NULL` string in the array (and format as `(null)`), should you add a test for that?

test/test.c
```diff
@@ -5466,6 +5466,31 @@ test_natsSrvVersionAtLeast(void)
...
+    test("Check one: ");
+    const char *oneArray[] = {"one"};
+    s = nats_formatStringArray(&str, oneArray, 1);
+    testCond((s == NATS_OK) && (str != NULL) && (strcmp(str, "[\"one\"]") == 0));
+    NATS_FREE(str);
+
```
kozlovic, line 5486: Would set `str = NULL;` after the free above.

kozlovic (changes requested):

test/test.c
levb: just did

test/test.c
levb: just changed it to also cover NATS_FREE_STRINGS, so N/A

test/test.c
```diff
@@ -5470,25 +5470,29 @@ static void
...
+    const char *threeArray[] = {"one", "two", "three"};
+    s = nats_formatStringArray(&array[2], threeArray, 3);
+    testCond((s == NATS_OK) && (array[2] != NULL) && (strcmp(array[2], "[\"one\",\"two\",\"three\"]") == 0));
+
+    test("Check NULL: ");
+    const char *nullArray[] = {NULL};
+    s = nats_formatStringArray(&array[3], nullArray, 1);
+    testCond((s == NATS_OK) && (array[3] != NULL) && (strcmp(array[3], "[\"(null)\"]") == 0));
+
+    NATS_FREE_STRINGS(array, 3);
```
kozlovic, line 5495: Should it not be 4?

kozlovic (changes requested):

test/test.c
levb: grrr, yes, will see if the new build catches it; I didn't run valgrind myself.

kozlovic (approved):
LGTM

PR #753, opened 2024-05-01
[TEST only] increased Test_KeyValueMirrorCrossDomains timeout value

levb:
Fixes test failure against server `main`


jnmoyne (approved):
LGTM

kozlovic (approved):
LGTM. And that's because the retry interval was increased in the server, correct?

jnmoyne:
> LGTM. And that's because the retry interval was increased in the server, correct?

The HB interval for source/mirror consumers was increased from 1s to 5s yes

PR #756, opened 2024-05-03
[Added] natsConnection_Reconnect

levb:
Implements https://github.com/nats-io/nats-architecture-and-design/issues/259

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3319,9 +3319,9 @@ natsConn_create(natsConnection **newConn, natsOptions *options)
 natsStatus
 natsConnection_Connect(natsConnection **newConn, natsOptions *options)
 {
-    natsStatus s = NATS_OK;
-    natsConnection *nc = NULL;
-    natsOptions *opts = NULL;
+    natsStatus      s       = NATS_OK;
+    natsConnection  *nc     = NULL;
+    natsOptions     *opts   = NULL;
 
     if (options == NULL)
     {
--- a/src/conn.c
+++ b/src/conn.c
@@ -81,8 +81,14 @@ _processConnInit(natsConnection *nc);
 static void
 _close(natsConnection *nc, natsConnStatus status, bool fromPublicClose, bool doCBs);
 
-static bool
-_processOpError(natsConnection *nc, natsStatus s, bool initialConnect);
+static natsStatus
+_tryReconnect(natsConnection *nc, natsStatus s, bool forcedReconnect, bool *started);
+
+static void
+_maybeReconnect(natsConnection *nc, natsStatus s) { _tryReconnect(nc, s, false, NULL); }
+
+static natsStatus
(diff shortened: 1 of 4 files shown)


src/conn.c
```diff
@@ -2134,29 +2142,30 @@ _evStopPolling(natsConnection *nc)
...
+            return NATS_OK;
         }
     }
 
     // Do reconnect only if allowed and we were actually connected
     // or if we are retrying on initial failed connect.
-    if (initialConnect || (nc->opts->allowReconnect && (nc->status == NATS_CONN_STATUS_CONNECTED)))
+    if (forcedReconnect || (nc->opts->allowReconnect && (nc->status == NATS_CONN_STATUS_CONNECTED)))
     {
-        natsStatus ls = NATS_OK;
```
kozlovic, line 2158: There was a reason we used a different status here. What we wanted to report as the error was the error that was given to processOpError(), not an internal one that we may encounter trying to setup the reconnect.

src/conn.c
```diff
@@ -3333,6 +3346,18 @@ natsConnection_Connect(natsConnection **newConn, natsOptions *options)
...
 
+natsStatus
+natsConnection_Reconnect(natsConnection *nc)
+{
+    natsStatus s = NATS_OK;
+
+    if (natsConnection_IsClosed(nc))
+        return nats_setDefaultError(NATS_INVALID_ARG);
+
+    IFOK(s, _forceReconnect(nc, NATS_OK, NULL));
```
kozlovic, line 3357: There is no need for the `IFOK` here since there is no change of `s` in lines above.

But overall, I think these changes are way more complicated than it needs to be (or I am missing something). If we want to cause a reconnect, locking the connection and and if the socket is valid, closing it should be enough, no?

Of course, verification that the connection is set in such a way that it allows for reconnection (I don't recall if there are options that would totally prevent a reconnect) before doing so, and if not, returning an error indication that reconnect() failed because the connection would not reconnect because of options setting, would be best.

kozlovic (changes requested):
Could you review my comment and see if we could make the changes only in the new function (no existing code would need to be changed)?

src/conn.c
levb: right, and I think my code still does that (`newErr`), but in the case of a natsConnection_Reconnect failing on a to fire off doReconnect we want to return the actual failure, right? (like, out of memory for instance)

src/conn.c
levb: > way more complicated than it needs to be (or I am missing something). If we want to cause a reconnect, locking the connection and and if the socket is valid, closing it should be enough, no?

@kozlovic you're likely right, I was modeling after the `go` changes that call an existing re-connect method, and I figured this to be the closest. I'll try what you suggested, in a separate branch.

levb: See https://github.com/nats-io/nats.c/pull/757

PR #757, opened 2024-05-03
[Added] natsConnection_Reconnect

levb:
Simplified version of https://github.com/nats-io/nats.c/pull/756

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -3319,9 +3319,9 @@ natsConn_create(natsConnection **newConn, natsOptions *options)
 natsStatus
 natsConnection_Connect(natsConnection **newConn, natsOptions *options)
 {
-    natsStatus s = NATS_OK;
-    natsConnection *nc = NULL;
-    natsOptions *opts = NULL;
+    natsStatus      s       = NATS_OK;
+    natsConnection  *nc     = NULL;
+    natsOptions     *opts   = NULL;
 
     if (options == NULL)
     {
--- a/src/conn.c
+++ b/src/conn.c
@@ -2139,14 +2139,9 @@ _evStopPolling(natsConnection *nc)
 static bool
 _processOpError(natsConnection *nc, natsStatus s, bool initialConnect)
 {
-    bool forceReconnect;
-
     natsConn_Lock(nc);
 
-    forceReconnect = initialConnect || nc->forceReconnect;
-    nc->forceReconnect = false;
-
-    if (!forceReconnect)
+    if (!initialConnect)
     {
         if (_isConnecting(nc) || natsConn_isClosed(nc) || (nc->inReconnect > 0))
         {
@@ -2158,7 +2153,7 @@ _processOpError(natsConnection *nc, natsStatus s, bool initialConnect)
 
     // Do reconnect only if allowed and we were actually connected
     // or if we are retrying on initial failed connect.
-    if (forceReconnect || (nc->opts->allowReconnect && (nc->status == NATS_CONN_STATUS_CONNECTED)))
+    if (initialConnect || (nc->opts->allowReconnect && (nc->status == NATS_CONN_STATUS_CONNECTED)))
     {
         natsStatus ls = NATS_OK;
 
@@ -3440,7 +3435,9 @@ natsConnection_Reconnect(natsConnection *nc)
         return nats_setDefaultError(NATS_CONNECTION_CLOSED);
     }
 
-    nc->forceReconnect = true;
+    if (!nc->opts->allowReconnect)
+        return nats_setDefaultError(NATS_INVALID_ARG);
+
     natsSock_Close(nc->sockCtx.fd);
 
     natsConn_Unlock(nc);
--- a/src/conn.c
+++ b/src/conn.c
@@ -81,14 +81,8 @@ _processConnInit(natsConnection *nc);
 static void
 _close(natsConnection *nc, natsConnStatus status, bool fromPublicClose, bool doCBs);
 
-static natsStatus
-_tryReconnect(natsConnection *nc, natsStatus s, bool forcedReconnect, bool *started);
-
-static void
-_maybeReconnect(natsConnection *nc, natsStatus s) { _tryReconnect(nc, s, false, NULL); }
-
-static natsStatus
-_forceReconnect(natsConnection *nc, natsStatus s, bool *started) { return _tryReconnect(nc, s, true, started); }
+static bool
+_processOpError(natsConnection *nc, natsStatus s, bool initialConnect);
 
 static natsStatus
 _flushTimeout(natsConnection *nc, int64_t timeout);
@@ -2108,9 +2102,7 @@ _connect(natsConnection *nc)
     {
         natsConn_Unlock(nc);
 
-        bool reconnectStarted = false;
-        s = _forceReconnect(nc, retSts, &reconnectStarted);
-        if ((s == NATS_OK) && reconnectStarted)
+        if (_processOpError(nc, retSts, true))
         {
             nats_clearLastError();
             return NATS_NOT_YET_CONNECTED;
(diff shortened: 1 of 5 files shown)


src/conn.c
```diff
@@ -2139,9 +2139,14 @@ _evStopPolling(natsConnection *nc)
 static bool
 _processOpError(natsConnection *nc, natsStatus s, bool initialConnect)
 {
+    bool forceReconnect;
+
     natsConn_Lock(nc);
 
-    if (!initialConnect)
+    forceReconnect = initialConnect || nc->forceReconnect;
```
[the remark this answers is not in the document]
levb: The `go` client disregards/overrides its equivalent of `nc->opts->allowReconnect` when `Disconnect()` is invoked, I posed a question about it in https://github.com/nats-io/nats.go/pull/1624#issuecomment-2091859745.

If we were to error out when not `nc->opts->allowReconnect` - no changes would be needed, otherwise I need to override it (present implementation).

src/conn.c
```diff
@@ -3422,6 +3427,28 @@ natsConnection_ConnectTo(natsConnection **newConn, const char *url)
...
+    natsConn_Lock(nc);
+    if (natsConn_isClosed(nc))
+    {
+        natsConn_Unlock(nc);
+        return nats_setDefaultError(NATS_CONNECTION_CLOSED);
+    }
+
+    nc->forceReconnect = true;
+    natsSock_Close(nc->sockCtx.fd);
+    nc->sockCtx.fd = NATS_SOCK_INVALID;
```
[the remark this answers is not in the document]
levb: natsSock_Close already checks for NATS_SOCK_INVALID

src/conn.c
```diff
@@ -2139,9 +2139,14 @@ _evStopPolling(natsConnection *nc)
...
 {
+    bool forceReconnect;
+
     natsConn_Lock(nc);
 
-    if (!initialConnect)
+    forceReconnect = initialConnect || nc->forceReconnect;
+    nc->forceReconnect = false;
+
+    if (!forceReconnect)
```
kozlovic, line 2149: Still, even if we override `allowReconnect`, I don't see why here we would change this line because the check of `_isConnecting()`, `natsConn_isClosed()`, etc.. is still valid. Meaning that we should not attempt to reconnect if any of these conditions are true.

src/conn.c
```diff
@@ -2153,7 +2158,7 @@ _processOpError(natsConnection *nc, natsStatus s, bool initialConnect)
 
     // Do reconnect only if allowed and we were actually connected
     // or if we are retrying on initial failed connect.
-    if (initialConnect || (nc->opts->allowReconnect && (nc->status == NATS_CONN_STATUS_CONNECTED)))
+    if (forceReconnect || (nc->opts->allowReconnect && (nc->status == NATS_CONN_STATUS_CONNECTED)))
```
kozlovic, line 2161: Again, in my mind, we should not need that, we would fall in the nc->opts->allowReconnect and connected case if we were to change `natsConnection_Reconnect()` to check that we are allowed to do this and if not fail the call.

src/conn.c
```diff
@@ -3422,6 +3427,26 @@ natsConnection_ConnectTo(natsConnection **newConn, const char *url)
...
+        return nats_setDefaultError(NATS_INVALID_ARG);
+
+    natsConn_Lock(nc);
+    if (natsConn_isClosed(nc))
+    {
+        natsConn_Unlock(nc);
+        return nats_setDefaultError(NATS_CONNECTION_CLOSED);
+    }
+
+    nc->forceReconnect = true;
```
kozlovic, line 3443: See my earlier comment. I understand that this function use-case is to refresh auth, but it seems to me a bit weird that since we have an option that explicitly configures the connection to reject reconnect, we would allow a call `natsConnection_Reconnect()` to work and override this option. Meaning that if users wanted to use `natsConnection_Reconnect()` they would probably not use the option to explicitly disallow reconnection.

src/nats.h
```diff
@@ -4033,6 +4033,18 @@ stanMsg_Destroy(stanMsg *msg);
 NATS_EXTERN natsStatus
 natsConnection_Connect(natsConnection **nc, natsOptions *options);
 
+/** \brief Causes the client to drop the connection to the current server and
+ * perform standard reconnection process.
+ *
+ * This means that all subscriptions and consumers should be resubscribed and
+ * their work resumed after successful reconnect where all reconnect options are
```
kozlovic, line 4040: Well, actually not `all reconnect options` if we consider that "allow reconnect" is a reconnect option ;-). So I would argue that either the Go client should be changed, or the spec modified to specifically cover cases of libraries have an option to disable reconnection (maybe not all do that).

kozlovic (changes requested):

src/conn.c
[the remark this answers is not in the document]
levb: Ok, but that also means that the user would not be able to reconnect only "manually", the connection will auto-reconnect on all relevant errors.

I'll make the change now.

src/conn.c
levb: It's the next check, where we wouldn't attempt to reconnect unless allowReconnect is set. A false negative (vs. a false positive).

src/conn.c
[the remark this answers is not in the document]
kozlovic: Sorry I don't understand. The original check here says to check for the current connection status so that we don't attempt to reconnect if we are in the middle of it. The check for `!initialConnect` was likely introduced when adding the ability to connect asynchronously (that is, call `natsConnection_Create()` (or Connect) while a server is not running does not fail, but attempt the reconnect right away.

src/conn.c
kozlovic: But that's my point. If allowReconnect is false, then `natsConnection_Reconnect()` should check for that and return an error (and not close the socket obviously). But that's something that needs to be discussed/agreed with the rest of the team.

kozlovic: One thing that the Go client is doing that we may want to do (and so, we may need some information that we are in the process of a "force reconnect") is that we could skip the wait that we have when attempting to reconnect. The "force reconnect" flag should then be reset only in the "doReconnect" function, when successfully reconnected, and likely checked in `natsConnection_Reconnect()` to be a no-op if that flag is currently set to true (which means that it would still be in progress).

levb: yes, done. I'll double-check this semantics with @piotrpio and @Jarema.

src/conn.c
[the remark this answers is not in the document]
levb: I just meant that one won't be able to make a connection with allowReconnect==false if so desired, but then forcefully `Reconnect()` on it (say if they want to re-auth?). That was my understanding of the use-case.

src/conn.c
```diff
@@ -3422,6 +3422,28 @@ natsConnection_ConnectTo(natsConnection **newConn, const char *url)
...
+
+    natsConn_Lock(nc);
+    if (natsConn_isClosed(nc))
+    {
+        natsConn_Unlock(nc);
+        return nats_setDefaultError(NATS_CONNECTION_CLOSED);
+    }
+
+    if (!nc->opts->allowReconnect)
+        return nats_setDefaultError(NATS_INVALID_ARG);
```
kozlovic, line 3439: You need to "unlock". You could also use a setError where you can provide an error text, to be more telling. And I think `NATS_ILLEGAL_STATE` may be more telling?

src/conn.c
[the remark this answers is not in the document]
kozlovic: That's why maybe don't do the changes until it is agreed with the rest of the team. I did not mean to have you make all the changes if you have to put it back because this is how the team wants the feature to work.

If as a team it is agreed that it should override, then simply add to the doc that this function ignores the "allow reconnect" option (currently it says that it respects them all).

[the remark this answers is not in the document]
levb: (will not merge until verified)

kozlovic (approved):
LGTM, but ping me for another review if the team decides to override the allow reconnect option.

levb:
https://synadiacommunications.slack.com/archives/C05CGNTKNF2/p1715003032479599?thread_ts=1715000908.579469&cid=C05CGNTKNF2

image

So, so special treatment, do not return an error from `Reconnect` if `allowReconnect==false`, just kill the socket and let the client do what it does (error out).

src/conn.c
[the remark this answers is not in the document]
levb: See the comment from the team feedback: ignore `allowReconnect` altogether, just close the socket and let the client do what it does.

kozlovic (approved):
LGTM. (Unrelated: the kv cross domains test is taking 20+ secs now and since you repeat the tests 3 times for some runs, it starts to add up. May want to see if that can be optimized in the test or if this is just the result from the server updates)

levb:
@kozlovic yes, I knew it when I committed. I was hoping to soon get the word from @jnmoyne that the timeout can be shortened, but yes, let me at least make all the "long" tests run once. TBH I am yet to see any new flakes appear from the x3 change, so maybe it should be changed back to x1 for all, like it was before.

PR #759, opened 2024-05-06
[CI only] run tests once, not x3 except for main, release

levb:
`main` and `release_**` will still run 3x, but all PR jobs will use 1x


kozlovic (approved):
LGTM

PR #763, opened 2024-06-12
Support linking with MinGW toolchains on Linux

XJ-0461:
In the top level CMakeLists file, the line:
`  set(NATS_EXTRA_LIB "Ws2_32")`

Causes linker errors when cross-compiling cnats to Windows under Linux using the mingw toolchains. Because of the uppercase 'W'.

`/usr/lib/gcc/x86_64-w64-mingw32/14.1.1/../../../../x86_64-w64-mingw32/bin/ld: cannot find -lWs2_32: No such file or directory`

On linux libraries are case-sensitive, and can be found (depending on your toolchain choice) at:
- /usr/x86_64-w64-mingw32/sys-root/mingw/lib/libws2_32.a
- /usr/i686-w64-mingw32/sys-root/mingw/lib/libws2_32.a
- /usr/x86_64-w64-mingw32ucrt/sys-root/mingw/lib/libws2_32.a

On Windows case-sensitivity is not a problem, so changing to lowercase "ws2_32" is fine.


XJ-0461:
@mtmk @levb. If you (or someone you know) could check this out I'd appreciate it.

mtmk (approved):
thanks @XJ-0461. I can confirm it's an issue on Linux mingw build and this PR fixes it. Also, Windows build isn't affected.

LGTM

kozlovic (approved):
LGTM

PR #764, opened 2024-06-13
[TEST ONLY] Fixed Test_JetStreamSubscribeIdleHeartbeat

levb:
https://github.com/nats-io/nats-server/pull/5482/files#diff-c6e9a3ce0d55948f2e7ea75a4de309986654f2ab11e0f8750625b88b28bd9ebfR2833-R2838 broke this test. The test was faking a JetStream-delivered message, resulting in inaccurate ACKs being sent to the server. Disabled auto-ACKs for the relevant messages to avoid messing up the server state, and the test is now passing.

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -27965,11 +27965,11 @@ test_JetStreamSubscribeIdleHearbeat(void)
         s = natsConnection_PublishString(nc, inbox, (_msg));               \
     }
 
-#define PUBLISH_FAKE_RESET()      \
-    {                             \
-        natsSub_Lock(sub);        \
-        sub->jsi->ackNone = true; \
-        natsSub_Unlock(sub);      \
+#define PUBLISH_FAKE_RESET()       \
+    {                              \
+        natsSub_Lock(sub);         \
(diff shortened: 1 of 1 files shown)


test/test.c
```diff
@@ -27951,18 +27951,35 @@ test_JetStreamSubscribeIdleHearbeat(void)
...
+        natsSub_Unlock(sub);                                               \
+                                                                           \
+        natsConn_setFilterWithClosure(nc, _setMsgReply, (void *)(_reply)); \
+        s = natsConnection_PublishString(nc, inbox, (_msg));               \
+    }
+
+#define PUBLISH_FAKE_RESET()      \
+    {                             \
+        natsSub_Lock(sub);        \
+        sub->jsi->ackNone = true; \
```
mtmk, line 27971: did you mean to say `false` here?

levb: yes, TY. Wonder that it still passses with no auto-acks.

mtmk (approved):
LGTM

Issue #762, opened 2024-06-11
Support URL (%-) encoding in server URL strings


levb:
### Observed behavior

Presently, the server URLs passed into `natsConnection_ConnectTo`, `natsOptions_SetURL`, or `natsOptions_SetServers`.

### Expected behavior

%-encoded characters should be supported, specifically in username and password.

### Server and client version

2.10, 3.8

### Steps to reproduce

%-encode a username with special characters in a URL, and see the connection fail

PR #765, opened 2024-06-13
[ADDED] Support %-encoded username/password in server URLs

levb:
Fixes https://github.com/nats-io/nats.c/issues/762


test/test.c
```diff
@@ -1764,6 +1764,30 @@ test_natsUrl(void)
...
+    test("'tcp://%4c%65v:p%@localhost':");
+    s = natsUrl_Create(&u, "tcp://%4c%65v:p%@localhost");
+    testCond((s == NATS_ERR) && (u == NULL) && (strstr(nats_GetLastError(NULL), "invalid percent encoding in URL: p%") != NULL));
+    nats_clearLastError();
+
+    test("'tcp://%4H%65v:p%@localhost':");
+    s = natsUrl_Create(&u, "tcp://%4H%65v:p%@localhost");
+    testCond((s == NATS_ERR) && (u == NULL) && (strstr(nats_GetLastError(NULL), "invalid percent encoding in URL: %4H%65v") != NULL));
+    nats_clearLastError();
+
```
mtmk, line 1790: could you also test it for special characters e.g. `%00`? not sure if that may be an issue

levb: Added. %00 makes the rest of the string be still decoded, but since it is a C string, the 0 byte terminates it.

mtmk (approved):
LGTM

PR #767, opened 2024-06-21
{CI only} Use pull_request_target instead of pull_request

levb:
We are having an issue running coverage upload from forked repo PRs. This should solve it, I think.

Testing for now from my own fork.

### DIFF
--- a/.github/workflows/on-pr-debug.yml
+++ b/.github/workflows/on-pr-debug.yml
@@ -1,6 +1,6 @@
 name: "Debug"
 on:
-  pull_request:
+  pull_request_target:
 
 permissions:
   contents: write # so it can comment


kozlovic (changes requested):
A quick look at this event seems to show that there may be some security risks. So unless you perfectly understand the context and are positive that this does not apply, I would not make this change: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target (look for the Warning section and following link).

levb:
@kozlovic I am still experimenting with this one

kozlovic:
@levb Sorry, I thought that I was asked to review, my mistake.

levb:
I am just going to take codecov out of the forked repo PR flow for now. It appears that the codecov people are working on the [issue](https://github.com/codecov/feedback/issues/301), and there is a [workaround](https://github.com/codecov/feedback/issues/126#issuecomment-1931862637) but it's kludgy. Will close this PR and open another one to remove codecov from fork PRs for now

PR #769, opened 2024-06-24
[DO NOT MERGE] single threaded nats client POC

levb:
PR for discussion only, not to be merged.

<TODO>


kozlovic (commented):
Too much to review. This is obviously very early and more of a POC, but you should still try to run a little benchmark for the publish to see how it will perform. I know that early in the C client development the publish perf was really bad compared to the Go client because each publish ended-up result in the tcp write (because notifying the flusher would do the flush right away). He we copy/move the date to a chain but how often the chain will be flushed/processed will dictate the publish performance.

levb:
@kozlovic ACK that it is way too early to review, and I am focused on other things for the time being anyway. I marked you to get your "FYI"-level attention, feel free to un-assign if this is too noisy.

> ...each publish ended-up result in the tcp write

the plan was to optimize it to use `writev`, like in https://github.com/nginx/nginx/blob/e734df6664e70f118ca3140bcef6d4f1750fa8fa/src/os/unix/ngx_writev_chain.c#L14 but I didn't get to prototype it. So everything would still be buffered (into a chain), and then segments written out of the chain, 1 iovec at a time per write event. I just didn't get to the sockets at all, figured I'd wrap it on top of the existing code for now. Ditto establishing connections - that too would need to be event driven for re-connections.

PR #770, opened 2024-07-07
Disable streaming by default

mtmk:
* Disable streaming by default since it's deprecated and building with it on Windows is not easy


.github/workflows/on-pr-debug.yml
```diff
@@ -77,10 +77,83 @@ jobs:
...
+            
+            local result=$?
+            eval "$current_set"
+            return $result
+          }
+
+          run_ctest
+          result=$?
+      
+          # Retry if failed
```
levb, line 140: I am 0/5 against auto-retries. They mask out the (annoying) flappers, but I'd rather be annoyed and more motivated to fix them, TBH.

.github/workflows/on-pr-debug.yml
```diff
@@ -77,10 +77,83 @@ jobs:
         env:
           VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
         run: |
-          cmake -B build -S . -DCMAKE_C_FLAGS=/W4 -DNATS_BUILD_STREAMING=OFF
+          cmake -B build -S . -DCMAKE_C_FLAGS=/W4
           cmake --build build --config Debug
 
       - name: Test
+        shell: bash
```
levb, line 84: 0/5 as a subsequent improvement (separate PR), maybe consolidate this with build-test.yml, so we have a windows and a unix versions, and keep the on-... workflow files shorter/simpler.

levb (changes requested):
Mostly LGTM, but I wonder if we need the auto-retry?

.github/workflows/on-pr-debug.yml
mtmk: unfortunately there aren't a definitive set of failing tests but there seems to be an issue with 'something' however much I excluded failing tests, one or two always fails! (I'm guessing nats-server process management, just a guess, btw having similar issues with nats.net tests on windows).

if we removed retires tests almost certain to never pass and hide potential real failures which might be useful. having said that, i see your point as well.

if you still think the same please let me know i can remove it.

levb: Since this logic applies to Windows only, I'm good with it for now, but would be great to get to the bottom of this issue. There are 2-3 "known" flappers at the moment, but the rest - 2/5 we should aim to pass, maybe with specific exclusions as applicable.

mtmk: no part of this PR anymore

levb (approved):

Issue #672, opened 2023-07-11
Tests failing `-fsanitize=thread`


levb:
The following tests seem to be failing `-fsanitize=thread` warnings, will disable until further investigation

- [ ] Test_AutoUnsubNoUnsubOnDestroy
- [ ] Test_ClientAutoUnsubAndReconnect
- [ ] Test_EventLoop
- [ ] Test_JetStreamOrderedConsSrvRestart
- [ ] Test_JetStreamOrderedConsWithErrors
- [ ] Test_PendingLimitsDeliveredAndDropped
- [ ] Test_ReconnectWithTokenHandler
- [ ] Test_SrvVersionAtLeast

Full list in test/no-sanitize.txt

PR #771, opened 2024-07-13
[FIXED] cleaning up sanitize=thread found several races

levb:
_note_: I've been working on a bigger PR for async fetch (aka Consume), it is not ready yet but since it will be a big PR I'll try to send in parts that are ready to simplify the review. This is part 1.

- Lock on `msgDlvWorker->lock` where we were not before.
- Use `natsSock_Shutdown` rather than `natsSock_Close` from `natsConnection_Reconnect` to avoid a potential data race with the readloop.
- Cleanup thread sanitizer CI runs and prior test failures. Fixes https://github.com/nats-io/nats.c/issues/672

Before this PR warnings generated by`-fsanitize=thread` in CI were ignored, since the detailed test logs were dropped and not scanned. This PR adds the log scanning.

It also uncovered several issues with the existing tests, all of them but one were relatively trivial. In `Test_ForceReconnect` detected a race caused by us forecefully `close`-ing an `fd` which is still being read on by the readloop. I could not think of a simple solution that would not have performance implications, so I disabled the test for thread sanitizer. The code seems to be doing the right thing anyway, and the readloop thread would error out, as designed.

The following tests were fixed:
- Test_SrvVersionAtLeast
- Test_ForcedReconnect
- Test_ClientAutoUnsubAndReconnect
- Test_AutoUnsubNoUnsubOnDestroy
- Test_EventLoop
- Test_JetStreamOrderedConsSrvRestart
- Test_JetStreamOrderedConsumerWithErrors

This PR also adds a few more CI jobs, including (experimentally for now) DEV_MODE. I will remove the DEV_MODE jobs if they prove flaky and not easily fixable. Same for the (added) debug/sanitize jobs, will monitor and remove/fix if too slow/flaky.


test/test.c
```diff
@@ -29485,6 +29500,9 @@ _jsOrderedErrHandler(natsConnection *nc, natsSubscription *subscription, natsSta
 {
     struct threadArg    *args = (struct threadArg*) closure;
 
+    if (err != NATS_MISSED_HEARTBEAT)
```
levb, line 29503: There were other errors coming in. At the end of the first cycle, there was a `NATS_ERR`, I didn't track it down although it seemed a bit suspiciously generic.

After the end of the test, the callback was getting called (twice?) more with `5` (connection closed), and was locking up the Mutex being destroyed in the main thread already. Ignoring other errors fixed the warning, not sure if the termination is clean.

.github/workflows/build-test.yml
```diff
@@ -158,7 +176,7 @@ jobs:
           export PATH=../deps/nats-server:../deps/nats-streaming-server:$PATH
           export NATS_TEST_SERVER_VERSION="$(nats-server -v)"
           flags=""
-          ctest --timeout 60 --output-on-failure --repeat-until-fail ${{ inputs.repeat }} 2>&1 | tee /tmp/out.txt
+          ctest $VVFLAG --timeout 60 --output-on-failure --repeat-until-fail ${{ inputs.repeat }} 2>&1 | tee /tmp/out.txt
```
kozlovic, line 179: This is strange. At least on travis, the test would fail if there is an error found by `stanitize-(address/test)` without the need to have verbose test output. Why is it different with Github-Actions? I mean the `ctest` should fail (return non zero and therefore make the run fail). Is it that the sanitize output is not present in GA while it is in Travis, and that makes acting on the failure impossible for us dev?

test/test.c
```diff
@@ -20243,6 +20248,15 @@ test_ForcedReconnect(void)
     natsMsg *msg = NULL;
     natsPid pid = NATS_INVALID_PID;
 
+#ifdef __SANITIZE_THREAD__
+    // threadSanitizer complains that we close the fd when a read may be in
+    // progress. Since it appears to work as desired, skipping the test.
```
kozlovic, line 20253: Well this is concerning in that we are testing a public function (natsConnection_Reconnect()). If this is causing a sanitize thread error, then maybe the approach of closing the socket is wrong? I mean it will cause the socket read error for sure, which triggers the reconnect so it works well, but it is technically wrong I guess? Would the sanitize complain (and would the feature still work) if we had use `natsSock_Shutdown()` instead of `natsSock_Close()`?

kozlovic (commented):
All good catches with missing unlocking in tests. I have a question regarding the need of verbose output for sanitize thread and about the reconnect implementation.

.github/workflows/build-test.yml
kozlovic: Ok, I think I understand. I have just compiled on my Ubuntu VM with stanitize-thread and I see that the test passes but if I enable logging (-V) then I see the warning. So it was not considered an error and did not fail the test. So I guess yes, it is worth having verbose test output to catch warnings too.

test/test.c
kozlovic: So I tried with shutdown instead of close and it seems to have solve the warning. Would have to test on all platforms to see if that introduces an issue or not.

.github/workflows/build-test.yml
levb: I found a better solution, using TSAN_OPTIONS, will push to this PR shortly.

test/test.c
levb: I'll make the change once you confirm, probably as a separate PR that fill fix the issue and re-enable the test. It will keep this one test-only, for release notes and such?

.github/workflows/build-test.yml
kozlovic: Good. I would recommend removing the ifdef for the force reconnect test to verify that it works, and then you could replace the socket close with shutdown to verify that it works.

test/test.c
kozlovic: Sure, can be done in separate PR. I did not mean that I was going to check on several platforms myself. I guess the run on GA that runs on Windows should verify that it works. But I think that we don't have Windows yet, so let me try on my Windows VM and verify.

kozlovic: Ok, it works and by the way, in that other PR we can also alter the test a little bit by setting the `natsOptions_SetReconnectWait(opts, 100)` to speed-up the test. Otherwise it waits for the default reconnect wait interval (2sec) since it reconnects to the same server.

levb: Because the ifdef-skip didn't work on Travis as intended, I added the Reconnect changes here (rather than figuring out how to skip properly, since we'd be un-skipping anyway).

test/CMakeLists.txt
```diff
@@ -45,6 +45,10 @@ foreach(name ${listOfTestNames})
   # Make sure the test passes
   set_tests_properties(Test_${name} PROPERTIES PASS_REGULAR_EXPRESSION "ALL PASSED")
 
+  # Set TSAN_OPTIONS for the test
+  set_tests_properties(Test_${name} PROPERTIES
```
kozlovic, line 49: Does this have an effect on any other builds? (where we don't need/want sanitize)

kozlovic (changes requested):
Remove redirect to temp file? By the way, if you would have preferred having different PRs, you could have had one with just the close->shutdown change, that's it, and then once merged, rebase the PR that was just CI/Tests. 
Finally, I had disabled sanitize=thread in Travis due to some bug in Ubuntu/travis combination. We may want to reactivate from time to time to see if that works now.

test/CMakeLists.txt
levb: I don't think so, since TSAN is for Thread SANitizer, and it's an environment variable for running the tests, not a build flag but I put it in an if for `NATS_SANITIZE` only just in case.

levb:
@kozlovic The travis build still does sanitize=thread for clang. I tried to enable gcc, and got the same cmake configuration-time failure, so disabled it back. I added a fuller matrix of sanitize to GH Actions, will see what it produces.

levb:
@kozlovic The fuller matrix found a couple more, in the code, missing libDlvWorker lock; please check.

.github/workflows/on-push-release.yml
```diff
@@ -22,3 +22,26 @@ jobs:
...
+      verbose_make_output: ON
+
+  sanitize-addr:
+    name: "Sanitize address"
+    uses: ./.github/workflows/build-test.yml
+    with:
+      sanitize: address
+      server_version: main
+
+  san-thread:
```
kozlovic, line 42: nit: Not sure why you would shorten to `san-` here while you used `santize-` for the address matrix run above :-)

src/conn.c
```diff
@@ -1137,17 +1137,29 @@ _resendSubscriptions(natsConnection *nc)
 
         adjustedMax = 0;
         natsSub_Lock(sub);
+        if (sub->libDlvWorker != NULL)
```
kozlovic, line 1140: There are macro for lock/unlock of this in `sub.c`. Maybe we move the definition of these in `sub.h` (we include this header in `conn.c`), and so you could replace these 2 lines with:
```
SUB_DLV_WORKER_LOCK(sub);
```

src/conn.c
```diff
@@ -1137,17 +1137,29 @@ _resendSubscriptions(natsConnection *nc)
...
+        {
+            natsMutex_Lock(sub->libDlvWorker->lock);
+        }
         // If JS ordered consumer, trigger a reset. Don't check the error
         // condition here. If there is a failure, it will be retried
         // at the next HB interval.
         if ((sub->jsi != NULL) && (sub->jsi->ordered))
         {
             jsSub_resetOrderedConsumer(sub, sub->jsi->sseq+1);
+            if (sub->libDlvWorker != NULL)
```
kozlovic, line 1150: and:
```
SUB_DLV_WORKER_UNLOCK(sub);
```

src/js.c
```diff
@@ -2045,6 +2045,10 @@ _hbTimerFired(natsTimer *timer, void* closure)
     natsStatus          s    = NATS_OK;
 
     natsSub_Lock(sub);
+    if (sub->libDlvWorker != NULL)
```
kozlovic, line 2048: Same than for `conn.c`. We also include `sub.h` here, so that would work.

src/js.c
```diff
@@ -2075,12 +2087,20 @@ _hbTimerFired(natsTimer *timer, void* closure)
     if (oc)
     {
         natsSub_Lock(sub);
+        if (sub->libDlvWorker != NULL)
+        {
+            natsMutex_Lock(sub->libDlvWorker->lock);
+        }
         if (!sub->closed)
```
kozlovic, line 2094: Humm... there are may places where we check this under the sub's lock only. Sure, it is set under both locks and checked only under the lib's lock in `nats.c` (the code that deals with the delivery's pool thread), but if we find many other reports, then I think it would be simpler to modify `_deliverMsgs` so that we capture `sub->closed` under the sub's lock after releasing the delivery's lock. Anyway, for now change is ok, but again, let's keep that in mind.

kozlovic (changes requested):

.github/workflows/on-push-release.yml
levb: shorter is better? :) (will change for consistency)

src/conn.c
levb: @kozlovic I stayed away from further exposing these macros. I am about to submit a re-factor (just passed all tests now without kicking!) that hopefully eliminates the need for double-locking, leaving the worker lock almost exclusively to protect the message queue. Ditto for the `sub->closed` comment.

It's ok with me if we keep this PR "hanging" for now until you see the upcoming solution, and then we can decide which way to fix? I just want to have it as my base branch to get the CI/unrelated test fixes out of the way.

kozlovic: Not sure I understand the last part, that is, do you want to merge this PR to be the base for the other work or the other work will be the (re)base for the CI changes. Either way, I am going to LGTM this PR so that you proceed as you wish.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #772, opened 2024-07-19
[CHANGED] refactored message delivery for future SubscribePullAsync

levb:
- Consolidated `deliverMsg` logic into a single function (see `dispatch.c`)
- Added `nats_OpenWithConfig` for thread pool control, and in preparation for JSON-ifying default options, a separate request from @jnmoyne 
- [EXPERIMENTAL, not used/tested] Allow using a thread pool for reply delivery.
- Code organization - broke up `nats.c` into ./glib/*.c


src/conn.c
```diff
@@ -1137,29 +1138,17 @@ _resendSubscriptions(natsConnection *nc)
 
         adjustedMax = 0;
         natsSub_Lock(sub);
-        if (sub->libDlvWorker != NULL)
```
levb, line 1140: No longer needed other than for `enqueueMsg`

src/conn.c
```diff
@@ -2773,58 +2735,33 @@ natsConn_processMsg(natsConnection *nc, char *buf, int bufLen)
...
-            sc = !sub->slowConsumer;
-            sub->slowConsumer = true;
-
-            // Undo stats from above.
-            sub->msgList.msgs--;
-            sub->msgList.bytes -= bufLen;
-        }
-        else
+        if (s == NATS_OK)
+            s = natsSub_enqueueMsg(sub, msg);
```
levb, line 2744: `natsSub_enqueueMsg` - main function for posting messages to dispatch queue, see dispatch.c

src/glib/glib.c
```diff
@@ -0,0 +1,491 @@
+// Copyright 2015-2024 The NATS Authors
```
levb, line 1: With the exception of `glib/dispatch_pool.c` these were copied from `nats.c` with minimal changes

src/js.c
```diff
@@ -840,19 +842,10 @@ _timeoutPubAsync(natsTimer *t, void *closure)
         m->sub = js->rsub;
         natsMsg_setTimeout(m);
 
+        // Best attempt, ignore NATS_SLOW_CONSUMER errors which may be returned
```
levb, line 845: I did not test the changes in this path beyond passing the existing tests.

src/js.c
```diff
@@ -1850,23 +1839,22 @@ _fetch(natsMsgList *list, natsSubscription *sub, jsFetchRequest *req, bool simpl
         rply = (const char*) reqSubj;
     if ((s == NATS_OK) && req->Heartbeat)
     {
-        int64_t hbi = req->Heartbeat / 1000000;
-        sub->refs++;
-        if (jsi->hbTimer == NULL)
+        s = nats_createControlMessages(sub);
```
levb, line 1842: best viewed with `-w` here , instead of `jsi->mhMsg` now use `sub->control` which keeps all distinct synthetic messages for the sub.

src/nats.h
```diff
@@ -186,6 +186,28 @@ typedef struct __natsOptions        natsOptions;
...
 
+
+/** \brief An initial configuration for NATS client. Provides control over the
+ * threading model, and sets many default option values.
+ *
+ * @see natsOptions_Create
+ */
+typedef struct __natsClientConfig
+{
+        int64_t DefaultWriteDeadline;
```
levb, line 197: This would be the "role model" for the future:
A `ClientConfig` that we can accept as JSON, used as the default, overridable with env. variables. In this PR I just wanted finer control over the thread pool than the prior ENV variable.

src/natstime.c
```diff
@@ -85,3 +85,12 @@ natsDeadline_GetTimeout(natsDeadline *deadline)
 
     return timeout;
 }
+
+int64_t
+nats_setTargetTime(int64_t timeout)
```
levb, line 90: Moved from nats.c

test/test.c
```diff
@@ -13627,12 +13730,14 @@ test_SlowAsyncSubscriber(void)
     CHECK_SERVER_STARTED(serverPid);
 
     s = natsConnection_Connect(&nc, opts);
-    IFOK(s, natsConnection_Subscribe(&sub, nc, "foo", _recvTestString, (void*) &arg));
+    IFOK(s, natsConnection_Subscribe(&sub, nc, "foo.*", _recvTestString, (void*) &arg));
```
levb, line 13733: not a functional change but was helpful in debigging to make the messages distinct.

src/glib/glib_dispatch_pool.c
```diff
@@ -0,0 +1,181 @@
...
+_startDispatcher(natsDispatcher *d, void (*threadf)(void *))
+{
+    natsStatus s = NATS_OK;
+
+    if (d->running)
+        return NATS_OK;
+
+    s = natsMutex_Create(&d->mu);
+    if (s != NATS_OK)
+        return s;
```
kozlovic, line 40: Is it intentional that we are not using `NATS_UPDATE_ERR_STACK` in this file? (I forgot, but maybe we can't/should not if it is deep down at library level).

src/glib/glib_dispatch_pool.c
```diff
@@ -0,0 +1,181 @@
...
+        // Get the next dispatcher
+        d = &pool->dispatchers[pool->useNext];
+        pool->useNext = (pool->useNext + 1) % pool->cap;
+    }
+    if ((s == NATS_OK) && (d->thread == NULL))
+        s = _startDispatcher(d, threadf);
+
+    // Assign it to the sub.
+    if (s == NATS_OK)
+        sub->dispatcher = d;
```
kozlovic, line 176: Does it need to be done before starting the dispatcher or not?

src/nats.h
```diff
@@ -186,6 +186,28 @@ typedef struct __natsOptions        natsOptions;
...
+
+        int64_t LockSpinCount;
+
+        // Subscription message delivery thread control
+        bool DefaultToThreadPool;
+        int ThreadPoolMax;
+
+        // Reply message delivery thread control
+        bool DefaultRepliesToThreadPool;
+        bool useSeparatePoolForReplies;
```
kozlovic, line 207: Capital `u`?

src/nats.h
```diff
@@ -1773,6 +1795,23 @@ typedef void (*stanConnectionLostHandler)(
  * @{
  */
 
+/** \brief Initializes the library.
+ *
+ * This initializes the library, with more control over how threads and/or
+ * thread pools are used to deliver messages to the app.
```
kozlovic, line 1801: `to the application`?

src/opts.c
```diff
@@ -1044,7 +1045,8 @@ natsOptions_UseGlobalMessageDelivery(natsOptions *opts, bool global)
     // Sets if the subscriptions created from the connection will
     // create their own delivery thread or use the one(s) from
     // the library.
-    opts->libMsgDelivery = global;
+    opts->useSharedDispatcher = global;
+    // opts->useSharedReplyDispatcher = global;
```
kozlovic, line 1049: Remove?

src/sub.c
```diff
@@ -23,54 +23,118 @@
...
-    while ((m = sub->msgList.head) != NULL)
+    _destroyControlMessage(sub->control->sub.timeout);
+    _destroyControlMessage(sub->control->sub.close);
+    _destroyControlMessage(sub->control->sub.drain);
+    _destroyControlMessage(sub->control->batch.expired);
+    _destroyControlMessage(sub->control->batch.missedHeartbeat);
+    NATS_FREE(sub->control);
+}
+
+// Should be called suring the subscription creation process, or under the sub's lock.
```
kozlovic, line 64: typo: `during`

src/sub.c
```diff
@@ -79,73 +143,41 @@ _freeSubscription(natsSubscription *sub)
...
     if (sub == NULL)
         return;
 
-    natsSub_Lock(sub);
+    natsMutex_Lock(sub->mu);
 
     refs = --(sub->refs);
 
-    natsSub_Unlock(sub);
+        natsMutex_Unlock(sub->mu);
```
kozlovic, line 157: extra tab/spaces?

src/sub.c
```diff
@@ -365,74 +266,88 @@ natsSub_close(natsSubscription *sub, bool connectionClosed)
...
+    if (sub->control == NULL)
+        return nats_setDefaultError(NATS_NO_MEMORY);
+    IFOK(s, _createControlMessage(&(sub->control->sub.timeout), sub));
+    IFOK(s, _createControlMessage(&sub->control->sub.close, sub));
+    IFOK(s, _createControlMessage(&sub->control->sub.drain, sub));
+    IFOK(s, _createControlMessage(&sub->control->batch.expired, sub));
+    IFOK(s, _createControlMessage(&sub->control->batch.missedHeartbeat, sub));
+
+    // no need to free on failure, sub's free will clean it up.
+    return s;
```
kozlovic, line 339: Update stack?

kozlovic (changes requested):
That's a big one! Will likely have to have another look, but this is a first pass. I think we need to run some perf test to make sure that we are not degrading performance too much.

src/conn.c
```diff
@@ -2773,58 +2734,33 @@ natsConn_processMsg(natsConnection *nc, char *buf, int bufLen)
...
```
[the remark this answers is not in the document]
levb: I did some cursory checks, but was planning to focus on it weekend, on the final 'PullSubscribeAsync' branch once I finish my struggle with the missing heartbeats. Any other ideas beyond [js-]pub->sub a million messages? I will add the comparison here. Also on this, did you have any tooling for CPU/mem profiling, and is there a built-in thread counter that I missed?

src/dispatch.c
```diff
@@ -0,0 +1,375 @@
...
+
+    return NATS_OK;
+}
+
+static inline void
+_removeHeadMessage(natsDispatchQueue *queue)
+{
+    natsMsg *msg = queue->head;
+
+    queue->head = msg->next;
```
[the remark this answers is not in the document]
levb: Yes

src/glib/glib_dispatch_pool.c
levb: No it is not intentional. TBH I did not yet have a chance to understand what is the idiomatic use of the macro, i.e. when is it appropriate, and when not? Should it be always do it with the exceptions of small or ...Impl style functions? Or would that tend to to over-run the error stack? I'll try to figure it out soon, and meanwhile please do point out any under/mis-use of it. Much appreciated!

src/glib/glib_dispatch_pool.c
levb: I had it first done as create/set/start, but then simplified because all this is done during the sub's creation, so no messages could have yet been enqueued for it in the (shared) queue. I chose to simplify the code, but if I made it obscure, I'd be happy to add the extra step, the perf. impact (if any) would be negligible. 0/5 would prefer a code change over adding a comment.

src/dispatch.c
```diff
@@ -0,0 +1,375 @@
...
+                // Spurious NULL message, where did it come from? (unreachable)
+                nats_unlockDispatcher(d);
+                continue;
+            }
+        }
+
+        userMsg = false;
+        if (msg != NULL)
+        {
+            // At this point sub is set, no need to check for NULL - either we
```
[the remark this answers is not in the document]
levb: Yes, it was in a slightly wrong place, and not very descriptive. Changed. 0/5 maybe best to add an `else sub=d->dedicatedSub` (and remove the default assignment at the top) to better reveal the intention?

src/conn.c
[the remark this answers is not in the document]
levb: image

So there is a slowdown, if minimal. Strangely, I see it more on my machine (M2) with a particularly large difference at 5 threads (500 subs). Increasing the number of messages 10x did not introduce a meaningful difference. Investigating.

src/glib/glib_dispatch_pool.c
kozlovic: Imagine that we have these:
```
natsStatus baz(void) {
    return nats_setError(NATS_ERR, "%s", "Some error");
}
natsStatus bar(void) {
   return baz();
}
natsStatus foo(void) {
   return bar();
}
natsStatus natsConnection_Connect(...
..
s = foo();
if (s != NATS_OK)
   return s;
```
This would produce this stack:
```
Error: 1 - Error - (conn.c:2029): Some error
Stack: (library version: 3.9.0-beta)
  01 - baz
```
But if you had this instead:
```
natsStatus baz(void) {
    return nats_setError(NATS_ERR, "%s", "Some error");
}
natsStatus bar(void) {
    natsStatus s = baz();
}
natsStatus foo(void) {
    natsStatus s = bar();
}
natsStatus natsConnection_Connect(...
..
s = foo();
if (s != NATS_OK)
```
That would produce this stack:
```
Error: 1 - Error - (conn.c:2029): Some error
Stack: (library version: 3.9.0-beta)
  01 - baz
  02 - bar
  03 - foo
  04 - natsConnection_Connect
```
So I would put it almost everywhere in case I want to know how deep the error originate from. There are cases where it is not needed because a function may return an error but the caller will always override the error returned (usually some utility or low level functions) with some more telling error.

src/glib/glib_dispatch_pool.c
kozlovic: As long as we are sure that there is no risk of the sub being used (during a dispatch) before this is set, that is fine.

src/conn.c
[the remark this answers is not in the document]
levb: @kozlovic  With the new benchmark changes in https://github.com/nats-io/nats.c/pull/774, will be posting updates soon, please stand by.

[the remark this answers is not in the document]
levb: @kozlovic after some struggles I had to roll back significantly, going back to the original 2 `deliverMsg` functions.

- My locking changes generally outperformed with a large number of threads/subs, but underperformed with a small number which is the dominant use case.
- The `bool`->`unsigned :1` change was surprisingly (for me) slower -1% to -3% on x86, while barely noticeable on my M2. It also didn't affect the direct injection test much, but was showing a consistent slowness on `Large` and especially `Small` (no flush) tests.

I left the 2 `deliverMsg` functions in their places for now to compare the diffs, will move them to `dispatch.c` in the next PR.

My trickery with changing the PR base to get the numbers compared here did not work, but here is the summary from my fork,

```
== BenchSubscribeAsync_Small ==
Best: -3.88%
Average: -4.58%
Worst: -6.61%
...
== BenchSubscribeAsync_Inject ==
Best: -7.36%
Average: -3.22%
Worst: -0.43%
...
== BenchSubscribeAsync_InjectSlow ==
Best: -0.35%
Average: -0.02%
Worst: 0.15%
...
== BenchSubscribeAsync_Large ==
Best: -0.03%
Average: -0.49%
Worst: 1.90%

```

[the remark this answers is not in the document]
kozlovic: Can you say if negative numbers here mean better or worse? Meaning is it that the new code is `-3.88%` number of messages per second, meaning it is slower than main?

kozlovic:
@levb I have to say that your PR reviews become hard to follow. You make use a lot of merge and separate PRs (which I understand why), but specifically for this PR, I am not sure what to review, and may have to do full review again. 

When branches don't live for too long, I think a rebase makes more sense. Say that you are submitting a PR and during the review process, we agree that something else needs to be done before. Maybe open a PR that has the new change, once committed, update (with a rebase) the original PR so that the new code "sits on top" of the main + the intermediate change. 

I think this one is different because you had to rollback some changes, but still.

I will wait on your comment to know what I have to review, and it is ok if the answer is "everything" :-)

src/conn.c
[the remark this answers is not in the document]
levb: Negative is better; these are "sums of averages of all runs", not directly proportionate to the TPM. I pasted from here: https://github.com/levb/nats.c/actions/runs/10243884983/job/28336496607?pr=102, you can see both run's results and the comparison details (the comparison does not display records with less than 5% difference).

(Unfortunately rebasing this PR did not change the github internal "base" reference from `main`, I'd need to cut a new PR to run the comparison here)

levb:
@kozlovic Yes the entire PR, please. If it makes it easier for you I'd be _very_ happy to replace this with a new (single commit) PR now, it would also have the proper base and run the benchmark properly.

Apologies for so many changes, the performance benchmark/investigation went down too many rabbit holes.

kozlovic:
@levb Just to be clear: you are going to post a new PR instead of this one, correct?

PR #774, opened 2024-07-22
benchmark for SubscribeAsync

levb:


bench/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+  return()
+endif()
+
+include_directories(${PROJECT_SOURCE_DIR}/src)
+include_directories(${PROJECT_SOURCE_DIR}/bench)
+
+if(NATS_BUILD_WITH_TLS)
+  include_directories(${OPENSSL_INCLUDE_DIR})
+endif(NATS_BUILD_WITH_TLS)
+if(NATS_BUILD_LIBUV_EXAMPLE AND LIBUV_DIR)
```
kozlovic, line 13: Do you plan to have bench for libuv/libevent and streaming? If not, you could remove those for now and add back only if needed?

bench/sub-async.c
```diff
@@ -0,0 +1,245 @@
...
+    IFOK(s, nats_SetMessageDeliveryPoolSize(c->maxThreads));
+    IFOK(s, natsOptions_SetErrorHandler(opts, asyncCb, NULL));
+    IFOK(s, natsOptions_UseGlobalMessageDelivery(opts, c->useGlobalDelivery));
+
+    IFOK(s, natsConnection_Connect(&conn, opts));
+
+    for (int i = 0; i < NSUBS; i++)
+    {
+        IFOK(s, natsConnection_Subscribe(&state[i].sub, conn, "foo", _onMessage, &state[i]));
+        IFOK(s, natsSubscription_SetPendingLimits(state[i].sub, INT_MAX, INT_MAX));
```
kozlovic, line 135: I have seen (in other PRs?) builds on Travis complain about `INT_MAX` being unknown. Regardless, unless you changed something that I missed, setting `-1` for both is the way to set them as unlimited.

bench/sub-async.c
```diff
@@ -0,0 +1,245 @@
...
+        IFOK(s, natsSubscription_AutoUnsubscribe(state[i].sub, NMSGS));
+        IFOK(s, natsSubscription_SetOnCompleteCB(state[i].sub, _onComplete, &state[i]));
+    }
+
+    for (int i = 0; i < NMSGS; i++)
+    {
+        snprintf(buf, sizeof(buf), "%d", i);
+        IFOK(s, natsConnection_PublishString(conn, "foo", buf));
+        IFOK(s, natsConnection_Flush(conn));
```
kozlovic, line 144: Is it intentional to flush for every single message? I understand that we are not bench'ing the publisher here, but I wonder if it also could have an impact on consumer bench perf since messages will come more "1 at a time" as opposed to be consumed quickly from the connection's readloop if they were more batched.
You could put the flush outside of the for loop or have it flushed only every n messages.

bench/sub-async.c
```diff
@@ -0,0 +1,245 @@
...
+            s = _bench(c, &start, &end);
+            if (s != NATS_OK)
+            {
+                fprintf(stderr, "Error: %s\n", natsStatus_GetText(s));
+                nats_PrintLastErrorStack(stderr);
+                exit(1);
+            }
+
+            totalDuration += (end - start);
+            // if (s == NATS_OK)
```
kozlovic, line 237: Remove?

kozlovic (changes requested):

bench/CMakeLists.txt
levb: I'll remove what I can without breaking my default dev build settings. (That's the reason I have stan - I normally build/test with it ON locally, and `natsp.h` won't compile otherwise.

1/5 I just cut&pasted from test, but prob. better to consolidate into a test.h and test/bench/... to reuse the makefile and macros. A separate PR though? This chain is already complicated as it is.

bench/bench.h
```diff
@@ -0,0 +1,270 @@
...
+
+    return (natsPid)pid;
+}
+
+#else
+
+typedef pid_t natsPid;
+
+static void
+_stopServer(natsPid pid)
```
levb, line 167: NS: add bench to the coverage build.

bench/sub-async.c
levb: This was intentional. Having it outside the loop was not working well with a larga number of subs, the server would time out after some number of messages. Then again, for a small number of SUBs flushing every time is a significant distortion. 0/5 maybe flush every `NMSGS/NSUBS` cycles?

bench/CMakeLists.txt
kozlovic: What I meant is that since it was a new directory with a separate make file and there was - at the moment - a single file that is neither STAN or libuv/libevent, there was no need for those "if" build statements. But I think I understand: if we have stan/libuv/libevent on, and although the bench directory does not have any related test, some of the includes would cause build issues maybe? It seems that you have removed it in the last commit, so it seems to be ok now?
But agree if you prefer to do in different PR, etc..

bench/sub-async.c
kozlovic: Yes, at least try to do a flush only from time to time, not for every single message.

bench/sub-async.c
kozlovic: Please use `-1` instead of `INT_MAX`.

kozlovic (changes requested):

kozlovic (approved):
LGTM.

levb:
@kozlovic unfortunately, It ended up very noisy, and I am still debugging. More changes coming.

levb:
@kozlovic apologies I pushed here instead of my fork by mistake; still need some touches. WIP, please stand by.

bench/CMakeLists.txt
levb: I merged bench->test to avoid all this duplication.

bench/sub-async.c
levb: This opened a large can of worms. Please see the new commit. This is now stable enough to be comparable with (if not always in favor of) the next PR. I will comment there, but both the "injection" and "publish" methods appear value-able.
 
The benchmark is a little redundant and takes too long at the moment, I'd like to get the measurements in CI for comparison; will tune down later. I am resolving this thread since the code is different now.

.github/workflows/on-pr-debug.yml
```diff
@@ -98,4 +106,4 @@ jobs:
       - name: Test
         run: |
           cd build
-          ./bin/Debug/testsuite
+          ctest -L 'test' --timeout 60 --output-on-failure
```
levb, line 109: cc @mtmk `ctest` seems to work fine?

test/bench_sub_async.c
```diff
@@ -0,0 +1,500 @@
...
+    int subs[] = {1, 2, 3, 5};
+
+    ENV env = {
+        .pubf = _publish,
+        .progressiveFlush = false,
+    };
+    RUN_MATRIX(threads, subs, 500 * 1000, &env);
+}
+
+// This benchmark publishes messages ASAP (no rate limiting) and measures
```
levb, line 89: NS: wrong comment

.github/workflows/on-pr-debug.yml
mtmk: doesn't run any tests
https://github.com/nats-io/nats.c/actions/runs/10116438126/job/27979286329?pr=774#step:5:10
```
Test project D:/a/nats.c/nats.c/build
No tests were found!!!
```

test/bench_sub_async.c
```diff
@@ -0,0 +1,500 @@
...
+            a += dur;
+        }
+    }
+
+    // cleanup
+    for (int i = 0; i < env->numSubs; i++)
+        natsSubscription_Destroy(env->subs[i].sub);
+    natsConnection_Destroy(nc);
+    natsOptions_Destroy(opts);
+    nats_CloseAndWait(0);
```
kozlovic, line 322: You should move this after `_stopServer` since we use mutex and hashtable there.

test/bench_sub_async.c
```diff
@@ -0,0 +1,500 @@
...
+
+            s = natsMsg_Create(&m, subject, NULL, buf, strlen(buf));
+            IFOK(s, _enqueueToSub(env->subs[n].sub, m));
+        }
+    }
+
+    return s;
+}
+
+static natsStatus _enqueueToSub(natsSubscription *sub, natsMsg *m)
```
kozlovic, line 409: I see why we would have this (to test the library without the issue with server delivery), but this may need to be maintained in addition to the rest of the library code.

kozlovic (changes requested):
Also @mtmk mentioned that the `ctest -L` does not seem to be working.

test/bench_sub_async.c
levb: In https://github.com/nats-io/nats.c/pull/772 there's already a `nats_enqueueMsgToSub`, so this will hopefully disappear soon. For now this was cut&pasted from `natsConn_processMsg`, the relevant path.

.github/workflows/on-pr-debug.yml
levb: @mtmk I think I cleaned it up well, borrowing from your PR. Not sure why many tests are flaky, for now retry=3. Please review when you have a chance.

examples/examples.h
```diff
@@ -45,7 +45,7 @@ volatile int64_t elapsed = 0;
 bool             print   = false;
 int64_t          timeout = 10000; // 10 seconds.
 
-natsOptions      *opts   = NULL;
+natsOptions      *gOpts   = NULL;
```
levb, line 48: to avoid warnings about local (idiomatic) `opts` variables.

src/util.c
```diff
@@ -1705,7 +1705,7 @@ nats_Base32_Init(void)
     int         i;
 
     for (i=0; i<(int)sizeof(base32DecodeMap); i++)
-        base32DecodeMap[i] = (char) 0xFF;
+        base32DecodeMap[i] = '\xFF';
```
levb, line 1708: beats me why this was warning on Win32, but it was (int const truncation). Still doing it in `nkeys.c` but I was unable to fix there without digging into the implementation.

kozlovic:
@levb I saw this test `JetStreamInfoAlternates` start to fail a lot in this PR. Seem unrelated and maybe due to server's main vs other releases?

levb:
@kozlovic yes, I think it was caused by https://github.com/nats-io/nats-server/pull/5506, but haven't had a chance to understand/adjust the test yet. If you know how - would MUCH appreciate it!

levb:
`402  Payment Required` - that's a nice wrinkle (Travis)

kozlovic (approved):
LGTM (looking at the flap test to see if I understand what's going on. If I do, will post PR if this is an issue with the test itself).

kozlovic:
@levb Are you going to merge this to main or is it to be merged only onto the other PR? I want to submit a PR for a fix to the test, but wanted to base it on that one so that there is less chance of a conflict?

PR #778, opened 2024-08-06
[CHANGED] refactored nats.c, prep for js_PullSubscribeAsync

levb:
This is a replacement for https://github.com/nats-io/nats.c/pull/772 that had too many changes/re-bases and had grown too complicated for review.

- Code organization - broke up nats.c into ./glib/*.c
- Added nats_OpenWithConfig for thread pool control, and in preparation for JSON-ifying default options, a separate request from @jnmoyne
- [EXPERIMENTAL, not used/tested] Allow using a thread pool for reply delivery.


levb:
performance benchmark: https://github.com/nats-io/nats.c/actions/runs/10267478369/job/28408116810?pr=778

Step: compare benchmarks
image

.github/workflows/on-push-release.yml
```diff
@@ -22,26 +22,3 @@ jobs:
       server_version: main
       ubuntu_version: ${{ matrix.ubuntu_version }}
       compiler: ${{ matrix.compiler }}
-
```
levb, line 25: These are largely redundant with the PR workflow, no real need to have them on push.

src/sub.c
```diff
@@ -23,54 +23,99 @@
...
+// Sets up the default (user-thread NextMsg-only) dispatcher for a sub. Avoid
+// the use of term "pull" since it's not to be confused with JetStream consumer
+// pull requests.
+static natsStatus
+_initOwnDispatcher(natsSubscription *sub)
+{
+    natsStatus s = NATS_OK;
+
+    if (sub->ownDispatcher.dedicatedTo != NULL)
+        return NATS_ILLEGAL_STATE; // already started
```
kozlovic, line 84: `return nats_setDefaultError(NATS_ILLEGAL_STATE);`

src/sub.c
```diff
@@ -23,54 +23,99 @@
...
 
-    if (sub->deliverMsgsThread != NULL)
+    sub->ownDispatcher.dedicatedTo = sub;
+    sub->ownDispatcher.mu = sub->mu;
+    s = natsCondition_Create(&sub->ownDispatcher.cond);
+    return s;
```
kozlovic, line 89: `return NATS_UPDATE_ERR_STACK(s);`

src/sub.c
```diff
@@ -23,54 +23,99 @@
...
+}
+
+static inline void _cleanupOwnDispatcher(natsSubscription *sub)
+{
+    nats_destroyQueuedMessages(&sub->ownDispatcher.queue);
```
kozlovic, line 94: Again, cleanup/destroy functions may be called as part of the creation of the object and you have to assume that the failure could happen at any point and therefore some fields may not have been initialized. So here ownDispatcher seem to be a struct, so it won't be `NULL`, but `queue` may, yet `nats_destroyQueuedMessages` does not seem to be checking for `NULL` there.

src/sub.c
```diff
@@ -178,37 +188,36 @@ _setDrainCompleteState(natsSubscription *sub)
...
     _setDrainCompleteState(sub);
     natsSub_Unlock(sub);
 }
 
 // _deliverMsgs is used to deliver messages to asynchronous subscribers.
-void
-natsSub_deliverMsgs(void *arg)
+void natsSub_deliverMsgs(void *arg)
 {
-    natsSubscription    *sub        = (natsSubscription*) arg;
```
kozlovic, line 197: Call me crazy, but I preferred the way I was doing before (fields aligned) :-) (but that's just personal preference).
Actually, I used even to align the `= ...` part of it, which had to be changed sometimes when adding a new field, etc..

src/sub.c
```diff
@@ -301,7 +310,7 @@ natsSub_deliverMsgs(void *arg)
     }
 
     natsSub_Lock(sub);
-    onCompleteCB        = sub->onCompleteCB;
```
kozlovic, line 304: That is an example of what I was talking about :-)

src/sub.c
```diff
@@ -315,17 +324,27 @@ natsSub_deliverMsgs(void *arg)
...
 
-bool
-natsSub_setMax(natsSubscription *sub, uint64_t max)
+// Should be called only during the subscription creation process, no need to lock
+static inline natsStatus
+_runOwnDispatcher(natsSubscription *sub, bool forReplies)
+{
+    natsStatus s = NATS_OK;
+    if (sub->ownDispatcher.thread != NULL)
+        return NATS_ILLEGAL_STATE; // already running
```
kozlovic, line 333: setDefaultError.

src/sub.c
```diff
@@ -315,17 +324,27 @@ natsSub_deliverMsgs(void *arg)
...
+
+    sub->dispatcher = &sub->ownDispatcher;
+    s = natsThread_Create(&sub->ownDispatcher.thread, natsSub_deliverMsgs, (void *) sub);
+    return s;
```
kozlovic, line 337: update stack

kozlovic (changes requested):
Some changes.

Now, I need to look back at some comment I made in an earlier PR regarding the delivery lock situation, that we could modify something and reduce a lot the use of it. It was a moot point since you started refactoring that (and got rid of altogether), but now that we are back at having the two, I wonder if this would still be applicable.

src/glib/glib_dispatch_pool.c
```diff
@@ -0,0 +1,180 @@
...
+_startDispatcher(natsDispatcher *d, void (*threadf)(void *))
+{
+    natsStatus s = NATS_OK;
+
+    if (d->running)
+        return NATS_OK;
+
+    s = natsMutex_Create(&d->mu);
+    if (s != NATS_OK)
+        return s;
```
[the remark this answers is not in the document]
levb: Should be the same since __func__ will still be defined.

src/glib/glib.c
```diff
@@ -0,0 +1,491 @@
...
+    {
+        s = natsThread_Create(&(gLib.gc.thread), nats_garbageCollectorThreadf, &gLib);
+        if (s == NATS_OK)
+            gLib.refs++;
+    }
+    if (s == NATS_OK)
+        s = natsNUID_init();
+
+    if (s == NATS_OK)
+        s = nats_initDispatcherPool(&(gLib.messageDispatchers), config->ThreadPoolMax);
```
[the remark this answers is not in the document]
levb: It just initializes the struct (mutex), and if `config->ThreadPoolMax` is 0 makes no other allocations. No threads are started until needed. It sort of fell out of using a pre-allocated struct for messageDispatchers in gLib; since it's always there and no easy way to check `== NULL` if it has been initialized, best to have it pre-initialized?

src/glib/glib.c
```diff
@@ -0,0 +1,491 @@
...
+        natsCondition_Broadcast(gLib.cond);
+    }
+
+    gLib.initializing = false;
+    gLib.wasOpenedOnce = true;
+
+    natsMutex_Unlock(gLib.lock);
+
+    if (s != NATS_OK)
+        _libTearDown();
```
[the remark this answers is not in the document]
levb: `gLib.messageDispatchers` is calloced as part of glib, is still NULL-ed out (and the cap is 0), it shouldn't crash. Even if `cap` is > 0, `->thread` is always NULL. Dirrot for replyDispatchers.

src/glib/glib_dispatch_pool.c
```diff
@@ -0,0 +1,180 @@
...
+        if (cond != NULL)
+            natsCondition_Signal(cond);
+
+        nats_unlockDispatcher(&pool->dispatchers[i]);
+    }
+}
+
+void nats_waitForDispatcherPoolShutdown(natsDispatcherPool *pool)
+{
+    for (int i = 0; i < pool->cap; i++)
```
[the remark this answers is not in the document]
levb: They are pre-allocated in gLib (not pointers) and NULL-ed out so this would be a single 'if'.

src/glib/glib_dispatch_pool.c
```diff
@@ -0,0 +1,180 @@
...
+    if (s != NATS_OK)
+        nats_freeDispatcherPool(pool);
+    return s;
+}
+
+void nats_signalDispatcherPoolToShutdown(natsDispatcherPool *pool)
+{
+    natsCondition *cond = NULL;
+
+    for (int i = 0; i < pool->cap; i++)
```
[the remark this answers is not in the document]
levb: Yes, both pool structs are pre-allocated in gLib (not pointers), but are NULLed out initially. The lock is to access (and resize?) the threads array when adding threads. Not sure why it remained rather than gLib.mu, both are rare; anyway the overhead should be minimal (1-2 mutexes created on startup)

src/sub.c
levb: Same, `queue` is not a pointer; a NULL-ed out struct, so the for loop bails.

src/glib/glib_dispatch_pool.c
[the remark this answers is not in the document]
kozlovic: Ok, got it.

src/glib/glib_dispatch_pool.c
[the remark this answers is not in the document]
kozlovic: Ok I see.

src/glib/glib.c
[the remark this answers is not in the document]
kozlovic: Ok.

src/sub.c
levb: I am not sure I am following :) I _think_ this one is VERY low overhead since it's done once, so kind of does not matter that we re-lock here?

src/sub.c
kozlovic: No problem.I checked my VSCode and don't see anything special that I have set that would override this behavior, so I would assume that you have something that makes that happen? I have something for the Editor that trims trailing spaces when saving a file.

src/sub.c
kozlovic: Oh no, I was just saying that I was aligning the `= ...` with the line below, that was a comment about "tabulation", that's all.

kozlovic (approved):
LGTM

levb:
@kozlovic Can I please get an LGTM on https://github.com/nats-io/nats.c/pull/777 (the base) so I can merge cleanly?

src/sub.c
levb: @kozlovic do you mind emailing me a copy of your VSCode `settings.json` so I can sync the auto-formatting?

kozlovic (approved):
LGTM (I already done so yesterday - see comment above your request)

PR #780, opened 2024-08-06
[ADDED] TLS: natsOptions_TLSHandshakeFirst()

### LINKED ISSUE #779: Support handshake_first
### Proposed change

The NATS Server supports a `handshake_first` option which configures TLS connections to begin the TLS negotiation immediately, rather than first issuing an INFO response to negotiate up to TLS.

The C client doesn't appear to support this option yet.

### Use case

If the server is configured with `handshake_first: true`, the server will expect TLS connections to start with the TLS negotiation, but the C client code does not have an option to start TLS connections that way.

### Contribution

_No response_

kozlovic:
This is to force a client to perform the TLS handshake first, that is, not wait for the INFO protocol from the server. This is needed if the server is configured to require clients to perform the TLS handshake first (before sending the INFO protocol).

Resolves #779


johnweldon (approved):
LGTM

Thank you!

kozlovic:
@levb Looking at a flapper on Windows (not specific to this PR it seems). If I find a fix (assuming it is just test), will add a commit here. At any rate, will ping when ready for review.

kozlovic:
@levb Ok, I added a fix for the test. Ready for review.

levb (approved):
LGTM

levb:
@kozlovic It's failing the test against an older server, was this a 2.10 feature? 
image

kozlovic:
Maybe. Let me see. If so, we just need to add the check for server version. Give me a minute.

kozlovic:
Yes, it is a 2.10.0 thing. I will post a quick PR to fix the test.

PR #783, opened 2024-08-07
Added .clang-format to standardize code formatting

levb:
@kozlovic as I now understand, we may have different formatters/versions locally configured; this proposal is to standardize on `clang-format`, and includes the proposed config file. I ran it on `sub.c` so we can discuss the diffs. I tried to minimize them, failed to eliminate entirely. I found the remaining diffs acceptable.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,224 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: Left
+AlignConsecutiveAssignments:
+  Enabled:         true
+AlignConsecutiveDeclarations:
+  Enabled:         true
+AlignConsecutiveMacros:
+  Enabled:         true
+AlignConsecutiveShortCaseStatements:
+  Enabled:         true
+AlignEscapedNewlines: Right
+AlignOperands:   Align
+AlignTrailingComments:
+  Kind:            Always
+  OverEmptyLines:  1
+AlignConsecutiveTableGenCondOperatorColons: 
+  Enabled:         true
+AlignConsecutiveTableGenDefinitionColons: 
+  Enabled:         true
+AllowAllArgumentsOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowBreakBeforeNoexceptSpecifier: Never
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortCompoundRequirementOnASingleLine: true
+AllowShortEnumsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLambdasOnASingleLine: All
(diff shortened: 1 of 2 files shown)


kozlovic (commented):
There are things a bit weird, something like:
```
int    a = 0;
char  *b = NULL;
```
That is, the fact that the `*` is one character before. So when adding new code, you would need a `tab` then backspace, which means really adding 3 spaces, etc..

Question: Is that file just for GitHub? Meaning that we add code as we want and it will reformat as part of the PR? Or is that something that is used by VSCode?

Finally, this is going to create a massive change, which is ok since it is just formatting, but be prepared to have `git blame` having to skip this PR when looking for the origin of some code ;-)

levb:
> Question: Is that file just for GitHub? Meaning that we add code as we want and it will reformat as part of the PR? Or is that something that is used by VSCode?

This is for VSCode/format-on-save/format-on-paste or format document. It is my understanding (from the last couple hours, no prior knowledge) that the VSCode C/CPP extension uses clang-format by default, and I have not yet fully understood the fallback path - there are bundled formatter(s), fallback configuration, etc. Complicated. I did `clang-format -style="LLVM" -dump-config > .clang-format`, tuned it a bit to my satisfaction. Since it's the default formatter, I _think_ it should work everywhere.

> There are things a bit weird, something like:
> 
> ```
> int    a = 0;
> char  *b = NULL;
> ```
> 
> That is, the fact that the `*` is one character before. So when adding new code, you would need a `tab` then backspace, which means really adding 3 spaces, etc..

I did not find the way in `clang-format` to preserve the status quo. It can left-align the `*`, so it'd be like
```
unsigned int i;
char *     c;
```
Personally, 1/5 I like the offset, makes pointers stand out.  I _think_ VSCode should also auto-format as you type; if not hitting Cmd+S triggers format-on-save, so it's not a problem for me. FWIW, I am not sure how to make `clang-format` align to the next indent boundary rather than the longest line.

> Finally, this is going to create a massive change, which is ok since it is just formatting, but be prepared to have `git blame` having to skip this PR when looking for the origin of some code ;-)

Right, thanks for pointing it out, I missed that. I'll remove sub.c from this PR, it's only here as an illustration; Then, if we were to go through with the big change, what do you think of https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/#fun-fact%3A-github-picks-up-%60.git-blame-ignore-revs%60?

kozlovic (changes requested):
Ok, it works with adding the `"[c]"` formatOnSave in VSCode.

If we were to go there, I guess we may want to add/modify some of the one I suggest. BUT, the nats.h looks ugly to me (weird comments before definition of a `#define`, etc...), so that begs the question: would it not be just easier for YOU to remove the "formatOnSave" and not have a .clang-format file? After all, this all started because you modifying files would remove the indentation that we had.

(Edited - removed a "But..." at the end of this message)

levb:
> begs the question: would it not be just easier for YOU

👍 yes. 

Having said that, I would like to get us to the point when auto-formatting the code is at the "go" level - standardized and automatic. This was an unsuccessful attempt, some day I'll make a better one. I did learn that a lot of my problems came from "format-on-paste" which I now disabled.

PR #784, opened 2024-08-07
Fixed flappers and test_SSLHandshakeFirst by skipping if server is < 2.10.0

kozlovic:

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -7847,7 +7847,7 @@ void test_AssignSubToDispatch(void)
         {
             testf("%d subs over %d threads: Verify that the dispatchers have been assigned: ", tc->numSubs, tc->expectedDispatchers);
             for (i = 0; (s == NATS_OK) && (i < tc->numSubs); i++)
-            {   
+            {
                 natsSub_Lock(subs[i]);
                 if (subs[i]->dispatcher != &pool->dispatchers[i % tc->expectedDispatchers])
                     s = NATS_ERR;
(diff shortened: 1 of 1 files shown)


test/test.c
```diff
@@ -18352,7 +18352,7 @@ void test_GetClientID(void)
         testCond(true);
         return;
     }
-    pid1 = _startServer("nats://127.0.0.1:4222", "-cluster nats://127.0.0.1:6222 -cluster_name abc", true);
+    pid1 = _startServer("nats://127.0.0.1:4222", "-a 127.0.0.1 -p 4222 -cluster nats://127.0.0.1:6222 -cluster_name abc", true);
```
levb, line 18355: @kozlovic is this (extra server flags) a windows-specific fix?

levb (approved):
LGTM

test/test.c
kozlovic: Not really. When we just specify the port, the server uses `0.0.0.0` for the TCP listen address, which then can resolve to multiple IPs that are sent to the client through gossip protocol. What I have seen (in my Windows VM, and probably what is happening on Windows CI) is that the server would return multiple addresses (say 192.168.x.y but also 10.5.x.y, etc..), which "interfere" with the tests expectations. On Linux or other CIs that we are running it on, there was probably nothing returned (maybe it is 127.0.0.1).
So the "fix" is not really Windows specific. It is simply making the test more deterministic.

Issue #786, opened 2024-08-09
Add SNI extension when connecting with TLS


thierryba:
### Proposed change

It would be nice (and consistent with nats-py and the nats-server to set the SNI extension when connecting via TLS. That allows for load balancing to work properly in the cloud.

### Use case

We hae nats instances for each of our customers. We are on the cloud. Today we have to use 1 load balancer per client. With the Fix to issue 780 and the addition of handshake_first, the last issue is that the Client Hello is missing the SNI extension. That can be done with 1 simple call to OpenSSL

### Contribution

definitely.

PR #787, opened 2024-08-09
Adding SNI extension if available

thierryba:
This adds the SNI extension and completes the ability to use nits-server per instance in the cloud with automatic load balancing.

This fixes https://github.com/nats-io/nats.c/pull/786


src/conn.c
```diff
@@ -737,6 +737,13 @@ _makeTLSConn(natsConnection *nc)
                 SSL_set_verify(ssl, SSL_VERIFY_PEER, _collectSSLErr);
         }
     }
+#if defined(NATS_USE_OPENSSL_1_1)
+    // add the host name in the SNI extension
+    if ((s == NATS_OK) && (nc->cur != NULL) && (!SSL_set_tlsext_host_name(ssl, nc->cur->url->host)))
```
kozlovic, line 742: This makes test `SSLVerifyHostname` fail if the library/tests are build with `NATS_FORCE_HOST_VERIFICATION=OFF`. So I would say that at the very least we need to add the condition of `NATS_FORCE_HOST_VERIFICATION` enabled as condition to call this function.

But I would also like to see a test that demonstrates why this is necessary for the test to succeed (that is, without this new line it would fail). The presence of it seem to not fail the existing tests (well, except when name verification is not forced).

kozlovic (changes requested):

src/conn.c
kozlovic: I take back that this change makes the test fails. It fails without your changes as long as the force host verification is disabled it seems. I am investigating.

kozlovic: Ok, but still we would need a test that shows that it works ;-)

thierryba: TBH I have no idea how to test this ;)

thierryba: there is nothing that tests the SSL config at this point. It is like a blackbox. I can tell you that it works because I tried it out on a specific setup but that's it.

kozlovic: I am assuming that you know it works because you tried in your env but don't know how to write a test about it? I mean, is it a matter of having the server's TLS configure a certain way? Any information you could give.

thierryba:
well it is not directly the server but the aws load balancer that uses that value to route traffic.

Envoyé à partir de Outlook pour Android<https://aka.ms/AAb9ysg>
________________________________

thierryba:
@kozlovic if we'd have a simple tcp server that could listen to connection we could probably configure the client to connect to it with tls_handshake_first... and we could check that the client hello does contain the SNI extension (which is in clear). Would that make sense?

test/test.c
```diff
@@ -21347,6 +21347,85 @@ void test_SSLHandshakeFirst(void)
...
+    IFOK(s, natsOptions_TLSHandshakeFirst(arg.opts));
+    IFOK(s, natsOptions_SetServers(arg.opts, &server, 1));
+    if (s != NATS_OK)
+        FAIL("@@ Unable to setup test!");
+
+    test("Check that when initiating a TLS connection the SNI extension is set: ")
+
+    arg.control = 3;
+
+    _startMockupServer(&sock, "localhost", "4222");
```
kozlovic, line 21374: `s = _start..`

test/test.c
```diff
@@ -21347,6 +21347,85 @@ void test_SSLHandshakeFirst(void)
...
+
+    s = _createDefaultThreadArgsForCbTests(&arg);
+    IFOK(s, natsOptions_Create(&(arg.opts)));
+    IFOK(s, natsOptions_SetSecure(arg.opts, true));
+    IFOK(s, natsOptions_TLSHandshakeFirst(arg.opts));
+    IFOK(s, natsOptions_SetServers(arg.opts, &server, 1));
+    if (s != NATS_OK)
+        FAIL("@@ Unable to setup test!");
+
+    test("Check that when initiating a TLS connection the SNI extension is set: ")
```
kozlovic, line 21370: Let's change that to `test("Start server and connect client: ");`

test/test.c
```diff
@@ -21347,6 +21347,85 @@ void test_SSLHandshakeFirst(void)
...
+    IFOK(s, natsThread_Create(&t, _connectToMockupServer, (void*) &arg));
+
+    if ((s == NATS_OK)
+        && (((ctx.fd = accept(sock, NULL, NULL)) == NATS_SOCK_INVALID)
+            || natsSock_SetCommonTcpOptions(ctx.fd) != NATS_OK))
+    {
+        s = NATS_SYS_ERROR;
+    }
+    
+    testCond(s == NATS_OK);
```
kozlovic, line 21386: Need a single `testCond()` per `test()`, so do `testCond((s == NATS_OK) && (ctx.fd > 0));`

test/test.c
```diff
@@ -21347,6 +21347,85 @@ void test_SSLHandshakeFirst(void)
...
+    testCond(ctx.fd > 0);
+
+    char buffer[1024];
+    memset(buffer, 0, sizeof(buffer));
+
+    int size = recv(ctx.fd, buffer, sizeof(buffer), 0);
```
kozlovic, line 21392: Before this:
```
test("Read ClientHello from client: ");
```

test/test.c
```diff
@@ -21347,6 +21347,85 @@ void test_SSLHandshakeFirst(void)
...
+    if (size <= 0)
```
kozlovic, line 21393: Replace with:
```
testCond(size > 0);
```

test/test.c
```diff
@@ -21347,6 +21347,85 @@ void test_SSLHandshakeFirst(void)
...
+
+
+    // remove all null chars to allow the use of strstr on the result
+    for (int i = 0; i < size; ++i) {
+        if (buffer[i] == 0)
+            buffer[i] = '0';
+    }
+
+    bool found = strstr(buffer, "localhost");
+#if defined(NATS_USE_OPENSSL_1_1)
```
kozlovic, line 21404: Before this line:
```
test("Check hostname is found in ClientHello: ");
```

kozlovic (changes requested):
Yes, that is the type of test that would verify that the hostname is actually set. Just made some requests to tweak few things.

kozlovic (approved):
LGTM. Thanks for adding this test!

levb:
Thanks, @thierryba!

PR #788, opened 2024-08-09
Fixed test SSLVerifyHostName when build NATS_FORCE_HOST_VERIFICATION=OFF

kozlovic:
This test would fail unless we force host verification, so adapt test to take into consideration the expected result based on the build environment variable.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -13,15 +13,18 @@ on:
       dev_mode:
         type: string
         default: "OFF"
-      lib_msg_delivery:
+      pool_dispatch:
         type: string
-        default: "OFF"
-      lib_write_deadline:
+        default: "NO-pool"
+      write_deadline:
         type: string
-        default: "OFF"
-      lib_force_host_verification:
+        default: "NO-write_deadline"
+      tls:
         type: string
-        default: "ON"
+        default: "TLS"
+      verify_host:
+        type: string
+        default: "verify_host"
       repeat:
         type: string
         default: "1"
@@ -33,9 +36,6 @@ on:
         description: "nats-server version to test with"
         default: "latest"
       streaming:
-        type: string
-        default: "ON"
-      tls:
         type: string
         default: "ON"
       type:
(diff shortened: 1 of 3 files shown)


kozlovic:
@levb We already have quite a bit of a matrix (even within GA I mean), but do you think we need to have a new one to test with the force_host_verification disabled? It is enabled by default and I would want the lib to be like that by default, but yeah, building with it disabled showed that I did not account for that originally in that test.

levb (approved):
LGTM;

I added a GHA job and pushed here, hope you don't mind.

levb:
@kozlovic I am now looking at the https://github.com/nats-io/nats.c/actions/runs/10322972901/job/28579373546?pr=788 failure, will send a separate PR for it, seems another test (args race) issue

.github/workflows/build-test.yml
```diff
@@ -19,6 +19,9 @@ on:
       lib_write_deadline:
         type: string
         default: "OFF"
+      lib_force_host_verification:
+        type: string
+        default: "OFF"
```
kozlovic, line 24: Again, I think default should be "ON". This is how the CMakeFile is, and how we want it to be. What we may want is a matrix run where we disable it, and then on that matrix run, we could even run only test that start with `SSL`...

levb: 0/5 leave it as a separate job for now (I reversed the default/new job settings).

The last time I tried to use the matrix fully I ran into a couple issues, the biggest being that the jobs (names) were difficult to distinguish in the workflow view; I'll try again soon.

.github/workflows/build-test.yml
```diff
@@ -19,6 +19,9 @@ on:
       lib_write_deadline:
         type: string
         default: "OFF"
+      lib_force_host_verification:
+        type: string
+        default: "ON"
```
kozlovic, line 24: Ok, I think I may have misunderstood how that works. If one clone nats.c and do a `cd buil & cmake ..`, the library will be built with "force host verification" enabled, and so I think that the GAs should be doing that.
However, we could *add* a matrix run (like in Travis) to build with that option set to "OFF" and maybe just run the SSL tests.

levb: @kozlovic now I am confused. I thought that's what I just did: see https://github.com/nats-io/nats.c/pull/788/files#diff-d415e6f64dcbe0bb765bd0e64194d42c8b00c5fb1df3da5ff52e1893571a5670R77; and no matrix yet, but the new job (in addition to the defaults) should cover. I added it with sanitize just because (sanitize) should be harmless to the functionality of the test, right? I'll try to compile a reasonable matrix (and express it as such soon).

kozlovic: If this is the definition of the CMakeLists.txt options, then yes, it should be "ON", but my assumption was that anything that we don't have already as a default in CMakeLists.txt would not need to be set, but maybe I am wrong.

So what I meant is that by default the option is and should be ON, so "default" tests should run with a library built with the force host verification on.

What I see is on-push-release-extra.yml that seem to have what is the same than the "matrix" runs in Travis with "TLS-OFF", "no-streaming", etc... Maybe it is simply here that we should have it disabled? I am really not sure anymore how those things work.

In Travis, we could have something like:
```
    - name: "gcc-9 - TLS hostname verification disabled"
      compiler: gcc
      addons:
        apt:
          sources:
            - ubuntu-toolchain-r-test
            - sourceline: ppa:ubuntu-toolchain-r/test
          packages:
            - g++-9
      env:
        - NATS_TEST_SERVER_VERSION=main
        - MATRIX_EVAL="CC=gcc-9"
        - BUILD_OPT="-DNATS_BUILD_TLS_FORCE_HOST_VERIFY=OFF -DNATS_BUILD_ARCH=64 -DCMAKE_BUILD_TYPE=Release" DO_COVERAGE="no" CTEST_OPT="-R SSL"
```
(and same for clang - note, I have not tested that, but you get the gist).

Note the cmake build option `-DNATS_BUILD_TLS_FORCE_HOST_VERIFY=OFF` and `CTEST_OPT="-R SSL"` that would make only the SSL tests to run.

levb: Ok, I'll try to simplify and matrix-ify this weekend.

`“One only understands the things that one tames,” said the fox.` I'll go from the bottom up.

PR #789, opened 2024-08-12
Fixing setting handshake_first without setting secure in the natsOptions

thierryba:
Using the new TLS handshake first without calling natsOptions_SetSecure(opts, true) is crashing because it does not create a sslCtx.


src/opts.c
```diff
@@ -372,7 +372,7 @@ natsOptions_TLSHandshakeFirst(natsOptions *opts)
     LOCK_AND_CHECK_OPTIONS(opts, 0);
 
     opts->tlsHandshakeFirst = true;
-    opts->secure            = true;
+    natsOptions_SetSecure(opts, true);
```
levb, line 375: @thierryba Thanks for the fix! `natsOptions_SetSecure` would be attempting to double-lock `opts`. I suggest having a "private" `natsOptions_setSecure` (lowercase) that does not lock, and then both `natsOptions_SetSecure` and `natsOptions_TLSHandshakeFirst` would wrap that, under their respective locks.

thierryba:
@levb sorry for force pushing... but I did implement it with a private function.

levb (approved):
LGTM.

Thx! Sorry I deleted my first comment by mistake - I got confused for a moment that the tests passed, didn't realize we had RECURSIVE set on the lock.

levb (approved):
LGTM

src/opts.c
```diff
@@ -367,12 +367,10 @@ natsOptions_SetSecure(natsOptions *opts, bool secure)
 natsStatus
 natsOptions_TLSHandshakeFirst(natsOptions *opts)
 {
-    natsStatus s = NATS_OK;
-
     LOCK_AND_CHECK_OPTIONS(opts, 0);
 
     opts->tlsHandshakeFirst = true;
-    opts->secure            = true;
+    natsStatus s = natsOptions_SetSecure(opts, true);
```
kozlovic, line 373: I know that we started to define variables anywhere in the code now, and I guess I should be ok with that as long as compilers (with our given settings) don't complain, but if you don't mind, leave the `natsStatus s;` defined above and here have:
```
s = natsOptions_SetSecure(opts, true);
if (s == NATS_OK)
    opts-tlsHandshakeFirst = true;
```

test/test.c
```diff
@@ -21324,6 +21324,22 @@ void test_SSLHandshakeFirst(void)
     s = natsOptions_TLSHandshakeFirst(opts);
     testCond(s == NATS_OK);
 
+    test("Set TLSHandshakeFirst option without setting secure: ");
+    {
+        // we start with a new natsOptions struct so that we can test
+        // that it does not crash with a minimal config
+        natsOptions *opts = NULL;
```
kozlovic, line 21331: Could we rename this options variable (to avoid shadow)?

test/test.c
```diff
@@ -21324,6 +21324,22 @@ void test_SSLHandshakeFirst(void)
...
+        s = natsOptions_Create(&opts);
+        IFOK(s, natsOptions_SetURL(opts, "nats://127.0.0.1:4443"));
+        IFOK(s, natsOptions_SetTimeout(opts, 500));
+        s = natsOptions_TLSHandshakeFirst(opts);
```
kozlovic, line 21335: `IFOK(s, `

test/test.c
```diff
@@ -21324,6 +21324,22 @@ void test_SSLHandshakeFirst(void)
...
+        IFOK(s, natsConnection_Connect(&nc, opts));
+        testCond(s == NATS_SSL_ERROR);
+        natsConnection_Destroy(nc);
```
kozlovic, line 21338: No need since it was not destroyed.

test/test.c
```diff
@@ -21324,6 +21324,22 @@ void test_SSLHandshakeFirst(void)
...
+        natsOptions_Destroy(opts);
```
kozlovic, line 21339: Would have to be with the renamed variable.

test/test.c
```diff
@@ -21324,6 +21324,22 @@ void test_SSLHandshakeFirst(void)
...
+    {
+        // we start with a new natsOptions struct so that we can test
+        // that it does not crash with a minimal config
+        natsOptions *opts = NULL;
+        s = natsOptions_Create(&opts);
+        IFOK(s, natsOptions_SetURL(opts, "nats://127.0.0.1:4443"));
+        IFOK(s, natsOptions_SetTimeout(opts, 500));
+        s = natsOptions_TLSHandshakeFirst(opts);
+        IFOK(s, natsConnection_Connect(&nc, opts));
+        testCond(s == NATS_SSL_ERROR);
```
kozlovic, line 21337: So to be clear, we are expecting an error because some other SSL needed options were not set (like skip verify hostname).

kozlovic (changes requested):
Good catch! However, I would like you to do some modifications if you don't mind. Thanks!

test/test.c
thierryba: sure

test/test.c
thierryba: yes adding a comment.

src/opts.c
```diff
@@ -372,7 +372,7 @@ natsOptions_TLSHandshakeFirst(natsOptions *opts)
     LOCK_AND_CHECK_OPTIONS(opts, 0);
 
     opts->tlsHandshakeFirst = true;
-    opts->secure            = true;
+    s = natsOptions_SetSecure(opts, true);
```
kozlovic, line 375: I would have set "opts->tlsHandshakeFirst = true" only on success of "SetSecure".

kozlovic (approved):
LGTM!

PR #790, opened 2024-08-12
[CI only] Streamlined/fixed sanitize/coverage matrices

levb:
Streamlined and fixed the test matrix for coverage, sanitize, and environment variable/commpile flag configurations.

### DIFF
--- a/.github/workflows/on-pr-debug.yml
+++ b/.github/workflows/on-pr-debug.yml
@@ -18,27 +18,6 @@ jobs:
       server_version: main
       type: Debug
 
-  coverage:
-    name: "Coverage"
-    uses: ./.github/workflows/build-test.yml
-    with:
-      coverage: ON
-      server_version: main
-      type: Debug
-    secrets:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
-  coverage-pooled:
-    name: "Coverage (pooled delivery)"
-    uses: ./.github/workflows/build-test.yml
-    with:
-      coverage: ON
-      server_version: main
-      type: Debug
-      lib_msg_delivery: ON
-    secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/on-pr-debug.yml
```diff
@@ -56,33 +35,61 @@ jobs:
...
+      tls: TLS
+      verify_host: verify_host
+      pool_dispatch: ${{ matrix.pooled_dispatch }}
+      write_deadline: ${{ matrix.write_deadline }}
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-  sanitize-no-host-verify:
-    name: "Sanitize address (NO force_host_verification)"
+  coverage-NO-verify_host:
```
kozlovic, line 67: Should we have a single coverage run (basically with the defaults of the library that has TLS/enfore_hostname_verification/etc..)? 
Or is it that we would get a much lower score (for all code that is in the "#else" of those "#ifdef") without these new runs?

kozlovic (approved):
LGTM, but wondering if we need all those code coverage runs?

.github/workflows/on-pr-debug.yml
levb: Right. What prompted this was the refactoring of deliveries - the thread pool refactor code generated a lot of missing coverage comments since it was not ran with coverage.

My thinking: the coverage overhead is relatively small (relative to sanitize) so it's ok to run a bigger matrix. For the same reason, the coverage runs are probably sufficient to assert functional correctness (no need to dial down the parameters, for example). We still run the no-coverage tests for the default build config.

FWIW, I failed to package them into a single compact matrix statement, but the current job set is probably more intention revealing anyway then the matrix would have been.

PR #797, opened 2024-09-04
Add implementation for filtering keys from kv store

saurabhojha:
A draft pr resolving the following issue in nats c client : https://github.com/nats-io/nats-architecture-and-design/issues/289

### DIFF
--- a/src/kv.c
+++ b/src/kv.c
@@ -1242,6 +1242,76 @@ kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
     return NATS_UPDATE_ERR_STACK(s);
 }
 
+NATS_EXTERN natsStatus
+kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts,char **filters,int numFilters) {
+    natsStatus      s;
+    kvWatchOptions  o;
+    kvWatcher       *w = NULL;
+    int             count = 0;
+    KV_DEFINE_LIST;
+
+    if (list == NULL)
+        return nats_setDefaultError(NATS_INVALID_ARG);
+
+    if(filters == NULL || numFilters == 0) {
+        return nats_setDefaultError(NATS_INVALID_ARG);
+    }
+
+    list->Keys = NULL;
+    list->Count = 0;
+
+    kvWatchOptions_Init(&o);
+    if (opts != NULL)
+        memcpy(&o, opts, sizeof(kvWatchOptions));
+
+    o.IgnoreDeletes = true;
+    o.MetaOnly = true;
+    o.IncludeHistory = false;
+    if (o.Timeout > 0)
+        timeout = o.Timeout;
+
+    s = kvStore_WatchMulti(&w, kv,filters,numFilters,&o);
+    if (s != NATS_OK)
+        return NATS_UPDATE_ERR_STACK(s);
+
+    KV_GATHER_LIST;
+
+    // Don't need the watcher anymore.
+    kvWatcher_Destroy(w);
+    // On success, create the array of keys.
+    if ((s == NATS_OK) && (n > 0))
+    {
+        list->Keys = (char**) NATS_CALLOC(n, sizeof(char*));
+        if (list->Keys == NULL)
+            s = nats_setDefaultError(NATS_NO_MEMORY);
+    }
+    // Transfer keys to the array (on success), and destroy
+    // the entries if there was an error.
+    for (i=0; h != NULL; i++)
+    {
+        e = h;
+        h = h->next;
+        if (s == NATS_OK && e->op == kvOp_Put)
+        {
+            DUP_STRING(s, list->Keys[i], e->key);
+            if (s == NATS_OK)
+                count++;
+        }
+        kvEntry_Destroy(e);
+    }
+    // Set the list's Count to `count`, not `n` since `count`
+    // will reflect the actual number of keys that have been
+    // properly strdup'ed.
+    list->Count = count;
+
+    // If there was a failure (especially when strdup'ing) keys,
+    // this will do the proper cleanup and re-initialize the list.
+    if (s != NATS_OK)
+        kvKeysList_Destroy(list);
+
+    return NATS_UPDATE_ERR_STACK(s);
+}
+
 void
 kvKeysList_Destroy(kvKeysList *list)
 {
--- a/src/nats.h
+++ b/src/nats.h
@@ -7172,6 +7172,9 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts);
 NATS_EXTERN natsStatus
 kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts);
 
+NATS_EXTERN natsStatus
+kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts,char **filters,int numFilters);
+
 /** \brief Destroys this list of KeyValue store key strings.
  *
  * This function iterates through the list of all key strings and free them.


src/kv.c
```diff
@@ -1242,6 +1242,76 @@ kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
     return NATS_UPDATE_ERR_STACK(s);
 }
 
+NATS_EXTERN natsStatus
```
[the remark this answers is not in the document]
kozlovic: Also, when adding the new function, check only for `filters` and `numFilters` validity and then invoke the internal, all other checks (kv not NULL, etc..) will be done there. And don't put `NATS_EXTERN` in the `.c` file, only in `nats.h`.

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    kvKeysList          l;
+    char                *k = NULL;
+    kvConfig            kvc;
+    kvWatchOptions      o;
+    int                 i;
+    const char **defaultSubject = (const char *[]){">"};
+    l.Count = 0;
+    l.Keys = NULL;
+
+    JS_SETUP(2, 10, 14);
```
kozlovic, line 31486: Not sure we need that recent of a server version for this, do we?

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    test("Populate: ");
+    s = kvStore_PutString(NULL, kv, "a.b", "a.b");
+    IFOK(s, kvStore_PutString(NULL, kv, "a.d", "a.d"));
+    IFOK(s, kvStore_PutString(NULL, kv, "c.d", "c.d"));
+    IFOK(s, kvStore_PutString(NULL, kv, "e.f", "e.f"));
+    IFOK(s, kvStore_PutString(NULL, kv, "e.a.f", "e.a.f"));
+    testCond(s == NATS_OK);
+
+    test("Get keys with filters (bad args): List is null");
+    // list is NULL
```
kozlovic, line 31504: Remove, it is clear from above :-)

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    s = kvStore_KeysWithFilters(NULL, kv, NULL,defaultSubject,1);
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
```
kozlovic, line 31506: Since you don't pass `l`, those checks (keys and count) are not needed.

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): filters is null");
+    // filters is NULL
```
kozlovic, line 31510: Again, not needed.

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    s = kvStore_KeysWithFilters(NULL, kv, NULL,NULL,0);
```
kozlovic, line 31511: Please to be consistent use spaces after the `,` when passing/defining args. This applies to to here but check other places too.

kozlovic: You need to pass `&l` here, otherwise you are possibly still checking that the list cannot be NULL.

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): numFilters is 0");
+    // filters is NULL
```
kozlovic, line 31516: Remove

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    s = kvStore_KeysWithFilters(&l, kv, NULL,defaultSubject,0);
```
kozlovic, line 31517: Reminder: spaces...

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): kv is NULL");
+    // filters is NULL
```
kozlovic, line 31522: Remove

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    s = kvStore_KeysWithFilters(&l, NULL, NULL,defaultSubject,1);
```
kozlovic, line 31523: Not repeating for the rest of the test: please check spaces...

kozlovic (changes requested):
Thank you for the contribution. I made some comments. Lots of them are just code style, but the most important is to not duplicate the code for the main function.

src/nats.h
```diff
@@ -7172,6 +7172,31 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts);
...
+ * @see kvStore_WatchMulti
+ *
+ * @param list list the pointer to a #kvKeysList that will be initialized and filled with resulting key strings.
+ * @param kv the pointer to the #kvStore object.
+ * @param opts the history options, possibly `NULL`.
+ * @param filters the list of subject like string filters: Cannot be `NULL`
+ * @param numFilters number of filters: Cannot be 0
+ */
+NATS_EXTERN natsStatus
+kvStore_KeysWithFilters(kvKeysList *list, kvStore *kv, kvWatchOptions *opts,char **filters,int numFilters);
```
kozlovic, line 7198: Make `char **filters` a `const char **filters`.

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
     JS_TEARDOWN;
 }
 
+void test_KeyValueKeysWithFilters(void)
+{
+    natsStatus          s;
+    kvStore             *kv = NULL;
+    kvKeysList          l;
+    char                *k = NULL;
```
kozlovic, line 31478: Not used, remove.

test/test.c
```diff
@@ -31470,6 +31470,140 @@ void test_KeyValueKeys(void)
...
+    kvConfig            kvc;
+    kvWatchOptions      o;
+    int                 i;
```
kozlovic, line 31481: Not used, remove both.

test/test.c
kozlovic: Well, it seems we do need that late of a version for one of the tests (multiple overlapping filters).

kozlovic (changes requested):
Some more modifications needed.

src/nats.h
```diff
@@ -7172,6 +7172,28 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts);
 NATS_EXTERN natsStatus
 kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts);
 
+/** \brief Returns all keys in the bucket which matches the list of subject like filters.
+ *  Get a list of the keys in a bucket filtered by a
```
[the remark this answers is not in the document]
saurabhojha: thanks for the suggestion. Made the change.

src/nats.h
```diff
@@ -7172,6 +7172,28 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts);
...
+ * for the entries list.
+ *
+ * @see kvWatchOptions_Init
+ * @see kvKeysList_Destroy
+ * @see kvStore_WatchMulti
+ *
+ * @param list list the pointer to a #kvKeysList that will be initialized and filled with resulting key strings.
+ * @param kv the pointer to the #kvStore object.
+ * @param opts the history options, possibly `NULL`.
+ * @param filters the list of subject like string filters: Cannot be `NULL`
```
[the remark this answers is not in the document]
saurabhojha: Thanks for pointing this out, changed as suggested

src/nats.h
```diff
@@ -7172,6 +7172,28 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts);
...
+ * will return #NATS_TIMEOUT and no keys.
+ *
+ * \warning The user should call #kvKeysList_Destroy to release memory allocated
+ * for the entries list.
+ *
+ * @see kvWatchOptions_Init
+ * @see kvKeysList_Destroy
+ * @see kvStore_WatchMulti
+ *
+ * @param list list the pointer to a #kvKeysList that will be initialized and filled with resulting key strings.
```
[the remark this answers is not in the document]
saurabhojha: Removed extra list, thanks for pointing this out.

src/nats.h
```diff
@@ -7172,6 +7172,28 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts);
...
+ * @param kv the pointer to the #kvStore object.
+ * @param opts the history options, possibly `NULL`.
+ * @param filters the list of subject like string filters: Cannot be `NULL`
+ * @param numFilters number of filters: Cannot be 0
```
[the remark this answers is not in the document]
saurabhojha: Fixed the punctuation here. Thanks for pointing this out.

src/kv.c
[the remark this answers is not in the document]
saurabhojha: Thanks for the suggestion. 
I have refactored the code to a static internal function _kvStore_Keys which will be called by both kvStore_Keys and kvStore_KeysWithFilters now.

test/test.c
saurabhojha: Good catch, fixed and tested this.

src/kv.c
```diff
@@ -27,6 +27,9 @@ static const char *kvSubjectsTmpl           = "$KV.%s.>";
 static const char *kvSubjectsPreTmpl        = "$KV.%s.";
 static const char *kvSubjectsPreDomainTmpl  = "%s.$KV.%s.";
 
+static natsStatus
```
kozlovic, line 30: Would not need that if you have the static before the public functions. See below.

src/kv.c
```diff
@@ -1178,16 +1181,31 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts)
 
 natsStatus
 kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
```
kozlovic, line 1183: I would have made this one the static and add the filters/numFilters param. Would have moved the 2 public APIs after it, so no need for forward declaration.

test/test.c
saurabhojha: Good catch, changed and tested this. Thanks for this.

src/kv.c
```diff
@@ -1178,16 +1181,31 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts)
...
+static natsStatus
+_kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts, const char **filters, int numFilters)
 {
     natsStatus      s;
     kvWatchOptions  o;
     kvWatcher       *w = NULL;
     int             count = 0;
     KV_DEFINE_LIST;
 
-    if (list == NULL)
```
kozlovic, line 1188: Would have left this one here since it is common to both Keys() and KeysWithFilter()

src/kv.c
```diff
@@ -1200,7 +1218,14 @@ kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
     if (o.Timeout > 0)
         timeout = o.Timeout;
 
-    s = kvStore_WatchAll(&w, kv, &o);
+    if (filters == NULL) {
+        s = kvStore_WatchAll(&w, kv, &o);
+    } else {
+        // Deliver policy gets set to DeliverLastPerSubject
+        o.IncludeHistory = false;
```
kozlovic, line 1225: That I don't understand. `kvStore_WatchAll()` is already calling `kvStore_WatchMulti()` with the ">" filter, and there was no need to add anything to it.

src/kv.c
```diff
@@ -1223,7 +1248,13 @@ kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
         h = h->next;
         if (s == NATS_OK)
         {
-            DUP_STRING(s, list->Keys[i], e->key);
+            if(filters!=NULL && numFilters > 0) {
```
kozlovic, line 1251: Same, I don't understand why you have to treat this differently than a WatchAll(). If there is a filter and the server is doing proper filtering, you will get anything you need same than WatchAll().

The point of my comment of making it this way was because I assumed that it was just a matter of passing filters vs the single filter ">". But rest was the same. I checked the Java implementation and does not look like he changed anything.

kozlovic (changes requested):
Before I continue the review, please make sure that your changes in the core Keys() function are needed. I am not sure they are.

src/kv.c
saurabhojha: My concern is changing the signature of this older function kvStore_Keys would cause a backward incompatible change by breaking the contract.
Let me know if you think otherwise.

kozlovic: I did not mean to change the public signature. I would simply have kept the body of this function but made it static, then have `kvStore_Keys()` and the new one invoke the static one. It is simply that the static would have the same params than the one with filters because it can do both. Let me know if that is still not clear.

kozlovic: Basically, the changes in `kv.c` could be as simple as that (diffs from main):
```
diff --git a/src/kv.c b/src/kv.c
index 2b079680..09727045 100644
--- a/src/kv.c
+++ b/src/kv.c
@@ -1176,8 +1176,8 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts)
 }
 
-natsStatus
-kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
+static natsStatus
+_kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts, const char **filters, int numFilters)
 {
     natsStatus      s;
     kvWatchOptions  o;
@@ -1200,7 +1200,10 @@ kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
     if (o.Timeout > 0)
         timeout = o.Timeout;
 
-    s = kvStore_WatchAll(&w, kv, &o);
+    if (numFilters > 0)
+        s = kvStore_WatchMulti(&w, kv, filters, numFilters, &o);
+    else
+        s = kvStore_WatchAll(&w, kv, &o);
     if (s != NATS_OK)
 
@@ -1242,6 +1245,25 @@ kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
 }
 
+natsStatus
+kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
+{
+    natsStatus s = _kvStore_Keys(list, kv, opts, NULL, 0);
+}
+
+natsStatus
+kvStore_KeysWithFilters(kvKeysList *list, kvStore *kv, kvWatchOptions *opts, const char **filters, int numFilters)
+{
+    natsStatus s;
+
+    if ((filters == NULL) || (numFilters <= 0))
+
+    s = _kvStore_Keys(list, kv, opts, filters, numFilters);
+}
+
 void
 kvKeysList_Destroy(kvKeysList *list)
 {
```

test/test.c
kozlovic: Still a space: `,0);`.

test/test.c
```diff
@@ -31470,6 +31470,133 @@ void test_KeyValueKeys(void)
...
+    s = kvStore_KeysWithFilters(&l, kv, NULL, NULL,0);
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): numFilters is 0");
+    s = kvStore_KeysWithFilters(&l, kv, NULL, defaultSubject,0);
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): kv is NULL");
```
kozlovic, line 31515: Before this test, you could add:
```
test("Get keys with filters (bad args): numFilters is <0");
s = kvStore_KeysWithFilters(&l, kv, NULL, defaultSubject, -10);
testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
nats_clearLastError();

test("Get keys with filters (bad args): empty string");
const char **filter0 = (const char *[]){"a.*", "", "b.*"};
s = kvStore_KeysWithFilters(&l, kv, NULL, filter0, 3);
testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
nats_clearLastError();
```
Note that you don't have to specifically check in your code for empty filter strings since this is done by WatchMulti. The only check you do is that it is not NULL or <=0, because this is what you need to differentiate which call `_kvStore_Keys()` is going to make. But there is no need to "redo" the checks that are done in the call stack.

kozlovic (changes requested):

test/test.c
```diff
@@ -31470,6 +31470,133 @@ void test_KeyValueKeys(void)
...
+    IFOK(s, kvStore_PutString(NULL, kv, "e.a.f", "e.a.f"));
+    testCond(s == NATS_OK);
+
+    test("Get keys with filters (bad args): List is NULL");
+    s = kvStore_KeysWithFilters(NULL, kv, NULL, defaultSubject, 1);
+    testCond(s == NATS_INVALID_ARG);
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): filters is NULL");
+    s = kvStore_KeysWithFilters(&l, kv, NULL, NULL,0);
```
kozlovic, line 31506: Space.

test/test.c
```diff
@@ -31470,6 +31470,133 @@ void test_KeyValueKeys(void)
...
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): numFilters is 0");
+    s = kvStore_KeysWithFilters(&l, kv, NULL, defaultSubject,0);
```
kozlovic, line 31511: Space.

test/test.c
```diff
@@ -31470,6 +31470,133 @@ void test_KeyValueKeys(void)
...
+    testCond((s == NATS_INVALID_ARG) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+
+    test("Get keys with filters (bad args): kv is NULL");
+    s = kvStore_KeysWithFilters(&l, NULL, NULL, defaultSubject,1);
```
kozlovic, line 31516: Space.

test/test.c
```diff
@@ -31470,6 +31470,133 @@ void test_KeyValueKeys(void)
...
+    test("a.b should not be returned post deletion")
+    const char **filter8 = (const char *[]){"a.b"};
+    s = kvStore_KeysWithFilters(&l, kv, NULL, filter8, 1);
+    testCond((s == NATS_OK) && (l.Keys == NULL) && (l.Count == 0));
+    nats_clearLastError();
+    kvKeysList_Destroy(&l);
+
+    // Purge the key and check if returned after filtering
+    test("Purge a.b:");
+    s = kvStore_Purge(kv, "a.d",NULL);
```
kozlovic, line 31585: Space.

kozlovic (changes requested):
By the way, you may want to combine all subsequent changes in a single commit before pushing it (to reduce CI runs).

src/kv.c
saurabhojha: In java we have the following:
https://github.com/nats-io/nats.java/blob/d69587382c854abe7c4d840cb85c5846e7b0351e/src/main/java/io/nats/client/impl/NatsKeyValue.java#L305
There is an explicit filtering for PUT kv operation type.

My assumption is if the server is not returning deleted/purged messages, then the IgnoreDelete option set as true should be indicating that. In that case why is the PUT operation explicitly handled in java client.

kozlovic: We set the IgnoreDelete set to true, which means that internally, when we get the result from the server, we will ignore deletes and already collect only "PUT" operations. So you don't need to do any additional work (see my diff I sent, it works with your new test):

https://github.com/nats-io/nats.c/blob/381f01cb75acb6ff0860a599ff29b17127d21dac/src/kv.c#L1005

kozlovic: In other words, where we "dup" the strings, we have already filtered properly in `KV_GATHER_LIST;`

kozlovic: This was based on the go client: https://github.com/nats-io/nats.go/blob/b61c7c554f188ff40e15fb3c4668bad959ccde86/kv.go#L1012

test/test.c
saurabhojha: Thanks for the suggestion. will add these tests as well.

src/kv.c
saurabhojha: Thanks for the explanation. I went through the go client as well. I was redoing the work already done in kvWatcher_Next.

src/kv.c
saurabhojha: Thanks for the explanation. my assumption was that you wanted to keep the name of the internal method same as kvStore_Keys() (that is why I was concerned about changing the signature as c doesn't support function overloading).

saurabhojha:
Hi @kozlovic , made the changes you requested. Please let me know if anything else has to be changed.

kozlovic (changes requested):
Let's move the options around.

But, this brings the question: @levb, had you consider having the filter options as part of the `kvWatchOptions` instead of adding a `kvStore_WatchMulti()`? In PR https://github.com/nats-io/nats.c/pull/750, I don't see me suggesting that, but since this has never been released yet, it would be possible to change it. If that had been the case, there would not even be a need for a new `kvStore_KeysWithFilters()` API, since `kvStore_Keys()` accepts a `kvWatchOptions`, we could have passed the filters through that (we would have needed to change a bit to use filters if specified and call Watch() - or if documenting that Watch() uses filters over provided key, no change at all).

kozlovic:
@saurabhojha @levb My comment about passing filters through options may be a moot point if the ADR asks libraries to add a new API.

src/kv.c
```diff
@@ -1176,8 +1176,7 @@ kvStore_WatchAll(kvWatcher **new_watcher, kvStore *kv, kvWatchOptions *opts)
     return NATS_UPDATE_ERR_STACK(s);
 }
 
-natsStatus
-kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts)
+static natsStatus _kvStore_Keys(kvKeysList *list, kvStore *kv, kvWatchOptions *opts, const char **filters, int numFilters)
```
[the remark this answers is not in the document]
saurabhojha: agreed on this, keeps the order of params through the function calls the same.

kozlovic (approved):
LGTM! But I will let @levb decide if we merge as-is or if he wants to revisit the WatchMulti or not. If we do merge this PR, I would recommend the "squash on merge", since there are a lot of commits on that one.

kozlovic:
@levb So sorry to hear that. Please rest, there is really no urgency in reviewing this PR.

saurabhojha:
@levb take care and please rest. this can be reviewed later as well.

levb:
Let me confirm with @Jarema, I like the simplification of using `kvWatchOptions`.

PR #800, opened 2024-09-23
[ADDED, BREAKING] Services: queue group now configurable and can be disabled

levb:
Compliance with https://github.com/nats-io/nats-architecture-and-design/pull/239 and https://github.com/nats-io/nats-architecture-and-design/pull/300

One can now customize the queue group used for microservice endpoints, or disable its use.


levb:
Looking into the test failure, please hold off the review.

src/sub.c
```diff
@@ -89,21 +89,23 @@ _initOwnDispatcher(natsSubscription *sub)
...
-static inline void _cleanupOwnDispatcher(natsSubscription *sub)
+static inline void
+_cleanupOwnDispatcher(natsSubscription *sub)
 {
     nats_destroyQueuedMessages(&sub->ownDispatcher.queue);
 
     if (sub->ownDispatcher.thread != NULL)
     {
-        natsThread_Join(sub->ownDispatcher.thread);
+        natsThread_Detach(sub->ownDispatcher.thread);
```
levb, line 99: @kozlovic this was an oversight on my end I think since this code can be executed in the dispatch thread itself. Anyway, this restores the original logic. It did not appear to affect the test failure here, the wait in the test did.

src/nats.h
```diff
@@ -7813,6 +7833,11 @@ struct micro_endpoint_info_s
      */
     const char *Subject;
 
+    /**
+     * @brief Endpoint's queue group, or omitted if no queue group is used.
```
kozlovic, line 7837: Is it "omitted if no queue group is used" or if no "override"? Meaning, if I don't set something here, are we going to use the default, unless the NoQueueGroup is set?

src/nats.h
```diff
@@ -7828,6 +7853,11 @@ struct micro_endpoint_stats_s
     const char *Name;
     const char *Subject;
 
+    /**
+     * @brief Endpoint's queue group, or omitted if no queue group is used.
```
kozlovic, line 7857: Same here.

src/nats.h
```diff
@@ -7781,7 +7789,19 @@ struct micro_endpoint_config_s
     const char *Subject;
 
     /**
-     * @briefMetadata for the endpoint, a JSON-encoded user-provided object,
+     * @brief overrides the default queue group for the service.
```
kozlovic, line 7792: `Overrides` (start of sentence)?

src/nats.h
```diff
@@ -7860,6 +7890,29 @@ struct micro_endpoint_stats_s
...
+ */
+struct micro_group_config_s
+{
+    /**
+     * @brief The subject prefix for the group.
+     */
+    const char *Prefix;
+
+    /**
+     * @brief overrides the default queue group for the service.
```
kozlovic, line 7904: `Overrides` (start of sentence)?

src/nats.h
```diff
@@ -7886,7 +7939,20 @@ struct micro_service_config_s
     const char *Description;
 
     /**
-     * @brief Metadata for the service, a JSON-encoded user-provided object, e.g. `{"key":"value"}`
+     * @brief overrides the default queue group for the service ("q").
```
kozlovic, line 7942: `Overrides` (start of sentence)?

test/test.c
```diff
@@ -33969,6 +33968,118 @@ void test_MicroGroups(void)
...
+        bool epNoQueueGroup;
+        const char *epQueueGroup;
+
+        const char *expectedServiceLevel;
+        const char *expectedGroup1Level;
+        const char *expectedGroup2Level;
+    } TC;
+    TC tcs[] = {
+        {.name="default",
+            .expectedServiceLevel="q", .expectedGroup1Level="q", .expectedGroup2Level="q"},
```
kozlovic, line 34001: Maybe replace `"q"` with `MICRO_DEFAULT_QUEUE_GROUP` since we include the header that defines it? True for all references of "q" in this file I believe.

test/test.c
```diff
@@ -34127,6 +34238,7 @@ void test_MicroBasics(void)
         testCond(
             (NATS_OK == nats_JSONGetStrPtr(array[0], "name", &str)) && (strcmp(str, "do") == 0)
             && (NATS_OK == nats_JSONGetStrPtr(array[0], "subject", &str)) && (strcmp(str, "svc.do") == 0)
+            && (NATS_OK == nats_JSONGetStrPtr(array[0], "queue_group", &str)) && (strcmp(str, "q") == 0)
```
kozlovic, line 34241: In case you did not "replace all", this would be one place to use `MICRO_DEFAULT_QUEUE_GROUP`.

test/test.c
```diff
@@ -34135,6 +34247,7 @@ void test_MicroBasics(void)
         testCond(
             (NATS_OK == nats_JSONGetStrPtr(array[1], "name", &str)) && (strcmp(str, "unused") == 0)
             && (NATS_OK == nats_JSONGetStrPtr(array[1], "subject", &str)) && (strcmp(str, "svc.unused") == 0)
+            && (NATS_OK == nats_JSONGetStrPtr(array[0], "queue_group", &str)) && (strcmp(str, "q") == 0)
```
kozlovic, line 34250: And here too.

kozlovic (changes requested):
Little changes, but otherwise LGTM.

src/nats.h
levb: Clarified in the comment (the actual value, or "nil")

src/nats.h
levb: ditto

kozlovic (approved):
LGTM

PR #801, opened 2024-09-30
FIXED: nats_GetJWTOrSeed to understand Windows \r\n lines

levb:
A customer ran into this issue, no ticket opened.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -2064,10 +2064,9 @@ nats_HostIsIP(const char *host)
 }
 
 static bool
-_isLineAnHeader(const char *ptr)
+_isLineAnHeader(const char *ptr, int len)
 {
     char    *last   = NULL;
-    int     len     = 0;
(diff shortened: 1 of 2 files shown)


src/util.c
```diff
@@ -2118,61 +2116,72 @@ _isLineAnHeader(const char *ptr)
...
-            if (pch == NULL)
-                break;
-
-            if (_isLineAnHeader(pch))
+            lineLen = _scan_line(&line, &next);
+            if (_isLineAnHeader(line, lineLen))
             {
                 // Is this the item we were looking for?
-                if (curItem == item)
+                if ((curItem == item) && (savedLen > 0))
```
kozlovic, line 2177: Check on `savedLen` not needed since you already break out of loop if `savedLen == 0`.

kozlovic (approved):
nit: unnecessary check, but otherwise LGTM

PR #810, opened 2024-10-02
fix(build): only append `-beta` for non-release builds

chenrui333:
for 3.9.0, the NATS_VERSION_STRING is `3.9.0-beta` instead of `3.9.0`, update the build config to only append `-beta` for non-release builds

relates to https://github.com/Homebrew/homebrew-core/pull/192554


kozlovic:
@chenrui333 Hum... I think this is going to be a problem. We changed a little while back the way releases are done, and they are no longer done from main branch, instead we have "release" branches. I reviewed the v3.9.0 PR yesterday and the "-beta" was removed.

My suspicion is that homebrew does it from main? I have added @levb here to comment.

kozlovic:
@levb From the release page, v3.9.0 points to the commit on the release branch https://github.com/nats-io/nats.c/commit/233ca8efc428d5b84b41f1a9aaedcd81441ed6bc, but this is the one before the update of release version: https://github.com/nats-io/nats.c/commit/28a5948ddb76956125c6d4af1ac10796c95a5d55.

We should redo as a v3.9.1 to fix all that.

@chenrui333 I don't think this PR is necessary, instead, I think this release was just incorrect. But I will @levb have the final word.

chenrui333:
yeah, works for me, but I also raised this question on the tag commit as well.

chenrui333:
looks like both of these two commits got tagged with 3.9.0:
- https://github.com/nats-io/nats.c/commit/8e6a22827bdd63cb3f4617ce6f9db5793f7d8e61
- https://github.com/nats-io/nats.c/commit/233ca8efc428d5b84b41f1a9aaedcd81441ed6bc

chenrui333:
since the 3.9.0 release is wrong, can we maybe re-tag it?

kozlovic:
@chenrui333 No, I would prefer to move to v3.9.1 than re-tag. I know this is no Golang, but we had very bad experience in the past because some proxies already captured a given tag and when we re-tagged a release it made a mess. So moving forward we decided to own our mistake and bump the version instead of re-tagging.

chenrui333:
yeah, I would also prefer to have a new release. Should I just close the related homebrew PR for the new release then? Thanks!

kozlovic:
> Should I just close the related homebrew PR for the new release then? Thanks!

Sure, once @levb release v3.9.1, the homebrew PR can be done. But homebrew has people or automatic processes that submit PRs when detecting a new NATS C release, so we may still have the hombrew PR for v3.9.0 popping before we are ready.

levb:
@chenrui333 @kozlovic This was my mistake to pass `-beta` through, working on a 3.9.1 now, it's already fixed in the 3_9 branch.

levb:
@chenrui333 https://github.com/nats-io/nats.c/releases/tag/v3.9.1

We add `-beta` in main to make clear that it is not a release-quality build, so 3/5 we should keep the existing CMakeLists.txt as is. Unfortunately the release process is not yet fully automated, and I made the mistake of not removing it. Please try v3.9.1?

chenrui333:
close via https://github.com/Homebrew/homebrew-core/pull/192663

chenrui333:
Thanks @levb @kozlovic!

Issue #812, opened 2024-10-10
Allow users to define required version in cmake


rolflussi:
### Proposed change

Would it be possible to provide a cnats-config-version.cmake file to allow any cmake based project to define the required version of this library within the `find_package` command?

### Use case

Any cmake project can define the required version in its CMakeLists.txt file like this:

``` 
find_package(cnats 3.9 REQUIRED)
```

### Contribution

I already have a first implementation on a fork, so I would be happy to open a pull request if you are interested.

levb:
@rolflussi please feel free to open a PR on this! Will be much appreciated!

PR #813, opened 2024-10-10
Generate cnats-config-version.cmake

rolflussi:
Implementation of https://github.com/nats-io/nats.c/issues/812

Allows other cmake projects to use this library with a specific version:

```cmake
find_package(cnats 3.9 REQUIRED)   # requires a version >= 3.9 and < 4.0
```

The cnats-config-version.cmake will make sure that the correct major version and at least the defined minor version is available.


CMakeLists.txt
```diff
@@ -258,6 +259,11 @@ set(NATS_VERSION_SUFFIX "-beta")
 
 set(NATS_VERSION_REQUIRED_NUMBER 0x030900)
 
+write_basic_package_version_file(
+  "${PROJECT_BINARY_DIR}/cnats-config-version.cmake"
+  COMPATIBILITY SameMajorVersion
```
levb, line 264: @rolflussi I do not have much personal experience with `cmake` package management/discovery, but what I understood from https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file, I think we should be using `SameMinorVersion` if binary compatibility is desired. Furthermore, binary compatibility may be broken even in patch releases (thus, `NATS_VERSION_REQUIRED_NUMBER` check), I am not sure what the best practice on this is. 

@kozlovic do you have an opinion on this?

kozlovic: I don't think we do (or should) break compatibility in a patch release. But yes, minor release may introduce new APIs, etc.. that may cause their local header file to not be in phase with the linked library (this is why we have some API to check for compatibility between the header and the lib's version).

rolflussi: Thanks for your feedback. I'll change it to SameMinorVersion

levb: (I misspoke re: patch releases; I was thinking a hypothetical emergency patch that'd break the memory layout, but we'd just release it as a `X.Y+1.0` from the `X.Y` branch.

levb (approved):
LGTM

levb:
Cherry-picked into v3.9.2

PR #816, opened 2024-11-01
[FIXED] microservice cleanup (flapping MicroServiceStops... tests)

levb:
- Removed the global hash of services, it was the most immediate source of the deadlock.
- Added the list of microservices to `natsConnection`. **NOTE** increased memory allocation for all connections, need a better way.
- Fixed the flow of replacing an existing endpoint in a service.
- Got rid of ep->name since it was synonimous with ep->cfg->Name
- Adjusted the tests, some minor fixes.

### DIFF
--- a/src/micro.c
+++ b/src/micro.c
@@ -184,15 +184,15 @@ micro_add_endpoint(microEndpoint **new_ep, microService *m, microGroup *g, micro
     {
         if (update)
         {
-                // If the endpoint already exists, update its config and stats.
-                // This will make it use the new handler for subsequent
-                // requests.
-                micro_lock_endpoint(ep);
-                micro_free_cloned_endpoint_config(ep->config);
-                err = micro_clone_endpoint_config(&ep->config, cfg);
-                if (err == NULL)
-                    memset(&ep->stats, 0, sizeof(ep->stats));
-                micro_unlock_endpoint(ep);
+            // If the endpoint already exists, update its config and stats.
+            // This will make it use the new handler for subsequent
+            // requests.
+            micro_lock_endpoint(ep);
+            micro_free_cloned_endpoint_config(ep->config);
+            err = micro_clone_endpoint_config(&ep->config, cfg);
+            if (err == NULL)
+                memset(&ep->stats, 0, sizeof(ep->stats));
+            micro_unlock_endpoint(ep);
         }
         else
         {
--- a/src/micro.c
+++ b/src/micro.c
@@ -301,7 +301,7 @@ _detach_service_from_connection(natsConnection *nc, microService *m)
     {
         if (nc->services[i] != m)
             continue;
-        
+
         nc->services[i] = nc->services[nc->numServices - 1];
         nc->numServices--;
         break;
@@ -372,7 +372,7 @@ _detach_endpoint_from_service(microService *m, microEndpoint *toRemove)
        ;
     if (ep == NULL)
         return;
-    
+
     m->numEndpoints--;
     if (prev_ep == NULL)
         m->first_ep = ep->next;
@@ -627,14 +627,14 @@ _on_service_error(microService *m, const char *subject, natsStatus s)
 
     if (found == NULL)
         return false;
-    
+
     err = microError_Wrapf(micro_ErrorFromStatus(s), "NATS error on endpoint '%s'", subject);
     micro_update_last_error(found, err);
     microError_Destroy(err);
 
     if (m->cfg->ErrHandler != NULL)
         (*m->cfg->ErrHandler)(m, found, s);
-    
+
     micro_release_endpoint(found);
     return true;
 }
--- a/src/micro.c
+++ b/src/micro.c
@@ -29,6 +29,9 @@ static void _on_error(natsConnection *nc, natsSubscription *sub, natsStatus s, v
 static void _free_cloned_service_config(microServiceConfig *cfg);
 static void _free_service(microService *m);
 
+static microError *
+_attach_service_to_connection(natsConnection *nc, microService *service);
+
 microError *
 micro_AddService(microService **new_m, natsConnection *nc, microServiceConfig *cfg)
 {
@@ -53,7 +56,7 @@ micro_AddService(microService **new_m, natsConnection *nc, microServiceConfig *c
     MICRO_CALL(err, _clone_service_config(&m->cfg, cfg));
 
     // Add the service to the connection.
-    MICRO_CALL(err, micro_ErrorFromStatus(natsConn_addService(m->nc, m)));
+    MICRO_CALL(err, _attach_service_to_connection(m->nc, m));
     MICRO_CALL(err, (m->refs++, NULL));
 
     // Wrap the connection callbacks before we subscribe to anything.
@@ -244,7 +247,62 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
 }
 
 static microError *
-_stop_service(microService *m, bool unsubscribe, bool finalRelease)
+_attach_service_to_connection(natsConnection *nc, microService *service)
+{
+    natsStatus s = NATS_OK;
+    if (nc == NULL || service == NULL)
+        return micro_ErrorInvalidArg;
+
+    natsMutex_Lock(nc->servicesMu);
+    if (nc->services == NULL)
+    {
+        nc->services = NATS_CALLOC(1, sizeof(microService *));
+        if (nc->services == NULL)
+            s = nats_setDefaultError(NATS_NO_MEMORY);
+    }
+    else
+    {
+        microService **tmp = NATS_REALLOC(nc->services, (nc->numServices + 1) * sizeof(microService *));
+        if (tmp == NULL)
+            s = nats_setDefaultError(NATS_NO_MEMORY);
+        else
+            nc->services = tmp;
+    }
+
+    if (s == NATS_OK)
+    {
+        nc->services[nc->numServices] = service;
+        nc->numServices++;
+    }
+    natsMutex_Unlock(nc->servicesMu);
+
+    return micro_ErrorFromStatus(s);
+}
+
+static void
+_detach_service_from_connection(natsConnection *nc, microService *m)
+{
+    if (nc == NULL || m == NULL)
+        return;
+
+    natsMutex_Lock(nc->servicesMu);
+    for (int i = 0; i < nc->numServices; i++)
+    {
+        if (nc->services[i] != m)
+            continue;
+        
+        for (int j = i; j < nc->numServices - 1; j++)
+            nc->services[j] = nc->services[j + 1];
+        nc->numServices--;
+        break;
+    }
+    natsMutex_Unlock(nc->servicesMu);
+
+    return;
+}
+
+static microError *
+_stop_service(microService *m, bool unsubscribe, bool release)
 {
     microError      *err            = NULL;
     microEndpoint   *ep             = NULL;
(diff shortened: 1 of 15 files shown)


src/micro_endpoint.c
```diff
@@ -16,53 +16,11 @@
 #include "microp.h"
 #include "util.h"
 
-static microError *_subjectWithGroupPrefix(char **dst, microGroup *g, const char *src);
```
levb, line 19: These were used only in micro.c, so moved there

src/micro_endpoint.c
```diff
@@ -132,31 +89,20 @@ micro_stop_endpoint(microEndpoint *ep)
...
     natsSubscription *sub = NULL;
 
-    if ((ep == NULL) || (ep->m == NULL))
+    if (ep == NULL)
         return NULL;
 
     micro_lock_endpoint(ep);
     sub = ep->sub;
-
-    if (ep->is_draining || natsConnection_IsClosed(ep->m->nc) || !natsSubscription_IsValid(sub))
```
levb, line 141: ep->is_draining was at best synonymous with sub->draining, removed

src/natsp.h
```diff
@@ -699,6 +699,9 @@ struct __natsConnection
     natsHash            *subs;
     natsMutex           *subsMu;
 
+    microService        **services;
```
levb, line 702: 0/5 These take up memory in all connections; would be nice to come up with a better mechanism of attaching extended data to connections, without the fixed overhead.

test/test.c
```diff
@@ -33582,39 +33582,30 @@ _startManyMicroservices(microService** svcs, int n, natsConnection *nc, microSer
...
-
-    for (int i = 0; i < n; i++)
-    {
-        if (NULL != microService_Destroy(svcs[i]))
-            FAIL("Unable to destroy microservice!");
-    }
-
-    _waitForMicroservicesAllDone(arg);
-}
+#define _waitForMicroservicesAllDone(_arg)                                              \
```
levb, line 33585: The functions were not failing correctly before.

src/micro_endpoint.c
```diff
@@ -334,7 +279,7 @@ micro_clone_endpoint_config(microEndpointConfig **out, microEndpointConfig *cfg)
     microEndpointConfig *new_cfg = NULL;
 
     if (out == NULL)
-        return micro_ErrorInvalidArg;
+        return microError_Wrapf(micro_ErrorInvalidArg, "failed to clone endpoint config");
```
kozlovic, line 282: Should you add the name `cfg->Name` as you do few lines down (in the review)?

src/micro.c
```diff
@@ -23,12 +23,14 @@ static inline void _unlock_service(microService *m) { natsMutex_Unlock(m->servic
 
 static microError *_clone_service_config(microServiceConfig **out, microServiceConfig *cfg);
 static microError *_new_service(microService **ptr, natsConnection *nc);
-static microError *_wrap_connection_event_callbacks(microService *m);
+static void _on_connection_closed(natsConnection *nc, void *ignored);
```
kozlovic, line 26: nit: here you use `ignored` and the next you use `not_used`, could make them the same :-)

src/micro.c
```diff
@@ -53,10 +55,18 @@ micro_AddService(microService **new_m, natsConnection *nc, microServiceConfig *c
 
     MICRO_CALL(err, _clone_service_config(&m->cfg, cfg));
 
+    // Add the service to the connection.
+    MICRO_CALL(err, _attach_service_to_connection(m->nc, m));
+    MICRO_CALL(err, (m->refs++, NULL));
```
kozlovic, line 60: I am not saying we don't need a ref count here, but it seems to me that the connection should be retained (inside `_attach_service_to_connection()` maybe?). Meaning that you probably want the connection object to be valid while a service exists, no?

src/micro.c
```diff
@@ -53,10 +55,18 @@ micro_AddService(microService **new_m, natsConnection *nc, microServiceConfig *c
...
+
     // Wrap the connection callbacks before we subscribe to anything.
-    MICRO_CALL(err, _wrap_connection_event_callbacks(m));
+    MICRO_CALL(err, micro_ErrorFromStatus(
+                        natsOptions_setMicroCallbacks(m->nc->opts, _on_connection_closed, _on_error)));
```
kozlovic, line 64: So for instance even if the `nc` connection is valid at the time we enter `micro_AddService`, but happen to be closed during execution of this function, it will leave things in a bad state. If service is part of a connection, I think it should be handled by the connection, even if `micro_X` calls are just a proxy to internal `natsConn_X` calls. Have you thought that way?

src/micro.c
```diff
@@ -69,86 +79,152 @@ micro_AddService(microService **new_m, natsConnection *nc, microServiceConfig *c
...
-    if (m->first_ep != NULL)
+    if (err == NULL)
     {
-        if (strcmp(m->first_ep->subject, ep->subject) == 0)
+        if (update)
         {
-            ep->next = m->first_ep->next;
-            prev_ep = m->first_ep;
-            m->first_ep = ep;
+                // If the endpoint already exists, update its config and stats.
```
kozlovic, line 192: FYI: this block seem to have an extra tabulation.

src/micro.c
```diff
@@ -170,88 +246,145 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
...
+    for (int i = 0; i < nc->numServices; i++)
     {
-        natsLib_stopServiceCallbacks(m);
-        m->stopped = true;
-        doneHandler = m->cfg->DoneHandler;
+        if (nc->services[i] != m)
+            continue;
+        
+        for (int j = i; j < nc->numServices - 1; j++)
+            nc->services[j] = nc->services[j + 1];
```
kozlovic, line 295: Question: Is the order of the services important? Meaning, do we have to keep the same order when one is removed? If yes, then code is ok here, if no, then you could move the last to this spot. But that being said, you would have to check the it was not the last one, which, in term of number of lines of code the code here is probably shorter ;-)

src/micro.c
```diff
@@ -170,88 +246,145 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
...
+        nc->numServices--;
+        break;
     }
+    natsMutex_Unlock(nc->servicesMu);
 
-    _unlock_service(m);
+    return;
```
kozlovic, line 301: Not sure this is needed...

src/micro.c
```diff
@@ -170,88 +246,145 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
...
 
-        // Relase the endpoint's server reference from `micro_add_endpoint`.
-        _release_service(m);
+        if (unsubscribe)
+        {
+            for (ep = m->first_ep; ep != NULL; ep = ep->next)
+            {
+                if (err = micro_stop_endpoint(ep), err != NULL)
+                {
+                    err = microError_Wrapf(err, "failed to stop service '%s', stopping endpoint '%s'", m->cfg->Name, ep->config->Name);
```
kozlovic, line 326: In most of the new code in this PR you did not use `'%s'` but simply `%s`. You may want to be consistent. I think code here makes it a bit more clear).

src/micro.c
```diff
@@ -170,88 +246,145 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
...
     }
 
+    if (release)
+        m->refs--;
+
+    refs = m->refs;
+    numEndpoints = m->numEndpoints;
+    _unlock_service(m);
+
+    if ((refs == 0) && (numEndpoints == 0))
```
kozlovic, line 341: Question: would it be symptomatic of a bug if we have `refs == 0` and `numEndpoints != 0`?

src/micro.c
```diff
@@ -170,88 +246,145 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
...
+
+    natsMutex_Lock(nc->servicesMu);
+    for (int i = 0; i < nc->numServices; i++)
     {
-        natsLib_stopServiceCallbacks(m);
-        m->stopped = true;
-        doneHandler = m->cfg->DoneHandler;
+        if (nc->services[i] != m)
+            continue;
+
```
kozlovic, line 293: I have modified this file and when saving, I see some modifications due to spaces, etc.. You should look at VSCode `Settings/Text Editor/Trim Trailing Whitespaces` and make sure the box is checked.

src/micro.c
levb: I did, in fact I was VERY inclined to make them "1st class citizens" in `natsConnection`, but not for this PR...  This all really came out from trying to fix a single race condition on that global services table I had before, wrong locking order in 1 place.

OT: In light of the `orbit` movement I've been contemplating adding simple "internal use" extensibility to both Connection and Subscription; for data (like a hashmap) and callbacks (say, with a `bool lock` parameter). My initial microservices implementation was naive, it is still marked experimental, and I would really love to externalize and re-factor it to orbit before it is final.

src/micro.c
levb: It is not important anywhere in the APIs, the tests depend on it. And 👍 will make the change, y.

src/micro.c
levb: No, because of the order of execution. As I understand it (and can prove, I think) `On(Connection)Closed` may be invoked in the midst of Sub's threads calling the `On(Sub)Complete` from their own threads. `greyhair ++`

src/micro.c
levb: See the next thread

src/micro.c
levb: I moved all "attach to connection" functionality into the func, and added a `natsConn_Lock` around + connection validation. Thanks for pointing it out. After the service is added to connection, it should unwind "normally" if it's terminated.

src/micro_endpoint.c
```diff
@@ -334,7 +279,7 @@ micro_clone_endpoint_config(microEndpointConfig **out, microEndpointConfig *cfg)
     microEndpointConfig *new_cfg = NULL;
 
     if (out == NULL)
-        return micro_ErrorInvalidArg;
+        return microError_Wrapf(err, "failed to clone endpoint config: '%s'", cfg->Name);
```
kozlovic, line 282: I checked the function, and it will be a no-op if `err` is `NULL`, which it is here. So you should use `micro_ErrorInvalidArg` here instead of `err` (which was originally the case, but I guess it is a copy/paste to get the config name included).

src/micro.c
```diff
@@ -170,88 +241,158 @@ microGroup_AddEndpoint(microGroup *g, microEndpointConfig *cfg)
...
+    if (s == NATS_OK)
     {
-        natsLib_stopServiceCallbacks(m);
-        m->stopped = true;
-        doneHandler = m->cfg->DoneHandler;
+        service->refs++; // no lock needed, called from the constructor
+        nc->services[nc->numServices] = service;
+        nc->numServices++;
     }
+    natsMutex_Unlock(nc->servicesMu);
```
kozlovic, line 287: The lock was acquired only under `if (s == NATS_OK)` after setting the callbacks, but here you unlock unconditionally. You should the `if` statement above (where you bump service's ref count) under the `if` statement where you acquire the service lock. This unlock would fall under that previous `if` statement too.

kozlovic (changes requested):

src/micro.c
levb: (hit the wrong button before) thanks for catching this!

kozlovic (approved):
LGTM

levb:
Cherry-picked into v3.9.2

PR #818, opened 2024-11-13
Added UpdatesOnly to kwWatchOptions

levb:
As requested on Slack, https://natsio.slack.com/archives/C01AWP40KGF/p1729791332242309

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -32121,6 +32121,7 @@ void test_KeyValueWatch(void)
     natsThread_Join(t);
     natsThread_Destroy(t);
     kvWatcher_Destroy(w);
+    w = NULL;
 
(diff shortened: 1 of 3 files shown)


test/test.c
```diff
@@ -32121,6 +32121,16 @@ void test_KeyValueWatch(void)
     natsThread_Join(t);
     natsThread_Destroy(t);
     kvWatcher_Destroy(w);
+
```
kozlovic, line 32124: nit: As best practice, I would set `w` to `NULL` since you are going to recreate below. But since if it is fails to be created we fail (testCond() would exit), then there is no risk of "double free" when calling `kvWatcher_Destroy()` again...

kozlovic (approved):
LGTM

kozlovic (commented):
LGTM

Issue #819, opened 2024-11-19
Error while build with specific options


oldnick85:
### Observed behavior

Build with 
NATS_BUILD_WITH_TLS=OFF 
NATS_BUILD_STREAMING=OFF 
NATS_BUILD_USE_SODIUM=OFF
leads to:

> src/glib/glib.c:111:56: error: expected expression before ‘natsLib’
>   111 |     const unsigned int offset = (unsigned int)offsetof(natsLib, refs);

This error occurs due to the absence `#include <stddef.h>` in glib.c:

> src/glib/glib.c:18:1: note: ‘offsetof’ is defined in header ‘<stddef.h>’; did you forget to ‘#include <stddef.h>’?
>    17 | #include "../crypto.h"
>   +++ |+#include <stddef.h>
>    18 | 

### Expected behavior

Expected successful build

### Server and client version

Error detected here: [https://github.com/nats-io/nats.c/commit/adb907418a96e659ba9b51f40a97fd86247aa2e5](https://github.com/nats-io/nats.c/commit/adb907418a96e659ba9b51f40a97fd86247aa2e5)

### Host environment

Ubuntu 24.04.1 LTS
AMD Ryzen™ 9 7900 × 24
gcc (Ubuntu 13.2.0-23ubuntu4) 13.2.0
cmake version 3.28.3

### Steps to reproduce

mkdir build; cd build
cmake -DNATS_BUILD_WITH_TLS=OFF -DNATS_BUILD_STREAMING=OFF -DNATS_BUILD_USE_SODIUM=OFF -DCMAKE_BUILD_TYPE=Release ..
cmake --build .

lightwinglc:
add #include <stddef.h> in glib.c will fix it.

PR #820, opened 2024-11-19
[FIXED] build error (#819)

oldnick85:
add stddef include to fix error build with NATS_BUILD_WITH_TLS=OFF NATS_BUILD_STREAMING=OFF NATS_BUILD_USE_SODIUM=OFF

### DIFF
--- a/src/glib/glib.c
+++ b/src/glib/glib.c
@@ -11,6 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stddef.h>
+
 #include "glibp.h"
 
 #include "../util.h"


src/glib/glib.c
```diff
@@ -11,6 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stddef.h>
```
kozlovic, line 14: It got lost during PR #778. The change here looks good to me. 

Since we need this header (so far) only because of "offsetof" that is only in this C file, it could simply be included here, but I will let @levb decide if he would rather have it in "glibp.h".

kozlovic (approved):
LGTM

levb (approved):
LGTM

levb:
Cherry-picked into v3.9.2

PR #821, opened 2024-11-19
[FIXED] Examples: Add missing 'stream' and 'durable' params in usage

kozlovic:
Also clarify that `stream` parameter is actually needed.


levb (approved):
LGTM

PR #832, opened 2025-01-14
[FIXED] C++ compiler errors

mtmk:
Fixes the case where C++ compiler errors with C2079 on C style struct definitions. Also added a simple cpp file to make sure future changes won't cause the same issue.

Error happens when `natsp.h` is included in a `.cpp` file:
```
D:\src\nats.c\src\natsp.h(377): error C2079: '__jsFetch::opts' uses undefined struct 'jsOptionsPullSubscribeAsync'
```

### DIFF
--- a/examples/check_cpp.cpp
+++ b/examples/check_cpp.cpp
@@ -1,3 +1,0 @@
-// Check if the C++ compiler is working
-// against C style struct definitions.
-#include "natsp.h"
--- a/examples/check_cpp.cpp
+++ b/examples/check_cpp.cpp
@@ -1,0 +1,3 @@
+// Check if the C++ compiler is working
+// against C style struct definitions.
+#include "natsp.h"
--- a/src/nats.h
+++ b/src/nats.h
@@ -1293,6 +1293,74 @@ typedef struct jsOptionsPullSubscribeAsync
 } jsOptionsPullSubscribeAsync;
 
 /**
+ * Async pull options.
+ *
+ * Part of #jsOptions.
+ */
+typedef struct jsOptionsPublishAsync
+{
+        int64_t                 MaxPending;             ///< Maximum outstanding asynchronous publishes that can be inflight at one time.
+
+        // If jsPubAckHandler is specified, the callback will be invoked
+        // for every asynchronous published message, either as a positive
+        // result, or with the error encountered when publishing that
+        // message. If this callback is specified, ErrHandler (see below)
+        // will be ignored.
+        jsPubAckHandler         AckHandler;             ///< Callback invoked for each asynchronous published message.
+        void                    *AckHandlerClosure;     ///< Closure (or user data) passed to #jsPubAckHandler callback.
+
+        // This callback is invoked for messages published asynchronously
+        // when an error is returned by the server or if the library has
+        // timed-out waiting for an acknowledgment back from the server
+        // (if publish uses the jsPubOptions.MaxWait).
+        jsPubAckErrHandler      ErrHandler;             ///< Callback invoked when error encountered publishing a given message.
+        void                    *ErrHandlerClosure;     ///< Closure (or user data) passed to #jsPubAckErrHandler callback.
+
+        int64_t                 StallWait;              ///< Amount of time (in milliseconds) to wait in a PublishAsync call when there is MaxPending inflight messages, default is 200 ms.
+
+} jsOptionsPublishAsync;
+
+/**
+ * Advanced stream purge options
+ *
+ * * `Subject` will filter the purge request to only messages that match the subject, which can have wildcards.<br>
+ * * `Sequence` will purge up to but not including this sequence and can be combined with subject filtering.<br>
+ * * `Keep` will specify how many messages to keep and can be combined with subject filtering.<br>
+ *
+ * \note `Sequence` and `Keep` are mutually exclusive, so both can not be set at the same time.
+ */
+typedef struct jsOptionsStreamPurge
+{
+        const char      *Subject;       ///< This is the subject to match against messages for the purge command.
+        uint64_t        Sequence;       ///< Purge up to but not including sequence.
+        uint64_t        Keep;           ///< Number of messages to keep.
+
+} jsOptionsStreamPurge;
+
+/**
+ * Advance stream information retrieval options
+ */
+typedef struct jsOptionsStreamInfo
+{
+        bool            DeletedDetails;         ///< Get the list of deleted message sequences.
+        const char      *SubjectsFilter;        ///< Get the list of subjects in this stream.
+
+} jsOptionsStreamInfo;
+
+/**
+ * Advanced stream options
+ *
+ * * `Purge` for advanced purge options.
+ * * `Info` for advanced information retrieval options.
+ */
+typedef struct jsOptionsStream
+{
+        jsOptionsStreamPurge Purge;                                ///< Optional stream purge options.
+        jsOptionsStreamInfo Info;                                 ///< Optional stream information retrieval options.
+
+} jsOptionsStream;
+
+/**
  * JetStream context options.
  *
  * Initialize the object with #jsOptions_Init.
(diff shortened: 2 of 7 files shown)


examples/check_cpp.cpp
```diff
@@ -0,0 +1,3 @@
+// Check if the C++ compiler is working
+// against C style struct definitions.
+#include "natsp.h"
```
kozlovic, line 3: This file is private: the `p` is for private. No application should include this file directly. Actually, it should not even be available if the `make install` since they would have access only to the `nats.h` file (not the private version).

So I would remove this file (which by the way does not compile - see Windows GA failure) and undo the change in `examples/CMakeLists.txt`.

src/nats.h
```diff
@@ -1239,6 +1239,59 @@ typedef void (*jsFetchCompleteHandler)(natsConnection *nc, natsSubscription *sub
  */
 typedef bool (*jsFetchNextHandler)(int *messages, int64_t *maxBytes, natsSubscription *sub, void *closure);
 
+/**
+ * Async pull subscriber options.
+ *
+ * Part of #jsOptions.
+ */
+typedef struct jsOptionsPullSubscribeAsync
```
kozlovic, line 1247: I have no issue making the `jsOptionsPullSubscribeAsync` a typedef structure, instead of being "embedded" in `jsOpts`, but if the issue is only because a user tries to include `natsp.h`, then I would not consider changing anything.

kozlovic (changes requested):
There are lot of build failures in GA (some may be unrelated and a temporary issue). But regardless, I think that the changes are not necessary since no application (cpp or c) should include `natsp.h` directly.

src/nats.h
mtmk: yes I think that is the case. do you think moving to typedefs is still worth while for future proofing?

examples/check_cpp.cpp
mtmk: no you're right when the library is installed natsp.h isn't available. I guess it's only affecting the source copy builds.

Do you think having a some form of cpp file for at least making sure C++ compiler will be happy in these kinds of scenarios is a good thing? if so how would you go about adding a cpp file to the build?

src/nats.h
kozlovic: At this point no (seeing all the build failures). 

At the very least, what you could do is update the PR by removing the cpp file and changes to examples' make file. Keep the changes in nats.h. Assuming there is no build issue, this PR should still be updated because then we would need to require a new version [here](https://github.com/nats-io/nats.c/blob/a9aef130ec008dc4f8fed16b4f7386000b71b9b8/CMakeLists.txt#L265) because the header is changed.

If there is still build failures, have a new branch that is just pointing to main and push to see if CI still has build issues. If not, then let's close this PR. If there is still build issues, they may be temporary and I would feel better wait for those issues to be resolved with the main branch before trying to include this modified PR.

examples/check_cpp.cpp
kozlovic: No, because again, we should not have an app include `nastp.h`, so there is no need to have that "work". See the other comment I just posted regarding the extraction of the js option.

src/nats.h
PhilBarila: We got into the current state because we include `natsp.h` because we wrote our own watcher around the `kvWatcher`.  We did that so we could invoke a callback on the data.  This is all before I got here.  I will be looking into the NATS API around this private data structure as soon as I post this, so I haven't yet looked at the API as it exists.  Is there an API that would invoke a callback per message that we either missed, or you added after this implementation?

mtmk: ok removed the cpp and reverted cmake. build seems fine now apart from couple of flappers presumably.

kozlovic: @mtmk Then we should still update the main make file as I mentioned earlier. That being said, it could be done as part of the next release. Also, we need to make sure that this PR does not end-up in the patch release. @levb can confirm.

Also, @PhilBarila does that mean that you are using other internals? That would not be too good, but regardless, if the change in this PR is enough to "fix" your issue, then fine. 

In that case, @mtmk, the check.cpp may instead be in the test directory. It could be a subdirectory, similar to [this](https://github.com/nats-io/nats.c/tree/main/test/dylib). It would also need to be added to the tests to run like we do it [here](https://github.com/nats-io/nats.c/blob/a9aef130ec008dc4f8fed16b4f7386000b71b9b8/buildOnTravis.sh#L69).

If this is confusing, I could do that as a separate PR later. Let me know.

PhilBarila: The `kvWatcher` is the only thing we used from the `natsp.h`.  If you have an alternative to writing our own jet-stream watcher, we'll take a very serious look at it ...

kozlovic: > Is there an API that would invoke a callback per message that we either missed, or you added after this implementation?

@PhilBarila, you can create a watcher, but they are sync by nature, so you need to call their "next" API. I am not sure what you did exactly using the internal structure, so it is difficult for me to say if there is equivalent or not (if it is just to be a callback instead of a sync call, it sounds like a thread calling [kvWatcher_Next()](https://github.com/nats-io/nats.c/blob/a9aef130ec008dc4f8fed16b4f7386000b71b9b8/src/kv.c#L945) and invoking your callback with the result may be enough)

PhilBarila: And pull the data out of it with `kvEntry_Value` ...  I'll have to do more digging, but that might work.  I am only just wading in, so there's lots more to learn before we have a clean path forward to stop depending on your private data ...  But we'll get it done as soon as is practical with all the other competing priorities ...

PhilBarila:
As the one who brought this up, I appreciate your prompt attention.  Thank you very kindly.

test/check_cpp/check_cpp.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright 2017-2018 The NATS Authors
```
kozlovic, line 1: Change date to 2025.

test/check_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+if(NOT BUILD_TESTING)
+  return()
+endif()
+
```
kozlovic, line 4: There are failures on Travis for clang with sanitize. Not exactly sure why. So maybe try to build statically, so add [this](https://github.com/nats-io/nats.c/blob/a9aef130ec008dc4f8fed16b4f7386000b71b9b8/test/CMakeLists.txt#L9-L17) to this make file.

test/check_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+include_directories(${PROJECT_SOURCE_DIR}/src)
+
+if(NATS_BUILD_WITH_TLS)
+  include_directories(${OPENSSL_INCLUDE_DIR})
+endif(NATS_BUILD_WITH_TLS)
+
+# Build the test program
+add_executable(check_cpp check_cpp.cpp)
+
+target_link_libraries(check_cpp nats ${NATS_EXTRA_LIB})
```
kozlovic, line 15: Replace `nats` with `nats_static`.

src/nats.h
```diff
@@ -1249,115 +1370,9 @@ typedef struct jsOptions
...
-                 */
-                struct jsOptionsStreamInfo
-                {
-                        bool            DeletedDetails;         ///< Get the list of deleted message sequences.
-                        const char      *SubjectsFilter;        ///< Get the list of subjects in this stream.
-
-                } Info;                                 ///< Optional stream information retrieval options.
-
-        } Stream;                                       ///< Optional stream options.
+        jsOptionsPublishAsync PublishAsync;                     ///< extra options for #js_PublishAsync
```
kozlovic, line 1373: nit: If you could align `PublishAsync;`, `PullSubscribeAsync;` and `Stream` fields, that would be nice.

kozlovic (changes requested):
It fails to compile on Travis for some build runs, so I wonder if that would work if we statically link, you could give it a try.

test/check_cpp/CMakeLists.txt
mtmk: 👍 that worked - current failures seem to be flappers although both travis builds failed on `ReconnectFailsPendingRequest`

kozlovic (approved):
LGTM

levb (approved):
LGTM, ack that this is not for the patch release

Issue #823, opened 2024-11-27
Circular lock dependancy between js_maybeFetchMore, natsConn_processMsg, and _subscribeMulti


joel22b:
### Observed behavior

The program hangs because of a deadlock. Seems to be due to locking order.

**js_maybeFetchMore lock order:**
1. Executes [`nats_lockSubAndDispatcher(sub);`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/js.c#L2929) to lock the subscription
2. It then calls [`_sendPullRequest`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/js.c#L2937) which calls [`natsConnection_PublishRequest`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/js.c#L1841) which then tries to lock the connection's mutex with [`natsConn_Lock(nc);`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/pub.c#L86) which is held by `_subscribeMulti`

**natsConn_processMsg lock order:**
1. Executes [`natsMutex_Lock(nc->subsMu);`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/conn.c#L2689) to lock to connection's sub mutex: `subMu`
2. It then tries to lock the subscription using [`nats_lockRetainSubAndDispatcher(sub);`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/conn.c#L2716) which is held by `js_maybeFetchMore`

**_subscribeMulti lock order:**
1. Calls [`natsConn_subscribeImpl`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/js.c#L2707) with `lock == true` so it locks the connection's mutex with [`natsConn_Lock(nc);`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/conn.c#L3033)
2. Tries to lock the connection's sub mutex with [`natsMutex_Lock(nc->subsMu);`](https://github.com/nats-io/nats.c/blob/v3.9.1/src/conn.c#L3054) which is held by `natsConn_processMsg`

### Expected behavior

That it wouldn't deadlock in this scenario.

### Server and client version

Nats-sever: 2.10.20
Nats.c: 3.9.1
Nats: 0.1.2 (Not relevant for this)

### Host environment

OS: Rocky Linux 8.4
Arch: amd64
GCC: 13.2.0

Linked libraries:
	linux-vdso.so.1 (0x00007ffd9effe000)
[...]

PR #834, opened 2025-02-02
[FIXED] GH-823 deadlock in js_MaybeFetchMore

levb:


mtmk (approved):
LGTM this fixes the deadlock for me

note about testing: I can reproduce the issue before the fix but on my machine with 10 cores it takes about 4-8 parallel pulls to reproduce it.

src/js.c
```diff
@@ -2927,20 +2927,22 @@ js_maybeFetchMore(natsSubscription *sub, jsFetch *fetch)
...
     jsSub *jsi = sub->jsi;
     jsi->inFetch = true;
     jsi->fetchID++;
     snprintf(fetch->replySubject, sizeof(fetch->replySubject), "%.*s%" PRIu64,
              (int)strlen(sub->subject) - 1, sub->subject, // exclude the last '*'
              jsi->fetchID);
-    natsStatus s = _sendPullRequest(sub->conn, jsi->nxtMsgSubj, fetch->replySubject, &buf, &req);
+    nats_unlockSubAndDispatcher(sub);
+
+    natsStatus s = _publishPullRequest(sub->conn, jsi->nxtMsgSubj, fetch->replySubject, &buf, &req);
```
kozlovic, line 2938: Is there not a possibility that js_maybeFetchMore be called concurrently for the same `sub->jsi`? If so, then `fetch->replySubject` that is "set" above under the lock is accessed here without and may have changed in a concurrent call. Or is the `fetch` object guaranteed to be used from a single thread?

levb: @kozlovic with the exception of the initial call, all other calls were made from the sub's delivery thread, but I decided to move it into a locally malloced buffer anyway, seems cleaner and these requests are relatively infrequent.

The initial invocation was inside the (recursive) lock, I took it out of there too.

kozlovic (approved):
LGTM

PR #836, opened 2025-02-12
[FIXED] Added missing `NATS_EXTERN` on some functions.

### LINKED ISSUE #835: Should natsConnection_Reconnect() be used?
### Observed behavior

Since v3.9.0, an additional API has appeared in nats.h: natsConnection_Reconnect(). However, it is not marked as NATS_EXTERN. Is this a typo or an unintended exposure of an internal API? Should natsConnection_Reconnect() be used?

### Expected behavior

In a system relying on a 3-node, 3-replica JetStream cluster, if 2 nodes fail, the client reconnects to the surviving node. However, JetStream becomes unavailable and the entire system is unusable. Clients originally connected to the surviving node do not receive any callback to report this.

It is desired that when a client fails to access JetStream, natsConnection_Reconnect() can be manually executed to trigger the Reconnected callback. Then in this callback, check if JetStream is unavailable; if so, report connection lost.

### Server and client version

server: 2.10.24
client: 3.9.3

### Host environment

OS: Win10

### Steps to reproduce

_No response_

kozlovic:
Namely `nats_OpenWithConfig` and `natsConnection_Reconnect`.

Also normalized the `NATS_EXTERN <return type>` in one line and the function name on the next line.

Updated the C++ compile test since it was failing on my mac env, and modified `nats_StringEquals` that was triggering a warning on my Windows VM build.

Resolves #835

### DIFF
--- a/src/nats.h
+++ b/src/nats.h
@@ -1912,7 +1912,7 @@ typedef void (*stanConnectionLostHandler)(
  * @param config points to a natsClientConfig. A copy of the settings is made,
  * so the config can be freed after initializing the NATS client.
  */
-natsStatus
+NATS_EXTERN natsStatus
 nats_OpenWithConfig(natsClientConfig *config);
 
 /** \brief Initializes the library.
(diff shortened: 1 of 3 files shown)


levb (approved):
LGTM

levb:
This will not be cherry-picked into 3.9.3, so removing the label. This change will be released with 3.10.0 going out at the same time.

PR #837, opened 2025-02-12
[FIXED] Default statistics handler of microservice had unnecessary check

### LINKED ISSUE #833: logical ‘or’ of equal expressions in micro_monitoring
It looks like perhaps a clipboard or other issue has occured, as on line 88 of micro_monitoring.c there is the following test:

`if ((m == NULL) || (m == NULL))`

I assume based on context this should read

`if ((m == NULL) || (m->cfg == NULL))`

kozlovic:
The default handler was invoked with a request that was already checked to have a service associated. The handler was extracting the service (that was not needed) and had an incorrect check such as `if ((m == NULL) || (m == NULL))`.

Resolves #833

### DIFF
--- a/src/micro_monitoring.c
+++ b/src/micro_monitoring.c
@@ -81,12 +81,8 @@ static microError *
 handle_stats_default(microRequest *req)
 {
     microError *err = NULL;
-    microService *m = microRequest_GetService(req);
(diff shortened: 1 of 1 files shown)


levb (approved):
LGTM

PR #839, opened 2025-02-18
fix: double free and deref after free in js.c

wooffie:
If we failed in natsThread_Create now we do double free for oci.

I added free on oci part for this and assigns NULL to avoid it.

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -3535,6 +3535,8 @@ jsSub_resetOrderedConsumer(natsSubscription *sub, uint64_t sseq)
         if (s != NATS_OK)
         {
             NATS_FREE(oci);
+            NATS_FREE(oci->ndlv);
+            oci = NULL;
             natsSub_release(sub);
         }
     }
--- a/src/js.c
+++ b/src/js.c
@@ -3533,13 +3533,12 @@ jsSub_resetOrderedConsumer(natsSubscription *sub, uint64_t sseq)
 
         s = natsThread_Create(&oci->thread, _recreateOrderedCons, (void*) oci);
         if (s != NATS_OK)
-        {
-            NATS_FREE(oci->ndlv);
             natsSub_release(sub);
-        }
     }
-    if ((s != NATS_OK) && (oci != NULL))
+    if (s != NATS_OK)
     {
+        if (oci != NULL)
+            NATS_FREE(oci->ndlv);
         NATS_FREE(oci);
     }
     return s;
--- a/src/js.c
+++ b/src/js.c
@@ -3534,15 +3534,12 @@ jsSub_resetOrderedConsumer(natsSubscription *sub, uint64_t sseq)
         s = natsThread_Create(&oci->thread, _recreateOrderedCons, (void*) oci);
         if (s != NATS_OK)
         {
-            NATS_FREE(oci);
             NATS_FREE(oci->ndlv);
-            oci = NULL;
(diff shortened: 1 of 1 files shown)


src/js.c
```diff
@@ -3535,6 +3535,8 @@ jsSub_resetOrderedConsumer(natsSubscription *sub, uint64_t sseq)
         if (s != NATS_OK)
         {
             NATS_FREE(oci);
```
levb, line 3537: 1/5 would it be cleaner to remove line 3537 `NATS_FREE(oci)` and just let the code fall through to line 3543 that will trigger the cleanup?

levb (changes requested):
Thanks for the fix! I made a small suggestion that I think would make the change even smaller.

src/js.c
wooffie: Okay, check it now. 
Possible situations:
1. oci == NULL (calloc fails) - we don't need to clear anything
2. oci->ndlv == NULL - we should clear only oci (oci->ndlv won't be cleared cause of s not NATS_OK)
3. all allocated good but natsThread_Create fails, -> clear oci->ndlv and after clear oci

src/js.c
```diff
@@ -3534,13 +3534,12 @@ jsSub_resetOrderedConsumer(natsSubscription *sub, uint64_t sseq)
         s = natsThread_Create(&oci->thread, _recreateOrderedCons, (void*) oci);
         if (s != NATS_OK)
         {
-            NATS_FREE(oci);
+            NATS_FREE(oci->ndlv);
```
levb, line 3537: 2/5 I would leave `NATS_FREE(oci->ndlv);` exactly where it was before since it is not related to this code block. In the case of `natsThread_Create` failing, the only immediate cleanup needed is releasing the `sub`; all else will be handled by the fallthrough, no?

wooffie: Yes, I think we can do this:
```c
        if (s != NATS_OK)
        {
            natsSub_release(sub);
        }
    }
    if ((s != NATS_OK) && (oci != NULL))
    {
        if (oci->ndlv != NULL)
             NATS_FREE(oci->ndlv);
        NATS_FREE(oci);
    }
```

We need to check it if DUP_STRING failing

levb: Just drop the braces around `natsSub_release`, and no need to check for NULL when you call NATS_FREE, it's just a wrapper on `free` which is NULL-safe.

wooffie: Okay, I leave `oci` NULL check, because if oci allocation failed we can try to free NULL+0x40 (`oci->ndlv`)

src/js.c
```diff
@@ -3533,14 +3533,12 @@ jsSub_resetOrderedConsumer(natsSubscription *sub, uint64_t sseq)
 
         s = natsThread_Create(&oci->thread, _recreateOrderedCons, (void*) oci);
         if (s != NATS_OK)
-        {
-            NATS_FREE(oci);
             natsSub_release(sub);
-        }
     }
-    if ((s != NATS_OK) && (oci != NULL))
+    if (s != NATS_OK)
```
levb, line 3538: @wooffie 0/5 Why not leave 3538-3543 exactly as they were? (and minimize the change/blame history)

wooffie: I was a little confused, made it

levb (approved):
LGTM

PR #846, opened 2025-02-20
[ADDED] natsConnection_ReadLastError replaces natsConnection_GetLastE…

### LINKED ISSUE #841: Race in `AuthFailToReconnect`
### Observed behavior

The test flakes. See https://github.com/nats-io/nats.c/actions/runs/13389959825/job/37397794491?pr=840 

### Expected behavior

Should pass consistently

### Server and client version

Server 2.11 dev, C client - `main`

### Host environment

_No response_

### Steps to reproduce

_No response_

levb:
Fixes #841 by introducing a new API `natsConnection_ReadLastError` that copies the string into a user-provided buffer, deprecating the potentially-unsafe `natsConnection_GetLastError`.


src/conn.c
```diff
@@ -4030,6 +4031,43 @@ natsConnection_GetLastError(natsConnection *nc, const char **lastError)
...
+    s = nc->err;
+    if (s == NATS_OK)
+        nc->errStr[0] = '\0';
+    else if (nc->errStr[0] == '\0')
+        snprintf(nc->errStr, sizeof(nc->errStr), "%s", natsStatus_GetText(s));
+
+    if ((buf != NULL) && (n > 0))
+    {
+        size_t errLen = strlen(nc->errStr);
+        strncpy(buf, nc->errStr, n-1);
```
kozlovic, line 4053: `strncpy` I believe will add with 0 if the source length is lower than the `n-1` you provide. I would have probably check the min of `errLen` and `n` to know how much to copy, but take into account that we would need to add the terminating `\0`.

src/conn.c
```diff
@@ -4030,6 +4031,43 @@ natsConnection_GetLastError(natsConnection *nc, const char **lastError)
...
+        buf[n-1] = '\0';
+
+        bool truncate = ((errLen > (n-1)) && (n > 4));
```
kozlovic, line 4056: Do we really need to do that (add the `...`)? If this function replaces (for data race reasons) the `natsConnection_GetLastError()`, then we were not doing that there, so I would not do it here.

And if you decide that it is not worth it at the end, then maybe simply replace the copy with a snprintf.

src/conn.c
```diff
@@ -4030,6 +4031,43 @@ natsConnection_GetLastError(natsConnection *nc, const char **lastError)
...
+        if (truncate)
```
kozlovic, line 4057: But if you do, it would be nice to add a quick test with making sure we have a last error and then invoke with a smaller buffer and verify that this code is correct.

mtmk (approved):
LGTM

kozlovic (changes requested):

src/conn.c
levb: We don't... but since we don't give out the size of the full text, I feel like we should at least provide a visual indication of truncation when it happens. Yes, I'll add a test.

src/conn.c
levb: Oh (regardless!) this was a mistake; better style would be to pass the (ptr, n) around "as is". I'll replace it with a `memcpy` since the lengths are known.

test/test.c
```diff
@@ -35085,6 +35072,44 @@ void test_MicroAsyncErrorHandlerMaxPendingBytes(void)
...
+    test("Set last error, code and test: ");
+    natsConn_Lock(nc);
+    strncpy(nc->errStr, "test error: illegal state", sizeof(nc->errStr));
+    natsConn_Unlock(nc);
+
+    test("Verify the custom error text: ");
+    s = natsConnection_ReadLastError(nc, buf, sizeof(buf));
+    testCond((s == NATS_ILLEGAL_STATE) && (strcmp(buf, "test error: illegal state") == 0));
+
+    natsConnection_Destroy(nc);
```
kozlovic, line 35110: You need to destroy the options too.

kozlovic (changes requested):

kozlovic (approved):
LGTM

PR #848, opened 2025-02-21
Fix possible null deref in pub

wooffie:
1. remove reply construct because it is done by natsConn_publish
https://github.com/nats-io/nats.c/blob/4f993dccf03e964279d8602234c8e4d2c1afe2cb/src/pub.c#L82

2. we have check msg != NULL and i just move it to natsConn_publish
 https://github.com/nats-io/nats.c/blob/4f993dccf03e964279d8602234c8e4d2c1afe2cb/src/pub.c#L282

### DIFF
--- a/src/pub.c
+++ b/src/pub.c
@@ -279,10 +279,12 @@ natsConnection_PublishString(natsConnection *nc, const char *subj,
 natsStatus
 natsConnection_PublishMsg(natsConnection *nc, natsMsg *msg)
 {
-    const char *reply = (msg != NULL ? msg->reply : NULL);
     natsStatus s;
 
-    s = _publishMsg(nc, msg, reply);
+    if (nc == NULL || msg == NULL)
+        return nats_setDefaultError(NATS_INVALID_ARG);
+
+    s = _publishMsg(nc, msg, NULL);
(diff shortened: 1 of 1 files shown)


src/pub.c
```diff
@@ -279,10 +279,9 @@ natsConnection_PublishString(natsConnection *nc, const char *subj,
 natsStatus
 natsConnection_PublishMsg(natsConnection *nc, natsMsg *msg)
 {
-    const char *reply = (msg != NULL ? msg->reply : NULL);
```
kozlovic, line 282: We need to rewrite this function and if `msg == NULL` return an invalid arg. The check of `nc` is done in the lower function, but could easily be added into this one.

As for removing the reply subject, of course we need it, but yes, the called function will take it from the provided message structure if not provided as a parameter, so I guess this change is ok.

kozlovic (changes requested):

kozlovic (approved):
LGTM

src/pub.c
kozlovic: No, I meant what the contributor did, that is, check for `msg == NULL` as an invalid arg and return if that is the case. He did add the check for `nc == NULL` (although that is done in the internal function so that was not entirely necessary).

PR #849, opened 2025-02-24
fix: srvpool typo

wooffie:
Don't check ptr, check value of ptr

### DIFF
--- a/src/srvpool.c
+++ b/src/srvpool.c
@@ -335,7 +335,7 @@ natsSrvPool_addNewURLs(natsSrvPool *pool, const natsUrl *curUrl, char **urls, in
             s = _addURLToPool(pool, url, true, tlsName);
         }
         natsStrHashIter_Done(&iter);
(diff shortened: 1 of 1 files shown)


src/srvpool.c
```diff
@@ -335,7 +335,7 @@ natsSrvPool_addNewURLs(natsSrvPool *pool, const natsUrl *curUrl, char **urls, in
             s = _addURLToPool(pool, url, true, tlsName);
         }
         natsStrHashIter_Done(&iter);
-        if ((s == NATS_OK) && added && pool->randomize)
+        if ((s == NATS_OK) && *added && pool->randomize)
```
kozlovic, line 338: Good catch, but that means that our tests were not catching that. I believe we have pool randomize tests, but I guess not one that ensured that the pool was not randomized when getting an URL we already had.

kozlovic (approved):
LGTM

levb (approved):
LGTM

PR #855, opened 2025-02-28
Updated version to v3.11-beta

levb:


CMakeLists.txt
```diff
@@ -259,11 +259,11 @@ endif(NATS_BUILD_WITH_TLS)
 # Versionning and Doc
 
 set(NATS_VERSION_MAJOR  3)
-set(NATS_VERSION_MINOR  10)
+set(NATS_VERSION_MINOR  11)
 set(NATS_VERSION_PATCH  0)
 set(NATS_VERSION_SUFFIX "-beta")
 
-set(NATS_VERSION_REQUIRED_NUMBER 0x030900)
+set(NATS_VERSION_REQUIRED_NUMBER 0x030B00)
```
kozlovic, line 266: Just FYI, this number really needs to change only when there would be an "incompatibility" between versions. So until we modify/change some of the structure/functions in nats.h, there is technically no need to bump the version required number.

kozlovic (approved):

CMakeLists.txt
levb: Right, sorry my mistake. I changed it back to 0A, and will update the release doc to explicitly say that it should change in the PRs that modify the API structs.

Issue #857, opened 2025-03-04
Need to define NATS_HAS_TLS with latest 3.10.0


thierryba:
### Observed behavior

With the latest 3.10, I need to define NATS_HAS_TLS otherwise my program does not even build.

### Expected behavior

it should be part of the config or clearly documented.

### Server and client version

3.10.0 nats.c. The problem is at compile time.

### Host environment

macOS 15, Xcode 16.2, cmake

### Steps to reproduce

build nats with TLS support
try to compile a program that include nats/nats.h

mtmk:
thanks for the report @thierryba. where are you building it? Trying it on ubuntu I don't get any issues:
```
$ wget https://github.com/nats-io/nats.c/archive/refs/tags/v3.10.0.tar.gz
$ tar zxvf v3.10.0.tar.gz
$ cd nats.c-3.10.0
$ cmake -S . -B build -DNATS_BUILD_STREAMING=NO
$ cmake --build build
[  1%] Building C object src/CMakeFiles/nats.dir/asynccb.c.o
...
[100%] Building CXX object test/check_cpp/CMakeFiles/check_cpp.dir/check_cpp.cpp.o
[100%] Linking CXX executable ../../bin/check_cpp
[100%] Built target check_cpp
$
```

levb:
@thierryba Can you please provide the host environment details? And to be clear, you are trying to build without the use of TLS?

thierryba:
@levb in fact I am building with TLS support (I have openssl). But that macro is defined nowhere. So I need to define it myself tone able to build.

mtmk:
https://github.com/nats-io/nats.c/blob/61c051f3f2dece1b8207a5baad0f780b6b69309e/CMakeLists.txt#L247

kozlovic:
@thierryba We use CMake to build the project. If you don't, it's fine, but you are responsible to add all required build/link commands to your build process.

thierryba:
I do use cmake too. That's my point. But nothing in the build adding the definition mentioned above. So I have added  add_definitions(-DNATS_HAS_TLS)  to my own cmakelists.txt in order for my source to build while including nats.h.
And that's the part that is broken if ou ask me.

kozlovic:
[...]

PR #858, opened 2025-03-06
[FIXED] SSL callback

mtmk:
Remove openssl dependency in nats.h.

* one of the proposed solutions for #857
* introduces a breaking change to recently introduced SSL callback option


src/conn.c
```diff
@@ -736,7 +736,12 @@ _makeTLSConn(natsConnection *nc)
                     s = nats_setError(NATS_SSL_ERROR, "unable to set expected hostname '%s'", nc->tlsName);
             }
             if (s == NATS_OK)
-                SSL_set_verify(ssl, SSL_VERIFY_PEER, nc->opts->sslCtx->callback != NULL ? nc->opts->sslCtx->callback : _collectSSLErr);
+            {
+                if (nc->opts->sslCtx->callback != NULL)
+                    nc->opts->sslCtx->callback(ssl);
```
kozlovic, line 741: nit: I guess it should be `(void*)ssl` to comply with the callback definition.

kozlovic (approved):
Except for a nit, LGTM. But I don't remember a case where we had to remove an API, but deprecating the existing one is not possible since it uses an SSL structure in the definition, so I guess we have no choice. Will let @levb decide if we go with this approach.

mtmk:
@lev @kozlovic @ckasabula
I would like to put this proposal on the table again for the next minor version:
* Removes Experimental API tag making it available for vcpkg users
* Simple change removes openssl.h dependency from nats.h
* Similar to cURL library API
* @ckasabula's original API proposal which especially creates a very convenient way of dealing with certs can still be added on top of this later (#825)

kozlovic:
@mtmk If you don't mind, I would like to close this one in favor of #908. We could add a callback equivalent of what you did here in the future, but since this PR would be a breaking change anyway, I think that the proposed PR #908 is a bit less. I have documented what the users would need to do for compilation to work with the new version, and already compiled applications would work well without any change.

PR #859, opened 2025-03-09
Use monotonic timer for RTT

mtmk:
Currently, on Windows, the RTT (round-trip time) resolution is too low because it doesn’t support sub-millisecond precision. To address this, I propose we use the `QueryPerformanceCounter` API, which offers higher resolution. Additionally, monotonic timers are preferable for time interval measurements. Unlike the real-time clock, which isn’t guaranteed to be incremental and can even move backward, monotonic timers ensure consistent and reliable results.

### DIFF
--- a/src/natstime.c
+++ b/src/natstime.c
@@ -61,20 +61,18 @@ nats_NowInNanoSeconds(void)
 int64_t
 nats_NowMonotonicInNanoSeconds(void)
 {
+    int64_t now = 0;
 #ifdef _WIN32
     LARGE_INTEGER frequency;
     LARGE_INTEGER counter;
     if (QueryPerformanceFrequency(&frequency) && QueryPerformanceCounter(&counter))
-        return (int64_t)(counter.QuadPart * 1000000000ULL / frequency.QuadPart);
-    return nats_NowInNanoSeconds();
+        now = (int64_t)(counter.QuadPart * 1000000000ULL / frequency.QuadPart);
 #elif defined CLOCK_MONOTONIC
     struct timespec ts;
     if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0)
-        return ((int64_t) ts.tv_sec) * 1000000000L + ((int64_t) ts.tv_nsec);
-    return nats_NowInNanoSeconds();
-#else
-    return nats_NowInNanoSeconds();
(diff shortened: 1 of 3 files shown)


src/natstime.c
```diff
@@ -58,6 +58,33 @@ nats_NowInNanoSeconds(void)
...
 }
 
+int64_t
+nats_NowMonotonicInNanoSeconds(void)
+{
+#ifdef _WIN32
+    LARGE_INTEGER frequency;
+    LARGE_INTEGER counter;
+    if (!QueryPerformanceFrequency(&frequency) || !QueryPerformanceCounter(&counter))
+        abort();
```
kozlovic, line 68: This does not seem - somehow - to match the description you have in nats.h. If not available, the time is still expressed in nanoseconds, but seems to indicate that it would still return the non monotonic "now".

So I would flip that and say that if both ok, then use next line, but if not, then return `nats_NowInNanoSeconds()`.

src/natstime.c
```diff
@@ -58,6 +58,33 @@ nats_NowInNanoSeconds(void)
...
+    return (int64_t)(counter.QuadPart * 1000000000ULL / frequency.QuadPart);
+#elif defined CLOCK_MONOTONIC
+    struct timespec ts;
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) != 0)
+        abort();
```
kozlovic, line 73: Same comments than above.

src/natstime.c
```diff
@@ -58,6 +58,33 @@ nats_NowInNanoSeconds(void)
...
+    return ((int64_t) ts.tv_sec) * 1000000000L + ((int64_t) ts.tv_nsec);
+#elif defined CLOCK_REALTIME
```
kozlovic, line 75: Anything else, I would simply return `nats_NowInNanoSeconds()` instead.

So you could rewrite the function so that we have an int64 value at the beginning set to 0, and if win32 and query API works, set the int64 value, if not windows but clock_monotonic defined (and returns 0), then set the int64. Then at the end of the function, if value is still 0, then set to `nats_NowInNanoSeconds()` and the end of function simply returns the int64 value. Does that make sense?

kozlovic (changes requested):

kozlovic (approved):
LGTM

levb (approved):

PR #860, opened 2025-03-12
[FIXED] make natsOptions_SetSSLVerificationCallback experimental

### LINKED ISSUE #857: Need to define NATS_HAS_TLS with latest 3.10.0

levb:
Fixes #857. An alternative to @mtmk's proposal.

This new API introduced in #826 created a build issue, as @ckasabula had warned.

I propose that for now we mark it as `EXPERIMENTAL` and hide behind a `NATS_WITH_EXPERIMENTAL` env. variable as this PR implements. This will fix the immediate widespread build issue.

Then I see 2 paths to fix this in a better way.
A) Go back to what @ckasabula originally proposed (+apologies!!!)
B) Leave the code "as is", but replace `NATS_HAS_TLS` with `NATS_NO_TLS ` since I believe it is purely an internal variable. That would alleviate the issue for all clients compiling the default TLS on. @kozlovic do you think it'd work?

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -4,6 +4,9 @@ on:
       arch:
         type: string
         default: "64"
+      benchmark:
+        type: string
+        default: "OFF"
       compiler:
         type: string
         default: "gcc"
@@ -13,18 +16,12 @@ on:
       dev_mode:
         type: string
         default: "OFF"
+      experimental:
+        type: string
+        default: "OFF"
       pool_dispatch:
         type: string
         default: "NO-pool"
-      write_deadline:
-        type: string
-        default: "NO-write_deadline"
-      tls:
-        type: string
-        default: "TLS"
-      verify_host:
-        type: string
-        default: "verify_host"
       repeat:
         type: string
         default: "1"
@@ -38,6 +35,9 @@ on:
       streaming:
         type: string
         default: "ON"
+      tls:
+        type: string
+        default: "TLS"
       type:
         type: string
         description: "Debug or Release."
@@ -46,15 +46,18 @@ on:
         type: string
         description: "Ubuntu version to use, e.g. '20.04'"
         default: "latest"
-      verbose_test_output:
-        type: string
-        default: "OFF"
       verbose_make_output:
         type: string
         default: "OFF"
-      benchmark:
+      verbose_test_output:
         type: string
         default: "OFF"
+      verify_host:
+        type: string
+        default: "verify_host"
+      write_deadline:
+        type: string
+        default: "NO-write_deadline"
 
     secrets:
       CODECOV_TOKEN:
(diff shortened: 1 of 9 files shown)


.github/workflows/build-test.yml
```diff
@@ -4,6 +4,9 @@ on:
       arch:
         type: string
         default: "64"
+      benchmark:
```
levb, line 7: Sorry I couldn't stand it and alphabetized the options in this PR. Unrelated.

mtmk:
* original PR #825
* merged PR #826

src/nats.h
```diff
@@ -27,13 +27,16 @@ extern "C" {
 #include "status.h"
 #include "version.h"
 
-#if defined(NATS_HAS_TLS)
+#ifdef NATS_WITH_EXPERIMENTAL
```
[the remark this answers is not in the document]
levb: It solves the issue because for all clients that don't need the new API; it is not even included. It does NOT solve the `NATS_HAS_TLS` for those who need to use it, agreed.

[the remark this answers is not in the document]
levb: @kozlovic We'd fix it properly (still TBD how, see my notes above, we have 3 options I think - @mtmk's and the 2 I listed) and remove the EXPERIMENTAL in 3.11)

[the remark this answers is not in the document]
kozlovic: Remove the `EXPERIMENTAL` "tag" on that function, but not the possible use of `NATS_WITH_EXPERIMENTAL` (as a build environment variable) in the future, correct? I am just trying to see if this is all necessary/useful or - as suggested earlier - we just quickly move to a 3.11.0.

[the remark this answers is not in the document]
levb: New minor releases are a potential inconvenience for developers; this would avoid the need for a quick 3.11. That was the idea.

Also, yes, I think we should keep NATS_WITH_EXPERIMENTAL in general for the future unstable APIs, so we can isolate them out.

src/nats.h
```diff
@@ -27,13 +27,16 @@ extern "C" {
...
+
+// natsOptions_SetSSLVerificationCallback
```
kozlovic, line 32: Remove?

src/opts.c
```diff
@@ -695,17 +695,18 @@ natsOptions_SkipServerVerification(natsOptions *opts, bool skip)
     if (s == NATS_OK)
     {
         opts->sslCtx->skipVerify = skip;
-        if (skip)
-        {
-            opts->sslCtx->callback = NULL;
-        }
+#ifdef NATS_WITH_EXPERIMENTAL
+        opts->sslCtx->callback = NULL;
```
kozlovic, line 699: Why removing the check for `skip`?

kozlovic (changes requested):

src/nats.h
levb: This was in case we ever have more than a handful EXPERIMENTAL APIs, but they prob. wouldn't be grouped together anyway; will remove when I get back soon

src/opts.c
kozlovic: Yes, but silly example, but say user sets the callback first (which sets skip to false internally and sets the callback), then this API is called with skip=false, it would be a no-op, but in this case it will now set skip to false AND delete the callback...

mtmk:
(just testing it with fedora 41. should be done soon...)

mtmk (approved):
LGTM

the only complication I can see, will be with vcpkg if we want to also publish the experimental features as well. so to me this is complicating installation a little but since vcpkg and the verification features are pretty new and possibly not being used by many it should be good until 3.11.

ckasabula:
Sorry to hear about the issues this change has caused. Let me know if I can help with the resolution of this for 3.12, e.g. revive the original approach or something else.
Have you considered just reverting it entirely for now?
I agree w/ @mtmk 's [comment](https://github.com/nats-io/nats.c/pull/860#pullrequestreview-2679988620) RE: vcpkg

PR #862, opened 2025-03-31
Add support for loading CA trusted certificates from directory

kerbert101:
This adds support for loading certificates from a directory. For now, I'm patching the library during the build of our container, but it would be nice if a dedicated function were available instead.

### DIFF
--- a/src/nats.h
+++ b/src/nats.h
@@ -2528,7 +2528,7 @@ natsOptions_LoadCATrustedCertificates(natsOptions *opts, const char *fileName);
  * You can get extra information by calling #nats_GetLastError.
  *
  * @param opts the pointer to the #natsOptions object.
- * @param path the file containing the CA certificates.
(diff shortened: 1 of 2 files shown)


src/nats.h
```diff
@@ -2518,6 +2518,22 @@ natsOptions_TLSHandshakeFirst(natsOptions *opts);
...
+ *
+ * Loads the trusted CA certificates from a directory.
+ *
+ * Note that the certificates are added to a SSL context for this #natsOptions
+ * object at the time of this call, so possible errors while loading the
+ * certificates will be reported now instead of when a connection is created.
+ * You can get extra information by calling #nats_GetLastError.
+ *
+ * @param opts the pointer to the #natsOptions object.
+ * @param fileName the file containing the CA certificates.
```
mtmk, line 2531: should be path

levb (approved):
LGTM

PR #863, opened 2025-04-08
[ADDED] Per-message TTL support (nats-server v2.11 feature)

levb:
See https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-43.md and https://github.com/nats-io/nats.go/pull/1825.

Also added all of the new JS error status codes from the server.


src/msg.c
```diff
@@ -213,7 +213,10 @@ _processKeyValue(int line, natsMsg *msg, char *endPtr, char **pPtr, char **lastK
     start = ptr;
     if (*ptr == '\r')
     {
-        if ((++ptr == endPtr) || ((*ptr == '\n') && (++ptr == endPtr)))
+        if ((++ptr == endPtr) ||
```
levb, line 216: The server was (sometimes?) returning an extra CRLF; not sure...

src/status.h
```diff
@@ -268,6 +268,39 @@ typedef enum {
     JSConsumerCreateFilterSubjectMismatchErr = 10131,   ///< Consumer create request did not match filtered subject from create subject
     JSConsumerCreateDurableAndNameMismatchErr = 10132,  ///< Consumer Durable and Name have to be equal if both are provided
     JSReplicasCountCannotBeNegativeErr = 10133,         ///< Replicas count cannot be negative
+    JSConsumerReplicasShouldMatchStreamErr = 10134,     ///< Consumer config replicas must match interest retention stream's replicas
```
levb, line 271: Added all of the missing codes from the server, including the 2 needed here for testing

src/js.c
```diff
@@ -476,6 +476,15 @@ _setHeadersFromOptions(natsMsg *msg, jsPubOptions *opts)
             s = natsMsgHeader_Set(msg, jsExpectedLastSubjSeqHdr, temp);
         }
     }
+    if ((s == NATS_OK) && (opts->MsgTTL != 0))
+    {
+        if (opts->MsgTTL < 0)
+            s = nats_setError(NATS_INVALID_ARG, "option 'MsgTTL' (%" PRId64 ") cannot be negative", opts->MsgTTL);
+        if (s == NATS_OK)
```
kozlovic, line 483: Should be a `else` here instead.

src/msg.c
```diff
@@ -213,7 +213,10 @@ _processKeyValue(int line, natsMsg *msg, char *endPtr, char **pPtr, char **lastK
...
+            ((*ptr == '\n') && (++ptr == endPtr)) ||
+            ((*ptr == '\r') && (++ptr == endPtr)) ||
```
kozlovic, line 218: I am not sure this is correct. We enter this "if" statement with `*ptr == '\r'`, so the check here would mean that we are ok with `\r\r`, which I don't think we should.

src/nats.h
```diff
@@ -611,6 +612,14 @@ typedef struct jsStreamConfig {
         /// @brief Sets the limits on certain options on all consumers of the
         /// stream.
         jsStreamConsumerLimits  ConsumerLimits;
+
+        /// @brief Allow the message to be sent with a time to live (TTL) value.
+        bool                    AllowMsgTTL;
+
+        /// @brief Enables and sets a duration for adding server markers for
+        /// delete, purge and max age limits.
```
kozlovic, line 620: No mention of the unit here? Since the added `MsgTTL` is in milliseconds, what is this one?

levb:
Not sure what's going on with the Windows build yet (wget missing?); the Ubuntu 20.04 release builds are affected by the GitHub's deprecation of that Ubuntu version, I filed a separate, unrelated PR to fix it.

kozlovic (changes requested):

src/msg.c
levb: Right, I'll re-group these

src/js.c
levb: Thx!!! Cut-paste error, was returning there before.

src/nats.h
levb: @kozlovic I'm updating the comment, but there is a potential (in-)consistency issue here. Here, in `jsStreamConfig` we already have the "prior art" of `MaxAge` which is in ns (go `time.Duration`) and we have several other places in the code where `time.Duration` is represented as nanoseconds in C, without conversions.

Yet, in `jsPubOptions` we have `MaxWait` which is in ms, so I felt compelled to make `MsgTTL` use the same units, esp. considering that the minimum allowed value is 1s.

Thoughts? Recommendation?

kozlovic: I don't like the inconsistencies, but at least make sure that we are explicit in the field's comment on what the unit is. If you think millisecond is the way to go, I have no problem. It is true, that the timeouts/maxwait, etc.. are in ms.

PS: In the Go client, those fields are documented that we need a v2.11.0+ server. Should you add that too?

kozlovic (approved):
LGTM (taking into account my previous comment about the MsgTTL unit comment).

src/nats.h
levb: Added.

PR #865, opened 2025-04-08
[BUILD ONLY] removed the older compilers jobs, n/a on Ubuntu 22.04

levb:
This job now fails since these older compilers do not appear available on Ubuntu 22.04 (after https://github.com/nats-io/nats.c/pull/864)

I scanned the README and the docs and we do not appear to make any claims as to the specific compiler version support.


kozlovic (approved):
LGTM

PR #866, opened 2025-04-09
[BUILD ONLY] Fix Windows build to use server main

mtmk:
This is caused by binary only tag no being able to processed by the nats-server download system


levb (approved):
LGTM

PR #867, opened 2025-04-09
Add support for wolfSSL TLS backend

kerbert101:
This adds support for wolfSSL as the TLS backend. It uses the OpenSSL Compatibility Layer of wolfSSL (wolfSSL must be compiled with --enable-opensslextra)

More information about the compatibility layer:

https://www.wolfssl.com/documentation/manuals/wolfssl/chapter13.html


CMakeLists.txt
```diff
@@ -70,10 +71,46 @@ if(NATS_COVERAGE)
...
 
-if(NATS_BUILD_WITH_TLS)
+if(NATS_BUILD_WITH_TLS AND NOT NATS_BUILD_WITH_WOLFSSL)
   set(OPENSSL_USE_STATIC_LIBS ${NATS_BUILD_OPENSSL_STATIC_LIBS})
   find_package(OpenSSL REQUIRED)
-endif(NATS_BUILD_WITH_TLS)
+endif(NATS_BUILD_WITH_TLS AND NOT NATS_BUILD_WITH_WOLFSSL)
+
+if(NATS_BUILD_WITH_WOLFSSL)
+  if(DEFINED ENV{NATS_WOLFSSL_DIR})
```
kozlovic, line 80: All this is a bit complex, no? Look at what we do for OpenSSL: `find_package(OpenSSL REQUIRED)`. Is there no equivalent for that?

CMakeLists.txt
```diff
@@ -253,9 +290,11 @@ if(UNIX)
     endif(NATS_BUILD_NO_SPIN)
   endif(APPLE)
 endif(UNIX)
+
 add_definitions(-D${NATS_OS})
 add_definitions(-D_REENTRANT)
-if(NATS_BUILD_WITH_TLS)
+
+if(NATS_BUILD_WITH_TLS OR NATS_BUILD_WITH_WOLFSSL)
```
kozlovic, line 297: Since I don't see where you would set `NATS_BUILD_WITH_TLS` to "false" when using WolfSSL, then not sure the "OR" is needed here, but we should also question if the following settings make sense if using WolfSSL library.

src/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ if(NATS_BUILD_WITH_TLS)
   set(NATS_OPENSSL_LIBS "${OPENSSL_LIBRARIES}")
 endif(NATS_BUILD_WITH_TLS)
 
+if(NATS_BUILD_WITH_WOLFSSL)
```
kozlovic, line 13: But then just above we possibly include OpenSSL libraries, is that the right thing to do?

src/CMakeLists.txt
```diff
@@ -37,15 +41,15 @@ endif(NATS_BUILD_STREAMING)
 # --------------------------------------
 if(NATS_BUILD_LIB_SHARED)
   add_library(nats SHARED ${SOURCES} ${GLIB_SOURCES} ${PS_SOURCES} ${S_SOURCES})
-  target_link_libraries(nats ${NATS_OPENSSL_LIBS} ${NATS_EXTRA_LIB} ${NATS_PROTOBUF_LIBRARIES} ${NATS_SODIUM_LIBRARIES})
+  target_link_libraries(nats ${NATS_OPENSSL_LIBS} ${NATS_WOLFSSL_LIBRARIES} ${NATS_EXTRA_LIB} ${NATS_PROTOBUF_LIBRARIES} ${NATS_SODIUM_LIBRARIES})
```
kozlovic, line 44: If we support different SSL libraries, I think we should have a `NATS_SSL_LIBS` that would be set to either one.

src/CMakeLists.txt
```diff
@@ -37,15 +41,15 @@ endif(NATS_BUILD_STREAMING)
...
   set_target_properties(nats PROPERTIES
     VERSION ${NATS_VERSION_MAJOR}.${NATS_VERSION_MINOR}.${NATS_VERSION_PATCH}
     SOVERSION ${NATS_VERSION_MAJOR}.${NATS_VERSION_MINOR})
 endif(NATS_BUILD_LIB_SHARED)
 
 if(NATS_BUILD_LIB_STATIC)
   add_library(nats_static STATIC ${SOURCES} ${GLIB_SOURCES} ${PS_SOURCES} ${S_SOURCES})
-  target_link_libraries(nats_static ${NATS_OPENSSL_LIBS} ${NATS_PROTOBUF_LIBRARIES} ${NATS_SODIUM_LIBRARIES})
+  target_link_libraries(nats_static ${NATS_OPENSSL_LIBS} ${NATS_WOLFSSL_LIBRARIES} ${NATS_PROTOBUF_LIBRARIES} ${NATS_SODIUM_LIBRARIES})
```
kozlovic, line 52: Same as above.

src/conn.c
```diff
@@ -44,6 +44,10 @@
 #define NATS_EVENT_ACTION_ADD       (true)
 #define NATS_EVENT_ACTION_REMOVE    (false)
 
+#ifdef NATS_HAS_WOLFSSL
+typedef int (*SSL_verify_cb)(int preverify_ok, X509_STORE_CTX *x509_ctx);
```
kozlovic, line 48: Is that not defined in WolfSSL?

src/conn.c
```diff
@@ -725,7 +729,9 @@ _makeTLSConn(natsConnection *nc)
 #endif
             if (nc->tlsName != NULL)
             {
-#if defined(NATS_USE_OPENSSL_1_1)
+// the SSL_set_hostflags function is not available in the
+// wolfSSL OpenSSL compatibility layer, so use the new function instead
+#if defined(NATS_USE_OPENSSL_1_1) && !defined(NATS_HAS_WOLFSSL)
```
kozlovic, line 734: Related to the comment in the main CMakeFile.txt. If `NATS_USE_OPENSSL_1_1` does not make sense in the context of WolfSSL, then it should not have been set in the make file.

kozlovic (changes requested):
Even if you were to add GitHub actions as instructed by @levb, the test would still use OpenSSL since you did not add the WolfSSL include/link in the test's CMakeFile.

If we are going to support another backend, we need to have more generic "SSL" build variables that will represent the chosen SSL library.

My other concern is that it looks like WolfSSL is GPLv2 and may not be compatible with Apache 2.0 license. But I don't know enough about it to know if it would be "safe" in this context.

kerbert101:
@levb In principle, I can fix the tests—I don't expect that to be a problem. I see that the protobuf and sodium libraries are included in the `nats.c.deps` repository. I think a similar approach should be taken for the wolfSSL dependency.

@kozlovic: I agree with your comments. I reused the block for protobuf-c to find wolfSSL. It might be improved, but I haven’t fully looked into it yet.

The 1.1.1 flag should indeed probably be set to false by default for the wolfSSL implementation.

Regarding the licenses: I currently don't have enough knowledge to say whether combining GPLv2 and Apache 2.0 is possible.


It might be a good idea to first consider whether it's desirable to support multiple TLS backends.

kozlovic:
@kerbert101 Sorry for the long silence, and you have probably no interest in this anymore, but you are right that we needed to decide if we would want to support different SSL backend, and at this stage I would say no. So I am closing this PR. I am sorry for wasting your time.

PR #868, opened 2025-04-11
fix: better alloc error check

wooffie:
Hi, I think we should check result of calloc instead calling argument.


kozlovic (approved):
LGTM

levb (approved):
LGTM

levb:
thx @wooffie !!!

PR #869, opened 2025-04-17
[ADDED] Pull consumer priority groups (nats-server v.2.11 feature)

levb:
See https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-42.md and https://github.com/nats-io/nats.go/pull/1826


src/js.c
```diff
@@ -1816,6 +1817,17 @@ js_checkFetchedMsg(natsSubscription *sub, natsMsg *msg, uint64_t fetchID, bool c
     if (strncmp(val, HDR_STATUS_NO_RESP_503, HDR_STATUS_LEN) == 0)
         return NATS_NO_RESPONDERS;
 
+    // Pull consumer pin ID mismatch
+    if (strncmp(val, HDR_STATUS_PIN_ID_MISMATCH, HDR_STATUS_LEN) == 0)
+        return NATS_PIN_ID_MISMATCH;
+
+    if (strncmp(val, HDR_STATUS_BAD_REQUEST, HDR_STATUS_LEN) == 0)
```
levb, line 1824: `HDR_STATUS_BAD_REQUEST` now returns `NATS_INVALID_ARG`, before it was defaulting to `NATS_ERR`

src/jsm.c
```diff
@@ -3602,6 +3628,60 @@ js_PauseConsumer(jsConsumerPauseResponse **new_cpr, jsCtx *js,
...
+                          js_lenWithoutTrailingDot(o.Prefix), o.Prefix,
+                          stream, consumer) < 0 )
+        {
+            s = nats_setDefaultError(NATS_NO_MEMORY);
+        }
+        if (freePfx)
+            NATS_FREE((char*) o.Prefix);
+    }
+
+    snprintf(jsonBuf, sizeof(jsonBuf), "{\"group\":\"%s\"}", group);
```
kozlovic, line 3669: Should be only if `if (s == NATS_OK)` because you could have set it to "NATS_NO_MEMORY" when creating the subject above.

kozlovic: Also, adding a little comment that `group` has already been verified to be no more than 16 characters, so `jsonBuf` set to 64 is big enough.

src/nats.h
```diff
@@ -1025,6 +1043,17 @@ typedef struct jsSequenceInfo
 
 } jsSequenceInfo;
 
+/**
+ * Describes the configuration of priority groups in a pull consumer.
+ */
+
```
kozlovic, line 1049: New line not needed here...

src/jsm.c
```diff
@@ -3206,6 +3212,26 @@ js_unmarshalConsumerInfo(nats_JSON *json, jsConsumerInfo **new_ci)
...
+
+    IFOK(s, nats_JSONGetArrayObject(json, "priority_groups", &priorityGroups, &priorityGroupsLen));
+    if ((s == NATS_OK) && (priorityGroups != NULL))
+    {
+        ci->PriorityGroups = (jsPriorityGroupState*) NATS_CALLOC(priorityGroupsLen, sizeof(jsPriorityGroupState));
+        if (ci->PriorityGroups == NULL)
+            s = nats_setDefaultError(NATS_NO_MEMORY);
+        else
+            ci->PriorityGroupsLen = priorityGroupsLen;
+
```
kozlovic, line 3224: There is some tab/space issue here (from a git diff from local pull of your branch). I think you may have lost the setting in VSCode that ensures that tab/spaces are removed if the line would otherwise be empty and there are also other parts in this PR where trailing spaces are not removed on file save.

src/nats.h
```diff
@@ -1047,6 +1076,8 @@ typedef struct jsConsumerInfo
         bool                    PushBound;
         bool                    Paused;
         int64_t                 PauseRemaining;        ///< Remaining time in nanoseconds.
+        jsPriorityGroupState    *PriorityGroups;        ///< Priority groups for the (pull) consumer.
```
kozlovic, line 1079: nit: the comments are moved further than the one above.

src/nats.h
```diff
@@ -6777,6 +6835,28 @@ js_PullSubscribeAsync(natsSubscription **newsub, jsCtx *js, const char *subject,
 NATS_EXTERN natsStatus
 natsSubscription_FetchRequest(natsMsgList *list, natsSubscription *sub, jsFetchRequest *request);
 
+/** \brief Un-pins a consumer priority group from a specific subscription.
+ *
+ * Un-pins a consumer priority group from a specific subscription. This applies
+ * exclusively to js_PullSubscribeAsync subscriptions, and causes the consumer
```
kozlovic, line 6841: Should it be `#js_PullSubscribeAsync`?

src/util.c
```diff
@@ -2447,6 +2447,49 @@ bool nats_IsSubjectValid(const char *subject, bool wcAllowed)
...
+    0x03  // 0x78–0x7F: 'x', 'y', 'z'
+};
+
+// Validates limited_term, see
+// https://github.com/nats-io/nats-architecture-and-design/blob/8693d2263e137e23748e3df583fd58247e4e48b7/adr/ADR-6.md?plain=1#L48
+natsStatus
+nats_validateLimitedTerm(const char *name, const char *term)
+{
+    int i = 0;
+    int len = (int) strlen(term);
```
kozlovic, line 2477: As it is in this PR, one could call this function with `term` being `NULL`, which would cause this to abort. So I would replace the line below with:
```
if (nats_IsStringEmpty(term))
    return ...

len = (int) strlen(term);
```
here we know that `len` will be > 0.

src/util.c
```diff
@@ -2581,54 +2624,58 @@ nats_freeMetadata(natsMetadata *md)
...
+natsStatus
+nats_marshalStringArray(natsBuffer *buf, bool comma, const char *fieldName, const char **values, int len)
+{
+    natsStatus s = NATS_OK;
+    int  i;
+    const char *sep = (comma ? "," : "");
+
+    if (!nats_IsStringEmpty(fieldName))
+    {
+        s = natsBuf_Append(buf, sep, -1);
```
kozlovic, line 2652: I checked the code of `natsBuf_Append` and I guess that works with `sep` being `""`, but we could "optimize" `natsBuf_Append` and return early when we evaluate `dataLen` and `dataLen == 0`.

src/util.c
```diff
@@ -2581,54 +2624,58 @@ nats_freeMetadata(natsMetadata *md)
...
-    if (s != NATS_OK)
-    {
-        natsBuf_Cleanup(&buf);
-        return s;
+        IFOK(s, natsBuf_AppendByte(buf, '"'));
     }
 
-    *out = natsBuf_Data(&buf);
+    IFOK(s, natsBuf_AppendByte(buf, ']'));
     return NATS_OK;
```
kozlovic, line 2679: No, you would have to `return NATS_UPDATE_ERR_STACK(s)`.

src/jsm.c
```diff
@@ -2940,24 +2940,20 @@ _marshalConsumerCreateReq(natsBuffer **new_buf, const char *stream, jsConsumerCo
...
+    if ((s == NATS_OK) && !nats_IsStringEmpty(cfg->PriorityPolicy))
+    {
+        s = natsBuf_Append(buf, ",\"priority_policy\":\"", -1);
+        IFOK(s, natsBuf_Append(buf, cfg->PriorityPolicy, -1));
+        IFOK(s, natsBuf_AppendByte(buf, '"'));
+    }
+    if ((s == NATS_OK) && (cfg->PinnedTTL > 0))
+        s = nats_marshalLong(buf, true, "priority_timeout", cfg->PinnedTTL);
+    if ((s == NATS_OK) && (cfg->PriorityGroups != NULL) && (cfg->PriorityGroupsLen > 0))
+        nats_marshalStringArray(buf, true, "priority_groups", cfg->PriorityGroups, cfg->PriorityGroupsLen);
```
kozlovic, line 2956: Missing `s = nats_marshalStringArray...`

src/jsm.c
```diff
@@ -2940,24 +2940,20 @@ _marshalConsumerCreateReq(natsBuffer **new_buf, const char *stream, jsConsumerCo
...
-            if (i > 0)
-                s = natsBuf_AppendByte(buf, ',');
-            IFOK(s, natsBuf_AppendByte(buf, '"'));
-            IFOK(s, natsBuf_Append(buf, cfg->FilterSubjects[i], -1));
-            IFOK(s, natsBuf_AppendByte(buf, '"'));
-        }
-
-        IFOK(s, natsBuf_AppendByte(buf, ']'));
-    }
+        nats_marshalStringArray(buf, true, "filter_subjects", cfg->FilterSubjects, cfg->FilterSubjectsLen);
```
kozlovic, line 2943: Missing `s = nats_marshalStringArray...`

src/status.h
```diff
@@ -133,6 +133,9 @@ typedef enum
     NATS_LIMIT_REACHED,                 ///< Attempt to receive messages than allowed by the byte limit, for
                                         ///  instance in js_PullSubscribeAsync().
 
+    NATS_PIN_ID_MISMATCH,               ///< Pin ID sent in the request does not match the currently pinned
```
kozlovic, line 136: trailing space here...

src/util.c
```diff
@@ -2581,54 +2624,58 @@ nats_freeMetadata(natsMetadata *md)
...
 }
+
```
kozlovic, line 2681: New empty line at the end of the file.

kozlovic (changes requested):

src/jsm.c
levb: Grrh, I did re-install everything recently, sorry about this. Fixed, I checked all files in this PR.

src/util.c
```diff
@@ -2447,6 +2447,47 @@ bool nats_IsSubjectValid(const char *subject, bool wcAllowed)
...
+    0x00, // 0x18–0x1F
+    0x00, // 0x20–0x27
+    0x08, // 0x28–0x2F: '/' (0x2F)
+    0xFF, // 0x30–0x37: '0'–'7'
+    0x03, // 0x38–0x3F: '8', '9', '=' (0x3D)
+    0xFE, // 0x40–0x47: 'A'–'G'
+    0xFF, // 0x48–0x4F: 'H'–'O'
+    0xFF, // 0x50–0x57: 'P'–'W'
+    0x03, // 0x58–0x5F: 'X', 'Y', 'Z', '_' (0x5F), '-' (0x2D)
+    0x00, // 0x60–0x67: '`' (not allowed), 'a'–'g'
```
mtmk, line 2465: should be fe?

src/util.c
```diff
@@ -2447,6 +2447,47 @@ bool nats_IsSubjectValid(const char *subject, bool wcAllowed)
...
+};
+
+// Validates limited_term, see
+// https://github.com/nats-io/nats-architecture-and-design/blob/8693d2263e137e23748e3df583fd58247e4e48b7/adr/ADR-6.md?plain=1#L48
+natsStatus
+nats_validateLimitedTerm(const char *name, const char *term)
+{
+
+    if (nats_IsStringEmpty(term))
+        return nats_setError(NATS_INVALID_ARG, "%s must not be be empty", name);
```
mtmk, line 2478: one too many 'be'

src/util.c
```diff
@@ -2473,13 +2473,11 @@ static const uint8_t _limited_term_allowed_bitmap[16] = {
...
-    int len = (int) strlen(term);
 
-    if (len == 0)
+    if (nats_IsStringEmpty(term))
         return nats_setError(NATS_INVALID_ARG, "%s must not be be empty", name);
 
-    for (; term[i] != '\0'; ++i) {
+    for (int i=0; term[i] != '\0'; ++i) {
         if (i >= 16)
             return nats_setError(NATS_INVALID_ARG, "%s must not be longer than 16 characters", name);
```
kozlovic, line 2483: I see that @mtmk posted a comment on the `_limited_term_allowed_bitmap`, we could also simply do the checks down below and make sure `c` is valid without the use of such function.

kozlovic (changes requested):
Changes I asked for looks good, so on my end I will LGTM, but there is a typo that @mtmk noticed that would need to be fixed. We could also revisit the need for the bitmap to check for valid `term`.

kozlovic (approved):
LGTM

Issue #870, opened 2025-04-17
Connection to TLS NATS via HTTP proxy


wolfkor:
### Proposed change

I need the ability to connect to a TLS encrypted NATS bus via a HTTP proxy using the C client.

### Use case

-

### Contribution

-

PR #871, opened 2025-04-19
added proxyConnectCb (#870)

wolfkor:
add proxyConnectCb for TLS connection via proxy (#870)


mtmk:
for context @wolfkor helped us adding this feature to .NET client https://github.com/nats-io/nats.net/pull/826 I'm assuming this is to implement the same here. I think the idea here has lots of potential even though the main use case now is to punch through (probably corporate) proxies. It sounds good to me in general. I shall leave it to @levb and @kozlovic if it's something they want to consider.

src/comsock.c
```diff
@@ -170,139 +171,150 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
...
-        memset(&hints,0,sizeof(hints));
-        hints.ai_socktype = SOCK_STREAM;
+        // Invoke the proxy connect callback.
+        ctx->fd = ctx->proxyConnectCb(host, port);
+        if (ctx->fd == NATS_SOCK_INVALID)
+	        s = nats_setError(NATS_SYS_ERROR, "proxy connect socket error: %d", NATS_SOCK_GET_ERROR);
+    }
+    else
+    {
+        if ((ctx->orderIP == 46) || (ctx->orderIP == 64))
```
kozlovic, line 185: The rest is a lot of changes but I assume just to put it under the `else` without other code modifications, correct? At the same time it look like the `case` of the `switch` below has not moved. It could be your preference to be this way, but to stay with the style of the existing code, I would move them too.
At this point, I wonder (again to minimize the changes) if it would not be worth copying the check for total timeout in the "if (hasProxyConnectCb)` and even return from that if. That will leave the rest of the code formatted/indented the way it was.

src/conn.c
```diff
@@ -403,6 +403,9 @@ _createConn(natsConnection *nc)
     // Set ctx.noRandomize based on public NoRandomize option.
     nc->sockCtx.noRandomize = nc->opts->noRandomize;
 
+	// Set the proxy connect callback
```
kozlovic, line 406: Formatting (again, a tab is 4 spaces)

src/nats.h
```diff
@@ -1672,6 +1672,14 @@ typedef void (*natsMsgHandler)(
 typedef void (*natsConnectionHandler)(
         natsConnection  *nc, void *closure);
 
+/** \brief Callback used to handle connections via proxy.
+ *
+ * This callback is used to handle proxy connection. It is invoked before the
+ * main connections and return the socket that will be used to connect to the server.
```
kozlovic, line 1678: Not sure what you mean by `It is invoked before the main connections..`.

Also, you don't describe what the user should return if there is an error (unable to create the socket). I don't think `NATS_INVALID_SOCKET` is exported, so you could say return `-1`, but we could also change the signature to return `natsStatus` and have the socket returned through params: `(*natsProxyConnHandler)(natsSock **fd, char* host, int port);`

src/nats.h
```diff
@@ -2951,6 +2959,19 @@ natsOptions_SetMaxPendingMsgs(natsOptions *opts, int maxPending);
...
 
+/** \brief Sets the proxy connection handler.
+ *
+ * Specifies the callback to invoke for proxy connection returning the socket to use.
+ *
+ * @see natsProxyConnHandler
+ *
+ * @param opts the pointer to the #natsOptions object.
+ * @param proxyConnHandler the proxy connection handler callback.
+ * the callback. `closure` can be `NULL`.
```
kozlovic, line 2970: Remove this line. But we may want to think if it would be helpful to actually pass a closure.

kozlovic (changes requested):
Sorry for the delay. I am not certain of the intended use-case, so I would need more information about how you plan to use that. Also, we would really need to add some tests (for instance provide a callback that returns an error and verify that connection fails), and one callback that does create socket ok. If you don't know how to add a test, I could add it after merging (if we accept the PR once we resolve the issues). Thanks!

src/nats.h
wolfkor: To clarify our situation, the following should be said. We have a NATS bus configured with TLS, running on a Linux server, which can only be communicated with through a proxy (HTTPS), with DNS also routed only through the proxy. We have C, C#, and Java applications that are supposed to use this bus. Java (nats.java) worked immediately with the appropriate proxy settings, but C# and C did not. In the C# project (nats.net), I submitted a pull request to resolve the issue. After some back-and-forth, the team there decided to leave all 'proxy connection work' to the user via a SocketConnectionFactory. Based on these experiences, I tried to deliver a similar implementation in my pull request.

wolfkor: deleted

src/nats.h
wolfkor: changed text and signature

src/conn.c
wolfkor: VS is using tabs instead of blanks :-(

src/comsock.c
```diff
@@ -170,139 +171,150 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
...
     {
-        struct addrinfo hints;
-        struct addrinfo *servinfo = NULL;
-        int             count     = 0;
-        struct addrinfo *p;
-
-        memset(&hints,0,sizeof(hints));
-        hints.ai_socktype = SOCK_STREAM;
+        // Invoke the proxy connect callback.
+        ctx->fd = ctx->proxyConnectCb(host, port);
```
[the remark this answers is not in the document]
wolfkor: callback is creating the socket and using it for proxy verification. No TLS at this point

src/comsock.c
```diff
@@ -170,139 +171,150 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
...
+        if (ctx->fd == NATS_SOCK_INVALID)
+	        s = nats_setError(NATS_SYS_ERROR, "proxy connect socket error: %d", NATS_SOCK_GET_ERROR);
```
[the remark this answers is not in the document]
wolfkor: VS...

src/comsock.c
wolfkor: done it

kozlovic:
@wolfkor I am going to close this PR and I have submitted PR #897 that fixes formatting issues and the test (and some little changes). I have used your previous commits and squashed them into one and then added a commit with my changes. You are mentioned as the author of the PR. I hope this approach is ok (I think it would have been painful for both of us to have you make the formatting changes in your PR with your current editor settings). Thanks!

PR #873, opened 2025-06-02
Fix broadcasting of messages with headers with NULL or all-whitespace values

habbbe:
Fix parsing of headers with NULL or all-whitespace values when lifting

Add tests for NULL and all-whitespace header values

### DIFF
--- a/test/test.c
+++ b/test/test.c
@@ -20072,12 +20072,14 @@ void test_HeadersNotSupported(void)
 void test_HeadersBasic(void)
 {
     natsStatus          s;
-    natsConnection      *nc     = NULL;
-    natsPid             pid     = NATS_INVALID_PID;
-    natsMsg             *msg    = NULL;
-    natsMsg             *rmsg   = NULL;
-    natsSubscription    *sub    = NULL;
-    const char          *val    = NULL;
+    natsConnection      *nc      = NULL;
+    natsPid             pid      = NATS_INVALID_PID;
+    natsMsg             *msg     = NULL;
+    natsMsg             *rmsg    = NULL;
+    natsSubscription    *sub     = NULL;
+    const char          *val     = NULL;
+    int                 count    = 0;
+    const char          **values = NULL;
 
     if (!serverVersionAtLeast(2, 2, 0))
     {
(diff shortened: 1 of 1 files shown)


test/test.c
```diff
@@ -20109,6 +20109,12 @@ void test_HeadersBasic(void)
     IFOK(s, natsMsgHeader_Set(msg, "Headers", "Hello Headers!"))
     testCond(s == NATS_OK);
 
+    IFOK(s, natsMsgHeader_Add(msg, "NULL header", NULL))
+    testCond(s == NATS_OK);
+
+    IFOK(s, natsMsgHeader_Add(msg, "Whitespace header", "  "))
+    testCond(s == NATS_OK);
+
```
kozlovic, line 20117: Let's add some `test()` so that when running the test in verbose, we see what we are testing and the `PASSED` result. The name of the keys are wrong. They need to be of the form `This-Is-Correct-Name`. The last part I suggest adding is to make sure that your changes still work when adding more than one of those "bad" values to the same key, and later retrieve them with `Values()` API:
```
    test("Add NULL header: ");
    IFOK(s, natsMsgHeader_Add(msg, "Null-Header", NULL))
    testCond(s == NATS_OK);

    test("Add whitespace header: ");
    IFOK(s, natsMsgHeader_Add(msg, "Whitespace-Header", "  "))
    testCond(s == NATS_OK);

    test("Add NULL and whitespace values under same header: ");
    IFOK(s, natsMsgHeader_Add(msg, "Special-Headers", NULL))
    IFOK(s, natsMsgHeader_Add(msg, "Special-Headers", "  "))
    testCond(s == NATS_OK);
```

kozlovic (changes requested):
I was going to say that we should simply fix the fact that we allow NULL or empty values, but I checked the Go client, and those are not rejected. Even worse, on the Go client, there is no error returned when a key is not present, we just return "", so user will not really know if the key was present but empty or the value is empty. 
So I am willing to accept those changes. I have however made some comments on the tests.

habbbe:
Added suggestions. I also added an additional test for completely empty value.

kozlovic (approved):
LGTM - Thank you for your contribution!

PR #874, opened 2025-06-04
Fixed compiler issue on macOS related to assert for status text

kozlovic:
Made this a test instead. Also fixed a flapper that was due to a recent change in the NATS Server where an INFO in response of the CONNECT protocol is now sent, causing the override server version that we use in the test to be replaced with the actual server version.

Relates to #873


levb (approved):
LGTM, but I would like to offer an alternative which I think is simpler and is somewhat re-usable, see https://github.com/nats-io/nats.c/pull/875.

PR #875, opened 2025-06-09
FIXED: alternate fix for macOS static_assert issue

levb:
(looking into `KeyValueDiscardOldToNew` failure...)


levb:
@kozlovic yes, I just rebased on your other PR to pull in that fix. I verified that this compiles on my MacOS in both C99 and C11 modes, and that the check fired in both. Waiting for the rebased version to pass...

levb:
@kozlovic because I am proposing a change to your PR, so it'd be the right order? 1/5 I should've done it to begin with...

kozlovic (approved):
LGTM

PR #878, opened 2025-06-23
[FIXED] Build: Unknown type name SSL_verify_cb

### LINKED ISSUE #877: Unknown type name SSL_verify_cb
### Observed behavior

I'm trying to compile nats.c v3.10.1 in Centos7 with OpenSSL 1.0.2k-fips 26 Jan 2017 (we need to be able to run our software on some old servers). But, compilation is failing with the error below:

```
/tmp/nats.c/src/conn.c: In function ‘_makeTLSConn’:
/tmp/nats.c/src/conn.c:740:17: error: unknown type name ‘SSL_verify_cb’
                 SSL_verify_cb cb = _collectSSLErr;
```

nats.c v3.9.3 compiles successfully in the same Centos7 environment.

I guess the below code from src/conn.c (~740) should be updated to have the use of the `cb` variable of type `SSL_verify_cb` within the `#ifdef NATS_WITH_EXPERIMENTAL` block. I see other variables of type `SSL_verify_cb` are defined within `#ifdef NATS_WITH_EXPERIMENTAL` blocks.

```
            if (s == NATS_OK)
            {
                SSL_verify_cb cb = _collectSSLErr;
#ifdef NATS_WITH_EXPERIMENTAL
                if (nc->opts->sslCtx->callback != NULL)
                    cb = nc->opts->sslCtx->callback;
#endif // NATS_WITH_EXPERIMENTAL
                SSL_set_verify(ssl, SSL_VERIFY_PEER, cb);
            }
```

### Expected behavior

The code should build

### Server and client version

nats.c v3.10.1

### Host environment

Centos7 with OpenSSL 1.0.2k-fips 26 Jan 2017

### Steps to reproduce

The above issue is reproduced with the below build procedure.

```
git clone https://github.com/nats-io/nats.c.git /tmp/nats.c
cd /tmp/nats.c/
git checkout v3.10.1
mkdir build
cd build
cmake3 .. -DNATS_BUILD_TLS_FORCE_HOST_VERIFY=OFF -DNATS_BUILD_TLS_USE_OPENSSL_1_1_API=OFF -DNATS_BUILD_OPENSSL_STATIC_LIBS=OFF -DNATS_BUILD_STREAMING=OFF
make
```

kozlovic:
We had several issues around the introducing of a verification callback. We reverted to use a `NATS_WITH_EXPERIMENTAL` build flag, but still, as part of the rewrite, there was something of the like:
```
SSL_verify_cb cb = _collectSSLErr;
```
Prior to 3.10, we would have called `SSL_set_verify(ssl, SSL_VERIFY_PEER, _collectSSLErr);` which was not causing an issue. The above would when building with OpenSSL prior to 1.1.1.

So move the `cb` initialization as a `SSL_verify_cb` only when under the experimental flag, otherwise, invoke the same code we had prior to 3.10.

Resolves #877

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -737,12 +737,14 @@ _makeTLSConn(natsConnection *nc)
             }
             if (s == NATS_OK)
             {
-                SSL_verify_cb cb = _collectSSLErr;
 #ifdef NATS_WITH_EXPERIMENTAL
+                SSL_verify_cb cb = _collectSSLErr;
                 if (nc->opts->sslCtx->callback != NULL)
                     cb = nc->opts->sslCtx->callback;
-#endif // NATS_WITH_EXPERIMENTAL
                 SSL_set_verify(ssl, SSL_VERIFY_PEER, cb);
+#else
+                SSL_set_verify(ssl, SSL_VERIFY_PEER, _collectSSLErr);
+#endif // NATS_WITH_EXPERIMENTAL
             }
         }
     }
@@ -3360,7 +3362,7 @@ _processUrlString(natsOptions *opts, const char *urls)
     serverUrls = (char**) NATS_CALLOC(count + 1, sizeof(char*));
     if (serverUrls == NULL)
         return NATS_NO_MEMORY;
-    
+
     if (s == NATS_OK)
     {
         urlsCopy = NATS_STRDUP(urls);


mtmk:
fyi i can confirm the build issue on centos7 and this fix solves it.

derekcollison (approved):
LGTM

PR #815, opened 2024-10-23
[FIXED] EventLoop: Socket now closed only after event loop done polling

kozlovic:
The socket was closed by the NATS library itself, which could cause some issue when the event loop thread would still be polling it. We now defer to the event loop adapter to make sure that the event loop library is done polling before invoking a new function that will take care of closing the socket.

I have updated the event loop test (that simulates what our adapters are doing). The mockup event loop implementation is a bit too simplistic but should be ok for now. If we have issues, we would have to make the events a linked list.

Resolves #814

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

PR #882, opened 2025-07-03
[FXIED] LibEvent: natsConnection_Close() not closing the TCP connection

### LINKED ISSUE #879: When using the Libevent library, natsConnection_Destroy() does not actually close the TCP connection
### Observed behavior

When using the NATS.C client library integrated with an external event library Libevent, I found that calling natsConnection_Destroy() does not actually close the connection. Packet capture with tcpdump shows no RST packet being sent.
After reading the source code of NATS.C and Libevent, I identified the root cause of the issue. It stemmed from a misunderstanding of how event_base_once() is used. As shown in the diagram below,

Image

Image

Image

Image

Image

This issue is easy to reproduce and occurs consistently. However, when reproducing it, it's important to add a getchar() or similar pause after calling natsConnection_Destroy(). Otherwise, if the process exits immediately, the operating system may send a RST packet, which could be mistaken as being triggered by natsConnection_Destroy() itself.

### Expected behavior

When integrating with an external event loop Libevent, natsConnection_Destroy() is expected to close the TCP connection.

### Server and client version

nats-server: v2.11.1-binary
client: 3.10.1

### Host environment

Server and Client are all running in Linux system 

### Steps to reproduce

"The files nats.c-3.10.1/examples/libevent-pub.c and nats.c-3.10.1/examples/libevent-sub.c in the source code can both be used to reproduce the issue. You can add a getchar() after natsConnection_Destroy(conn);. Before running the program, use tcpdump to capture packets. When the program reaches the getchar() call, you can observe whether an RST packet is sent.

kozlovic:
This was likely introduced in PR #815 when delaying the close of the socket based on the event loop adapter invoking a NATS C callback to indicate that it is safe to do so. The `libuv` adapter is doing the right thing, however the `libevent` was not since it was creating a one-time event based on `EV_TIMEOUT` passing the socket to close, however, in that mode the socket from the callback is always set to `-1`. We have to pass the socket to close though the `nle` object.

I also made the `natsLibevent_Detach()` function schedule an event to ensure that a `_freeCb` is now invoked *after* execution of the `_closeCb`.

Resolves #879

### DIFF
--- a/src/adapters/libevent.h
+++ b/src/adapters/libevent.h
@@ -32,6 +32,7 @@ typedef struct
     struct event        *read;
     struct event        *write;
     struct event        *keepActive;
+    natsSock            socketToClose;
 
 } natsLibeventEvents;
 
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

PR #883, opened 2025-07-04
[IMPROVED] Timer insertion

kozlovic:
We use a linked list to hold timers. They are ordered by absolute time. Inserting a timer was always looking for the spot starting at the beginning. We now check first if it should be added directly at the end of the list, which in most cases - say creating new connections that creates a timer for sending PINGs, they will likely be sequential.

Added a test that shows that for 100,000 timers, the time spent is now very small.


derekcollison (approved):
LGTM

PR #885, opened 2025-07-04
Testing: add custom inbox prefix that is very long

### LINKED ISSUE #884: Incorrect use of sizeof() in the natsConnection_RequestMsg() function
### Observed behavior

No error has occurred so far, but it could be misused in the future.

In the line natsConn_addRespInfo(&resp, nc, respInbox, sizeof(respInbox));, the sizeof() operator always returns the size of the pointer itself, not the size of the content it points to. An explicit size should be passed here instead. However, since the natsConn_addRespInfo function does not actually use this size internally, no error has occurred.

```
natsStatus
natsConnection_RequestMsg(natsMsg **replyMsg, natsConnection *nc,
                          natsMsg *m, int64_t timeout)
{
...
    char                respInboxBuf[32 + NUID_BUFFER_LEN + NATS_MAX_REQ_ID_LEN + 1]; // <inbox prefix>.<nuid>.<reqId>
    char                *respInbox = respInboxBuf;



    // If the custom inbox prefix is more than the reserved 32 characters
    // in respInboxBuf, then we need to allocate...
    if (nc->inboxPfxLen > 32) 
    {
        respInbox = NATS_MALLOC(nc->inboxPfxLen + NUID_BUFFER_LEN + NATS_MAX_REQ_ID_LEN + 1);
       ...
    }
        ===> printf("The size of respInbox:%d\n", sizeof(respInbox));
...
        s = natsConn_addRespInfo(&resp, nc, respInbox, sizeof(respInbox));
...
}
```

### Expected behavior

The size of respInbox should be equal to the actual size.

### Server and client version

[server 2.10.1](nats-server: v2.11.1-binary)
client: 3.10.1

### Host environment

_No response_

### Steps to reproduce

_No response_

kozlovic:
This is to make sure that we properly allocate the inbox with a very long prefix that is larger than the stack variable.

Resolves #884

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -1328,7 +1328,7 @@ natsConn_destroyRespPool(natsConnection *nc)
 // subject (that is set in respInbox). The respInfo object is returned.
 // Connection's lock is held on entry.
 natsStatus
-natsConn_addRespInfo(respInfo **newResp, natsConnection *nc, char *respInbox, int respInboxSize)
+natsConn_addRespInfo(respInfo **newResp, natsConnection *nc, char *respInbox)
 {
     respInfo    *resp  = NULL;
     natsStatus  s      = NATS_OK;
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

Issue #880, opened 2025-07-03
Improper sequencing of resource deallocation in the js_release() -> _freeContext() function may result in a core dump


yanyongcheng:
### Observed behavior

Running the js_pub.c example in asynchronous mode may occasionally result in a core dump. Upon analysis, the issue seems to stem from an incorrect sequence of resource cleanup in the js_release() function. 
Please refer to the comments marked with '===>'
```
void
js_release(jsCtx *js)
{
...
       ===> _freeContext(js);_
}
```
```
static void
_freeContext(jsCtx *js)
{
    natsConnection *nc = NULL;

    natsStrHash_Destroy(js->pm);
    natsSubscription_Destroy(js->rsub);
    _destroyOptions(&(js->opts));
    NATS_FREE(js->rpre);
    natsCondition_Destroy(js->cond);
    ===>natsMutex_Destroy(js->mu); // The js->mu have been destroyed first, then destroy the js->pmtmr
    ===>natsTimer_Destroy(js->pmtmr); //This will call stop callback of timer (_timeoutPubAsyncComplete)_
    nc = js->nc;
    NATS_FREE(js);

    natsConn_release(nc);
}
```

```
static void
_timeoutPubAsyncComplete(natsTimer *t, void *closure)
{
    jsCtx *js = (jsCtx*) closure;
    ===>js_release(js);
}
```

```
void
js_release(jsCtx *js)
{
...
   ===> js_lock(js); // This will attempt to get the mutex (js->mu), But this have been destroyed before. so will cause  a coredump !!!
    doFree = (--(js->refs) == 0);
    js_unlock(js);

    if (doFree)
        _freeContext(js);
}
```

### Expected behavior

When running js_PublishAsync, no core dump should occur.

### Server and client version

nats-server: v2.11.1-binary
client:3.10.1

### Host environment

server and client are all running in Linux system.

### Steps to reproduce

Just run the js-pub.c in the examples dir. However, it may not occur every time, since the core dump happens probabilistically.This makes it harder to reproduce and debug. It may be necessary to use memory checking tools like ASAN to trace the issue.

kozlovic:
The retain has to be done before creating the timer (and release in case of error creating the timer). Will fix and see if I can add a test that would have caught that.

kozlovic:
[...]

PR #886, opened 2025-07-04
[IMPROVED] JetStream: handling of publish async timeouts

kozlovic:
When publishing a JetStream message asynchronously but with a publish max wait, we would schedule an internal timeout message to our internal subscription to handle ack timeout, even if the message was actually already ack'ed. The handler for the ack would disregard that timeout message because it would not be found in the hash table but we optimize here by not scheduling if the message has already been handled in `_handleAsyncReply()`.

This was reported in #880

### DIFF
--- a/src/js.c
+++ b/src/js.c
@@ -842,19 +842,23 @@ _timeoutPubAsync(natsTimer *t, void *closure)
 
     while (((pm = js->pmHead) != NULL) && (pm->deadline <= now))
     {
-        natsMsg *m = NULL;
+        // Check if the corresponding message is still in the hashtable.
+        char *id = (pm->subject+js->rpreLen);
+        if (natsStrHash_Get(js->pm, id) != NULL)
+        {
+            natsMsg *m = NULL;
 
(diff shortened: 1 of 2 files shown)


src/js.c
```diff
@@ -842,19 +842,23 @@ _timeoutPubAsync(natsTimer *t, void *closure)
...
     {
-        natsMsg *m = NULL;
-
-        if (natsMsg_Create(&m, pm->subject, NULL, NULL, 0) != NATS_OK)
-            break;
-
-        natsMsg_setTimeout(m);
+        // Check if the corresponding message is still in the hashtable.
+        char *id = (pm->subject+js->rpreLen);
+        if (natsStrHash_Get(js->pm, id) != NULL)
```
derekcollison, line 847: Should this be == NULL?

derekcollison: Or does this mean only create the new message if we have an antry, otherwise do not create one at all?

derekcollison (approved):
LGTM

PR #889, opened 2025-07-22
[FIXED] EventLoop: Libuv may crash if connection destroyed while consuming

### LINKED ISSUE #888: Why is there no retain(nc) in the natsLibevent_Attach() function, and then a corresponding decrement in natsLibevent_Detach()?
### Observed behavior

Why is there no retain(nc) in the natsLibevent_Attach function, and then a corresponding decrement in natsLibevent_Detach()? 
Instead, the retain(nc) is done in natsConnection_ProcessReadEvent(). What's the purpose? and it doesn't feel very intuitive.

### Expected behavior

retain(nc) in the natsLibevent_Attach function, and then a corresponding decrement in natsLibevent_Detach()

### Server and client version

client:3.10.1

### Host environment

_No response_

### Steps to reproduce

_No response_

kozlovic:
If the user destroys the connection while the read event was in progress it could be possible for the library to crash trying to process a read event.

We need to immediately stop polling when the read event is removed.

Changing also the way we retain the connection: we will now retain on the very first attach, and the adapters will have to call `natsConnection_Destroy()` when the event loop is fully detached.

Resolves #888


derekcollison (approved):
LGTM

PR #892, opened 2025-07-23
[FIXED] MicroServices: Wrong marshaling of `average_processing_time`

### LINKED ISSUE #890: Micro service stats marshalls duration as a string instead of a number
### Observed behavior

From https://github.com/nats-io/nats.go/issues/1905

[`marshal_stats`](https://github.com/nats-io/nats.c/blob/665febee3d316a5d91982ba8d67a0152300c8437/src/micro_monitoring.c#L328) outputs `"average_processing_time"` as string, when the specs expect it to be a number.
[`nats_marshalDuration`](https://github.com/nats-io/nats.c/blob/665febee3d316a5d91982ba8d67a0152300c8437/src/util.c#L2380) perhaps needs to be modified to write a number value (nanoeconds) instead of converting it to string.

As a result of this problem [nast cli panics when getting info on nats.c based services](https://github.com/nats-io/nats.go/issues/1905).

### Expected behavior

`average_processing_time` is serialized as a number.

### Server and client version

Server: 2.11.6
nats cli: 0.2.3
nats.c: master 665febee3d316a5d91982ba8d67a0152300c8437

### Host environment

macOS

### Steps to reproduce

- Run a [nats.c example hello service](https://github.com/nats-io/nats.c/blob/main/examples/micro-hello.c)
- Execute nats service info hello

kozlovic:
The field was incorrectly marshaled as a string duration instead of a number (nano seconds). Also, the computation of the average time would not take into account the possibility of having zero number of requests, which would produced a NaN and an invalid number during marshal.

Added test to check statistics before any request and that the field is a proper number (not a string). Also verified that it now works well with the NATS CLI.

Resolves #890

### DIFF
--- a/src/micro.c
+++ b/src/micro.c
@@ -933,8 +933,12 @@ microService_GetStats(microServiceStats **new_stats, microService *m)
                 MICRO_CALL(err, micro_strdup((char **)&stats->Endpoints[len].QueueGroup, micro_queue_group_for_endpoint(ep)));
                 if (err == NULL)
                 {
-                    avg = (long double)ep->stats.ProcessingTimeSeconds * 1000000000.0 + (long double)ep->stats.ProcessingTimeNanoseconds;
-                    avg = avg / (long double)ep->stats.NumRequests;
+                    avg = 0.0;
+                    if (ep->stats.NumRequests >= 1)
+                    {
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #897, opened 2025-07-28
[ADDED] `natsOptions_SetProxyConnHandler` option

### LINKED ISSUE #870: Connection to TLS NATS via HTTP proxy

kozlovic:
From: @wolfkor

Based on PR #871 (it had too many formatting issues that the author had difficulty fixing. Also added fixes for the test)
    
Resolves #870


kozlovic:
@mtmk Thank you for the review. I made a new commit with some changes to the API to be more inline with the code style and add a "closure" to the callback which may be useful. I did use that in the test to check that the callback is indeed invoked, etc... So could you please have another look? (you can focus on the last commit instead of reviewing the whole PR).

kozlovic:
@mtmk Re: `closure` name. I would agree, but I have used that term in almost all callbacks/functions where I pass user provided data, so for consistency sake, I would rather use the same terminology.

PR #900, opened 2025-08-11
Add SSL certificate callback

Matus-p:
this allows to dynamically reload certificate when it is renewed


kozlovic:
@Matus-p Thank you for the PR. I am quite busy at the moment so I may not have the time to fully review/merge. I appreciate you adding the test file, which seems quite complete. I think your code is correct, but I am not sure I want to add to the experimental approach, which was a bit of a knee-jerk reaction when we introduced the other callback that broke builds because suddenly the `nats.h` header needed TLS to be able to compile.

I just wanted to let you know that we are not ignoring your PR and will review/comment when time permits.

Matus-p:
@kozlovic thank you for the first feedback.
What I need is to have nats client load client certificate each time it makes a TLS connection - this will automatically load new certificate in case it was rotated. It is already done in nats.go client ( https://github.com/nats-io/nats.go/blob/v1.44.0/nats.go#L982 )

If the dependency on TLS header in nats.h is not wanted, I think I could make a new function (lets say natsOptions_LoadCertificatesChainDynamic) that will have the same signature as natsOptions_LoadCertificatesChain but under the hood will use the SSL_CTX_set_cert_cb instead of SSL_CTX_use_certificate_chain_file / SSL_CTX_use_PrivateKey_file.

Let me know if this would be the preferred way.

kozlovic:
@Matus-p But you would still need a callback with the `SSL` signature, no? Or are you saying that you would only pass some file names?

By the way, I did enable test suite for your PR and apparently Windows does not like the test you have added. Again, I don't have the time to look at this at the moment.

Matus-p:
@kozlovic SSL headers won't be needed. I will pass only natsOptions and filenames. The cert callback would be internal to nats.c library. I may prepare a PR and you will see if it fits better than this PR or not.

i will try to have a look at that windows test to see if i can figure it out

Matus-p:
@kozlovic I have implemented a different approach, you can see it here: https://github.com/nats-io/nats.c/pull/901
thank you

kozlovic:
Closing this PR since we have an alternate approach that will add less friction.

PR #901, opened 2025-08-20
[Added] TLS: natsOptions_LoadCertificatesChainDynamic

Matus-p:
Loads the certificate chain and key from a file on every connection attempt.
This allows to dynamically reload certificate when it is renewed.


kozlovic (approved):

PR #905, opened 2025-09-05
[CHANGED] TLS/Build: Requires SSL 1.1.1+ to compile, init/cleanup removed

### LINKED ISSUE #904: TLS cleanup calls sk_SSL_COMP_free() but does not reset pointers
### Observed behavior

This means that any subsequent attempt to shut down SSL by something else in the same process (typically via OPENSSL_cleanup()) crashes due to a double free. Both NATS_HAS_TLS and NATS_USE_OPENSSL_1_1 are defined.

### Expected behavior

Either 1) do not call sk_SSL_COMP_free() at all, or 2) reset all pointers to memory freed by sk_SSL_COMP_free().

### Server and client version

Client: 3.7.0
Server: 2.10.9

### Host environment

C++ app using nats-c running on RHEL8.

### Steps to reproduce

Use nats-c with TLS enabled and call nats_Close(). Any subsequent cleanup of SSL core dumps with this partial traceback:

...
#6 0x00007ffff09e1f20 in OPENSSL_sk_pop_free () from /lib64/libcrypto.so.1.1
#7 0x00007ffff0d16359 in ssl_library_stop () from /lib64/libssl.so.1.1
#8 0x00007ffff0976e4a in OPENSSL_cleanup () from /lib64/libcrypto.so.1.1
#9 0x00007ffff4dcd26c in __run_exit_handlers () from /lib64/libc.so.6
#10 0x00007ffff4dcd3a0 in exit () from /lib64/libc.so.6
...

kozlovic:
Code was still trying to support SSL versions prior to 1.1, and some of the code for 1.1+ has been marked as deprecated and were no-ops.

This PR removes the openssl initialization and cleanup calls (that may have caused issues if application themselves try to cleanup SSL on exit), and has changed the CMakeLists.txt file to:
- Remove the `NATS_BUILD_TLS_USE_OPENSSL_1_1_API` option
- Requires OpenSSL 1.1.1 at minimum

Users that use `-DNATS_BUILD_TLS_USE_OPENSSL_1_1_API=ON` (which has been already the default for quite some time) would need to upgrade their script to avoid a warning. If this was explicitly set to `OFF`, and users use a SSL library pre 1.1.1, they may need to upgrade their SSL library.

Internal: also removed `.travis.yml` since we have been using GitHub actions.

Resolves #904


mtmk (approved):
LGTM

PR #907, opened 2025-09-10
[FIXED] TLS: Possible hang during handshake

### LINKED ISSUE #906: Unrecoverable hang during SSL handshake after connection lost
### Observed behavior

I have a NATS client running at a remote site that occasionally experiences network unreliability. During these periods of unreliability we see unstable TCP connections. Generally, the NATS client will detect the problem and reconnect. But on some rare occasions I've found that the client will hang while attempting to reconnect and will never recover until the process is restarted. I've only seen this happen around 2-3 times in the last year.

The last time this happened I captured a stack trace with eu-stack. Most of the threads are just running natsCondition_Wait waiting for messages to come in. But the nats.c reconnect thread is interesting:
```
TID 3471564:
#0  0x00007f5d080fd9ec read
#1  0x00007f5d088dd091 sock_read
#2  0x00007f5d088cd44b bread_conv
#3  0x00007f5d088d0445 bio_read_intern
#4  0x00007f5d088d05c7 BIO_read
#5  0x00007f5d08d96e9c ssl3_read_n.part.0
#6  0x00007f5d08d98e4a ssl3_read_bytes
#7  0x00007f5d08da85d0 state_machine.part.0
#8  0x00000000004a24ae _makeTLSConn
#9  0x00000000004a4978 _processConnInit
#10 0x00000000004a8ed9 _doReconnect
#11 0x00000000004de0d0 _threadStart
#12 0x00007f5d08089c02 start_thread
#13 0x00007f5d0810ec40 __clone3
```

The thread never advances out of this state. I think this is what is causing the hang. It appears that several function calls were optimized out so I couldn't perfectly trace the code but think that this is happening.
1. The TCP connection drops, nats.c detects that and being reconnecting.
2. The new TCP connection is established in _doReconnect.
3. Really poor network conditions cause the established connection to be dropped almost immediately.
4. nats.c moves into _processConnInit and _makeTLSConn.
5. _makeTLSConn sets the connection to blocking mode.
6. The OpenSSL handshake begins and `read()` is called.

Here we are stuck permanently. No data is ever read because the connection has been lost. Since we are in blocking mode we cannot timeout the `read()`. It does not appear that TCP keepalives are enabled so the kernel can't tell us that the connection has dropped either.

### Expected behavior

_makeTLSConn can return with an error if the TCP connection has been lost when no data has been read.

### Server and client version

nats.c 3.10.1
nats-server 2.9.19
nats-streaming-server 0.25.5

### Host environment

RHEL 9.4 x86_64

### Steps to reproduce

I have not found a reliable way to reproduce this issue.

kozlovic:
The reason was that we switched the connection to blocking mode for the SSL handshake duration. In some cases, this could lead to the connection being blocked in the handshake for too long (longer than the provided connection timeout).

We now don't switch to blocking mode and check if `SSL_do_handshake` wants a read or write and wait for such event to be available.

Resolves #906


src/conn.c
```diff
@@ -749,23 +745,40 @@ _makeTLSConn(natsConnection *nc)
...
-            s = nats_setError(NATS_SSL_ERROR,
-                              "SSL handshake error: %s",
-                              (nc->errStr[0] != '\0' ? nc->errStr : NATS_SSL_ERR_REASON_STRING));
+            int sslErr = SSL_get_error(ssl, hsErr);
+            if ((sslErr == SSL_ERROR_WANT_READ) || (sslErr == SSL_ERROR_WANT_WRITE))
+            {
+                int waitMode = (sslErr == SSL_ERROR_WANT_READ ? WAIT_FOR_READ : WAIT_FOR_WRITE);
+
+                if ((s = natsSock_WaitReady(waitMode, &(nc->sockCtx))) == NATS_OK)
+                    goto DO_HANDSHAKE;
```
mtmk, line 762: (optional) although infinite loop seems unlikely, a loop counter would be a nice to have, perhaps as a follow up pr later on if you agree as well.

mtmk (approved):
LGTM (with one minor comment for future)

src/conn.c
mtmk: i was thinking odd cases like deadline being null or something but it is fairly tight. it's fine as it is. no need for a follow up. thanks for the explanation.

PR #914, opened 2025-09-17
[ADDED] Option `natsOptions_AllowConcurrentTLSHandshakes`

### LINKED ISSUE #899: With TLS enabled, connection creation in multi-threaded scenarios is significantly slower
### Observed behavior

When I tried to create TLS connections using multiple threads, I found that no matter how many threads I used, the connection speed didn’t improve.

After reviewing the source code, I found that in the _makeTLSConn() function, nc->opts->sslCtx is globally shared—meaning all nc instances use the same sslCtx. As a result, sslCtx->lock becomes a global lock rather than a connection-level lock. Moreover, the critical section protected by this lock is too broad, including time-consuming operations like SSL_do_handshake(), which also involves network transmission delays.

Do you have any suggestions for optimizing this issue? I'd like to test them locally first.

Image

Image

### Expected behavior

TLS connection setup is relatively fast—it may be slightly slower than plaintext mode, but it shouldn't be this slow.

### Server and client version

client: 3.10.1

### Host environment

_No response_

### Steps to reproduce

_No response_

kozlovic:
We used to have a ssl context lock held when setting up SSL when connecting, which included the SSL handshake.

Introduce an option that allows the user to explicitly allow the connections created with the same `natsOptions` object to perform the SSL handshake concurrently. If any SSL callback is configured, the user must ensure that their implementation is thread-safe, or not use the new option.

Resolves #899


src/conn.c
```diff
@@ -673,11 +673,21 @@ _makeTLSConn(natsConnection *nc)
...
     nc->errStr[0] = '\0';
 
+    // We will lock by default, unless the option `tlsConcurrentHandshakes` allows
+    // us to perform the handshake concurrently. Regardless, we will perform the
+    // handshake under lock for the very first one for this context. This is because
+    // otherwise we get thread sanitizer reports (looks like there is some lazy
+    // initialization happening in the OpenSSL library for the very first handshake).
+
     natsMutex_Lock(nc->opts->sslCtx->lock);
+    useLock = !nc->opts->tlsConcurrentHandshakes || nc->opts->sslCtx->firstHandshake;
```
mtmk, line 688: minor suggestion: would it be possible to check for `tlsConcurrentHandshakes` (since it's read-only) first, before taking the lock to avoid contention altogether?

mtmk (approved):
LGTM with a minor suggestion

PR #915, opened 2025-09-18
[FIXED] Connection drain could cause missed reply and/or 100ms delay

### LINKED ISSUE #911: New subscription first outgoing message waiting (warmup)
### Observed behavior

I'm observing some weird approximately 100ms waiting while publishing first response for incoming message. 

Total runtime for steps lower is 103-105ms, but if I remove step 6 (case without sending response back but no connection/subscription reestablishing), 100ms waiting goes away and time decreases to 3-5ms stable.
If I repeat step 6 exactly 5 times - no additional 100ms waiting (as well as more 3-5ms per run which is strange but not the case for me now).

Is this normal? Exact 100ms looks like a hard coded timeout to me for some OS reasons.

### Expected behavior

No 100ms waiting.

### Server and client version

nats-server: v2.11.6
nats.c version: 3.9.3

### Host environment

Ubuntu 22.04, amd64, native.

### Steps to reproduce

Application is CPP, gtest based. 
1. Initialize nats.c library;
2. Open new connection to NATS server:
2.1. IP, Port, No auth;
2.2. Set send ASAP = true;
3. Create new subscription;
4. Send request (natsConnection_RequestMsg, 1s timeout, payload is "hello, subscription!") to subscription (3);
5. Handle request in callback;
6. Publish response (natsConnection_PublishMsg, payload is "hello, requester!") through same connection (by connection pointer from callback);
7. Receive sent response to request in (4);

kozlovic:
The more important issue is that we should have drained the internal `respMux` subscription after all user subscriptions have been drained so that we don't miss replies that would be sent by the user subscriptions callbacks.

Draining the internal `respMux` subscription would often cause the check for number of subscriptions to be positive, which then would lead to a wait of 100ms before checking again that the number would be down to 0.

This PR addresses both issue: make sure that the internal subscription is drained last, and use condition variable to notify when subscriptions are removed and the number is down to the expected number.

Resolves #911


src/conn.c
```diff
@@ -3705,11 +3727,13 @@ _iterateSubsAndInvokeFunc(natsStatus callerSts, natsConnection *nc, subIterFunc
...
 {
-    natsStatus s;
+    natsStatus  s;
+    int64_t     sid;
 
     natsSub_Lock(sub);
-    s = natsConn_enqueueUnsubProto(nc, sub->sid);
+    sid = sub->sid;
     natsSub_Unlock(sub);
+    s = natsConn_enqueueUnsubProto(nc, sid);
```
mtmk, line 3736: observation: are we taking the risk of unsubbing multiple times since it would be a noop on the server side?

mtmk (approved):
LGTM

src/conn.c
kozlovic: What makes you think we would be UNSUB'ing multiple times? The change here is probably not necessary, but felt like I did not want to have the sub's lock while in the function the enqueue the protocol. The caller has both the connection lock and subMu lock, so there is no risk that this subscription is UNSUB'ed twice here.

PR #934, opened 2025-11-20
[FIXED] ObjectStore: Wrong encoding of meta subject

### LINKED ISSUE #933: unable to read from objectStore
### Observed behavior

I'm unable to get an object from objectStore.

### Expected behavior

Either a file is created or a string is available

### Server and client version

Server: v2.12.2
nats.c : v3.11.0

### Host environment

_No response_

### Steps to reproduce

I have no ideas anymore, what i'm doing wrong. The simple test code is
```
objStore_Watch(&watch, obs, nullptr);
objStoreInfo *info = nullptr;
status = objStoreWatcher_Next(&info, watch, 100);

printf("New object found: %s\n", info->Meta.Name);
printf("Info: \n");
printf("   Name: %s\n", info->Meta.Name);
printf("   NUID: %s\n", info->NUID);
printf("   Chunks: %d\n", info->Chunks);
printf("   Size: %ld\n", info->Size);

char* out = nullptr;
status = objStore_GetString(&out, obs, info->Meta.Name, nullptr);
fprintf(stderr, "   > String: %u - %s\n", status, natsStatus_GetText(status));
free(out);

// download file
status = objStore_GetFile(obs, info->Meta.Name, "test.file", nullptr);
fprintf(stderr, "   > Download: %u - %s\n", status, natsStatus_GetText(status));
```
In another shell i use nats.client to create an object
```
nats object put test Makefile --name="Makefile.hallo"
```
The watcher jumps in and detects a new message in the objectStore, but i'm not able to get the object itself. Neither with `objStore_GetString`, nor with `objStore_GetFile`. 
The information in `objStoreInfo` looks okay, but then i get no result.
```
New object found: Makefile.hallo
Info: 
   Name: Makefile.hallo
   NUID: 44bd0Q7Xa8R5Bkyru0ZwOB
   Chunks: 1
   Size: 1345
   > String: 13 - Not Found
   > Download: 13 - Not Found
```
It makes no difference if i create the objectStore with `nats object add test` or via a call `js_CreateObjectStore`.

kozlovic:
The library was incorrectly using base64 raw URL encoding instead
of URL encoding, missing padding. For some object names, this will
cause other client libraries to be unable to read objects written
by the C library, or the C library not being able to get meta data
from objects written by other libraries (or nats CLI).

The fix here is to use the proper encoding.

However, existing C applications may no longer be able to retrieve
objects that they previously stored.

A tool written in Golang has been added to `tools/objstorefix`.
Running this against a NATS Server will list all object stores
and check the meta data records for all of them, fixing the ones
with bad encoding, which means rewrite them with correct encoding
and delete the old ones. After the first bad meta subject is found
all others are written again to preserve the meta history order.

For sealed object stores, a temporary stream is used to copy over
all data with proper meta subjects, when that is done, the original
object store is deleted and recreated with the content of the
temporary store. Once that is done, the object store is sealed again.

Note that if the tool does not detect any bad encoding for a given
store, no change is happening (for regular or sealed stores).

```
cd tools/objstorefix
go run main.go -url "nats://<nats server host>:<port>" -user "<user>" -password "<password>"
```
(this assumes that you have Golang installed on your system).

Resolves #933

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -614,7 +614,7 @@ objStoreInfo_Destroy(objStoreInfo *info)
 static natsStatus
 _encodeName(char **en, const char *name)
 {
-    natsStatus  s = nats_Base64RawURL_EncodeString((const unsigned char*) name, (int) strlen(name), en);
+    natsStatus  s = nats_Base64URL_EncodeString((const unsigned char*) name, (int) strlen(name), en);
     return NATS_UPDATE_ERR_STACK(s);
 }
 
--- a/test/test.c
+++ b/test/test.c
@@ -36835,6 +36835,8 @@ void test_ObjectStore_StoreMgt(void)
     const char          *esha = NULL;
     const char          **keys= NULL;
     const char          **vals= NULL;
+    natsSubscription    *sub  = NULL;
+    natsMsg             *msg  = NULL;
     int                 count = 0;
 #ifndef NATS_HAS_TLS
     const char          *nosha= "SHA-256=bm90IHN1cHBvcnRlZA==";
@@ -36875,6 +36877,10 @@ void test_ObjectStore_StoreMgt(void)
     s = objStore_GetInfo(&info, obs, "notfound", NULL);
     testCond((s == NATS_NOT_FOUND) && (info == NULL) && (nats_GetLastError(NULL) == NULL));
 
+    test("Create subscription to check encoding: ");
+    s = natsConnection_SubscribeSync(&sub, nc, "$O.TEST.M.>");
+    testCond(s == NATS_OK);
+
 #ifndef NATS_HAS_TLS
     nats_hashNoErrorOnNoSSL(true);
 #endif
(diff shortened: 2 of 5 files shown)


kozlovic:
@ripienaar Do you have any command in `nats` CLI to fix things up? If there is, I wonder if I could not port the code that I have here to the CLI. (see the description for what the issue is). If not, no big deal, I can host the go code in this repo.

kozlovic:
I need to revisit the go program to fix the store because I guess we should maintain the order of meta updates, so if an error is encountered then we need to rewrite all subsequent meta updates (even the ones that would have had proper subject name). Also, it would not work for a sealed object store, so I need first to detect if there is any error, and if there is, then create a temporary stream and write everything in that stream, then delete the original stream, copy back, etc.. sigh...

mtmk (approved):
LGTM

Issue #942, opened 2025-11-27
CRL is loaded but not enabled


thierryba:
### Observed behavior

By looking at the code for natsOptions_SetCATrustedCertificates, I have seen that nats.c calls X509_STORE_add_crl bit does not enable it by calling 

X509_STORE_set_flags(cts, X509_V_FLAG_CRL_CHECK)

### Expected behavior

I would expect the it tb be loaded so that we can use a CRL.

### Server and client version

server 2.12.2
nats.c 3.11

### Host environment

macOS + Xcode 26

thierryba:
create a PR here https://github.com/nats-io/nats.c/pull/943

PR #943, opened 2025-11-27
Enabled checks against the CRL passed in parameter

thierryba:
This is the implementation of[ issue 942](https://github.com/nats-io/nats.c/issues/942)


kozlovic (approved):
LGTM. Thanks!

PR #957, opened 2025-12-30
[ADDED] Handler invoked when a library thread is started.

### LINKED ISSUE #956: Setting core affinities for threadpool threads
I'm using the nats.c client in a latency-sensitive Linux application and would like to better control CPU placement of threads created internally by the library (IO, timers, subscription dispatch, etc.)

Is there currently any supported way to:

1. Specify CPU/core affinity for threads created by nats.c, or

2. Provide hooks/callbacks to customize thread creation (e.g. setting affinity, scheduling policy, or other pthread attributes)?

----------------

If this is not supported today, is the recommended approach to pin the entire process externally (e.g. via sched_setaffinity / taskset), or are there best practices you recommend for applications that need strict CPU isolation?

kozlovic:
New options `ThreadStartedHandler` and `ThreadStartedHandlerClosure` in the object `natsClientConfig` are added. The handler will be invoked when a thread created from the library is started. This can be used to further configure the thread.

Resolves #956

### DIFF
--- a/src/nats.h
+++ b/src/nats.h
@@ -2384,10 +2384,10 @@ typedef void (*jsPubAckHandler)(jsCtx *js, natsMsg *msg, jsPubAck *pa, jsPubAckE
 
 /** \brief Callback invoked when a thread created by the library starts.
  *
- * This callback can be used to configured a thread that is started by the library.
+ * This callback can be used to configure a thread that is started by the library.
  *
- * For instance, one could use this callback to speficy thread affinity.
- * On Linux, such code could look like:
+ * For instance, one could use this callback to specify thread affinity.
+ * On Linux, such code could look like this:
  *
  * \code{.unparsed}
  *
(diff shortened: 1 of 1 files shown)


src/unix/thread.c
```diff
@@ -35,6 +35,7 @@ _threadStart(void *arg)
 {
     struct threadCtx *c = (struct threadCtx*) arg;
 
+    nats_threadStartedHandler();
```
ArashPartow: perhaps before this line call `pthread_setname_np` to set the thread name and do something similar for win32 version.

kozlovic: It is non portable, lots of OS have different function names, etc.. for instance on macOS, you just pass the name, on linux it is thread id and name, there are glibc version requirements, etc... 

I strongly disagree on this one. If you need a name I think passing it to the callback and have you set it your callback is reasonable.

src/nats.h
```diff
@@ -2371,6 +2381,48 @@ typedef void (*jsPubAckErrHandler)(jsCtx *js, jsPubAckErr *pae, void *closure);
...
+ * s = nats_OpenWithConfig(&c);
+ *
+ * (...)
+ *
+ * void pin_current_thread(void *closure)
+ * {
+ *     pthread_t current_thread = pthread_self();
+ *     cpu_set_t cpuset;
+ *     CPU_ZERO(&cpuset);
+ *     CPU_SET(cpu, &cpuset);
```
mtmk: `cpu` is undefined

PR #965, opened 2026-02-05
[Added] Adding consumer prioritized mode

AdamPayzant:
Adding prioritized mode to consumer config and priority field to fetch request per OSSN-193


src/nats.h
```diff
@@ -1284,6 +1284,10 @@ typedef struct jsFetchRequest
         const char      *ID;            ///< the "pinned" ID of this subscription, if any; to use with the
                                         /// "pinned_client" priority policy
         const char      *Group;
+        // The use of this field require nats-server v2.12.0 or later
+        uint8_t        Priority;        ///< Sets the priority used when sending fetch requests for consumer with
```
kozlovic: @AdamPayzant I did change a bit the comment here.

src/util.c
```diff
@@ -2246,6 +2246,13 @@ nats_marshalULong(natsBuffer *buf, bool comma, const char *fieldName, uint64_t u
     return NATS_UPDATE_ERR_STACK(s);
 }
 
+natsStatus
+nats_marshalUInt8(natsBuffer *buf, bool comma, const char *fieldName, uint8_t uval)
```
kozlovic: I renamed to be more consistent with the rest.

src/util.c
```diff
@@ -2246,6 +2246,13 @@ nats_marshalULong(natsBuffer *buf, bool comma, const char *fieldName, uint64_t u
...
+{
+    natsStatus s = _marshalLongVal(buf, comma, fieldName, false, 0, (uint64_t) uval);
```
kozlovic: Not necessary I guess, but casted to `(uint64_t)` to satisfy the function definition.

kozlovic:
@AdamPayzant Can you let me know the reason you add `nats.h` in `test.c`? Again, no big deal but it should be needed, so I would want to know what you need it.

kozlovic:
@AdamPayzant I will update the PR with an update to the copyright of the files modified in this PR and remove the `nats.h` include in `test.c`. I will squash the commits. If you think the include is needed, we will add it the next time we modify the file.

kozlovic:
@AdamPayzant Ok, done. Will see if the tests run fine and then I will merge. By the way, you need to add a sign-off for each commit you do. Check with the team, but basically you should have a git commit with `-s -S` so that each commit is signed and verified (GPG key).

PR #966, opened 2026-02-06
fix(pkg-config): handle absolute install paths in libnats.pc

Kirskov:
GNUInstallDirs may set CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_INCLUDEDIR to absolute paths (e.g. on NixOS), causing libnats.pc to produce broken paths like /usr/local//nix/store/.../lib. Use IS_ABSOLUTE to check whether to prepend / or use the path directly.


Kirskov:
I tried in local the upgrade of CMAKE in 4.2.3 and everything seems to build

Issue #740, opened 2024-03-26
Request returns error status on callback, rather than from the method itself


jakeobrien:
### Observed behavior

When calling `natsConnection_Request()` that results in  a `Not Permitted` error, the error status is not returned from the `natsConnection_Request()` method.  Instead, the `Not Permitted` status is sent to the callback assigned in `natsOptions_SetErrorHandler()`. The request method completes with a status of `Timeout` several seconds after the error callback. 

This makes it impossible to tie the error to the request. 

An alternative way to do this would be to match the subscription subject in the error callback to the replyID of the request `natsMsg` handed to the `natsConnection_Request()` method, but neither of these are populated. The subscription in the callback is empty, and the request given to the request method is also empty when the method times out.

### Expected behavior

Ideally, the `natsStatus` would be returned from the `natsConnection_Request()` method and it would complete as soon as the error is received.

### Server and client version

nats-server: v2.10.3
nats.c client: v3.7.0

### Host environment

Client is on various configurations of iPhone hardware and iOS version (recent).

### Steps to reproduce

- Set up a callback with `natsOptions_SetErrorHandler()` and log the info returned
- Set up a Nats server that returns a `Not Permitted` error to requests
- Set up `natsConnection_Request()`  to log the returned status and request
- Call `natsConnection_Request()` with a timeout
- The error callback will return with an empty subscription struct first
- The request will complete only after the timeout without the underlying `Not Permitted` status

PR #967, opened 2026-02-11
fix(conn): return NATS_NOT_PERMITTED from natsConnection_Request() instead of timing out on publish permission violation

Kirskov:
Resolves https://github.com/nats-io/nats.c/issues/740


src/conn.c
```diff
@@ -2904,6 +2904,7 @@ _processPermissionViolation(natsConnection *nc, char *error)
     natsConn_Lock(nc);
     nc->err = NATS_NOT_PERMITTED;
     snprintf(nc->errStr, sizeof(nc->errStr), "%s", error);
+    _clearPendingRequestCalls(nc, NATS_NOT_PERMITTED);
```
kozlovic: No, you can't do that. This will invalidate ALL pending requests, regardless if there would have been subject to a permission error or not. The NATS protocol - as it is - simply does not allow for the invalid request to be uniquely identified and therefore the issue cannot be reliably solved in the client library.

A bit of a hack approach would be to scan the subject from the publish permission error and then clear the pending requests only for requests that have been sent on that exact subject. But I am not sure I would accept this approach since it is really tied to the error text message sent back from the server. Any change there could break this approach.

I think there would be a need for a different NATS protocol where a request would need to be provided some kind of ID (by the client) and responses from the server should include that ID. Again, this would be completely new protocols that are unlikely to be added at this point.

test/test.c
```diff
@@ -10784,6 +10784,107 @@ void test_PermViolation(void)
...
+    natsMsg             *reply   = NULL;
+    natsOptions         *opts    = NULL;
+    natsPid             pid      = NATS_INVALID_PID;
+    struct threadArg    args;
+    int64_t             start    = 0;
+    int64_t             elapsed  = 0;
+
+    // Set up the error handler callback to log the info returned.
+    s = _createDefaultThreadArgsForCbTests(&args);
+    if (s == NATS_OK)
```
kozlovic: Could have been a `IFOK(s, ..`

test/test.c
```diff
@@ -10784,6 +10784,107 @@ void test_PermViolation(void)
...
+    // Call natsConnection_Request() with a 5s timeout on a denied subject.
+    // Before the fix, the request would complete only after the timeout
+    // without the underlying Not Permitted status.
+    test("Request to denied subject returns NOT_PERMITTED: ");
+    start = nats_Now();
+    s = natsConnection_RequestString(&reply, conn, "bar", "hello", 5000);
+    elapsed = nats_Now() - start;
+    testCond(s == NATS_NOT_PERMITTED);
+
+    // The request should return well before the 5s timeout.
```
kozlovic: This could be removed.

Kirskov:
> Please review my comments, but I am inclined to close the PR since I don't think that there is a safe way to fix the issue.

Thank you for your input, I close the PR then.

PR #980, opened 2026-03-25
[CHANGED] JetStream and core requests now share same replies sub.

AdamPayzant:
Currently 2 muxers are created for a jetstream and nats core.
This can cause scale issues when thousands of clients are
connected to a cluster. This implements the solution brought
up while discussing https://github.com/nats-io/nats.c/pull/971

Note that the pub ack handler will now be invoked if the
connection or JS context is closed with `NATS_CONNECTION_CLOSED`
for pending pub acks, and `NATS_CONNECTION_DISCONNECTED`
if the connection breaks and the option `natsOptions_SetFailRequestsOnDisconnect()`
is set to `true`.

Signed-off-by: Edward Payzant <adam@synadia.com>

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2798,10 +2798,13 @@ _close(natsConnection *nc, natsConnStatus status, bool fromPublicClose, bool doC
     // to ensure that server has received all pending data.
     // Note that _flushTimeout will release the lock and wait
     // for PONG so this is why we do this early in that function.
+    // Note: Do not check for `nc->bw` length, because even if 0,
+    // it could be that we just wrote data into the socket and the
+    // buffer is currently empty, but we should still send the PING
+    // out to ensure data is fully flushed.
     if (fromPublicClose
             && (nc->status == NATS_CONN_STATUS_CONNECTED)
-            && nc->sockCtx.fdActive
-            && (natsBuf_Len(nc->bw) > 0))
+            && nc->sockCtx.fdActive)
     {
         _flushTimeout(nc, 500);
     }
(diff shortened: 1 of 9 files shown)


kozlovic:
@AdamPayzant I have been reviewing the PR and there are quite a few issues. It made me realize that some of the directions we discussed may not be exactly right. I agree with one of your comment, in that we don't need a special token for the subject because we already store `js` in `respInfo`.

Again, comments on this PR would be too extensive (because the direction may not be totally right anyway), so if you don't mind I will post later a new commit that will hopefully address the issues. If you could not update the PR so that we don't step on each other, that would be great. Thanks!

AdamPayzant:
@kozlovic Please go ahead. Thank you

kozlovic:
@claude re-review

kozlovic:
@claude re-review

kozlovic:
@claude re-review

kozlovic:
@AdamPayzant Sorry for the big changes. I worked originally on your branch with all changes squashed and unstaged so I could see what has been changes and update what needed to be. But then my changes were impossible to correctly apply, so I had to squash the branch and also rebased main.

I would still like to add JS tests that demonstrate the disconnect/close/js_destroy that now invokes the JS pub ack handler. Will add later. If you want to review, it is going to be very difficult from the diff view. I would recommend that you checkout the branch locally and go through the code.

kozlovic:
@AdamPayzant  I have been working on that one quite a bit and ran some benchmarks and the drop in performance is too big (for concurrent NATS core requests, more than 20% decrease). I am exploring another approach with much less changes to the original code and behavior for the JetStream side of things (that is, no change in disconnect/close/ctx destroy). I will hopefully post a new PR early next week. For now, I am closing this PR.

PR #982, opened 2026-04-10
[FIXED] `natsConnection_Close` was not always properly flushing

kozlovic:
In PR #189, we improved the behavior of a close to perform a flush (with a timeout of 2 seconds), and added the test `ConnCloseDoesFlush`. That was 8 years ago. Then, with the addition of write deadline, in PR #268 7 years ago, the code was changed a bit and the timeout was reduced to 500ms but more importantly, we added the check of the write buffer not being empty as the condition to trigger the flush.

But recently, the `ConnCloseDoesFlush` test started to flap and it was determined that the reason was that the buffer was empty, but the lack of PING/PONG caused data to be dropped in some cases. I am not sure why it started to show up recently though.


AdamPayzant (approved):
LGTM

PR #985, opened 2026-04-13
[ADDED] JetStream: Option to avoid internal subscription for PubAcks

kozlovic:
Normally, when calling `js_PublishAsync[Msg]()` for the first time, the library creates an internal subscription to receive publish acknowledgements from the server.

In some situations, users may want to reduce the overall number of subscriptions, so the option `jsOptionsPublishAsync.MuxReplies` will cause the JetStream context to skip the creation of the internal subscription and instead use the connection's responses muxer.

In addition, the connection response muxer has been made "direct", meaning that the responses are dispatched directly from the connection read loop. This helps with performance.


kozlovic:
@claude re-review

kozlovic:
@claude re-review.

kozlovic:
@claude re-review

kozlovic:
@claude re-review

kozlovic:
@AdamPayzant I went through more than 5 iterations before publishing this PR, so as I discussed with you when we chatted last time, I knew this was going to be a difficult PR.

I opted for the JS option because I can see a slight degradation in JS PubAsync throughput in some environments, so I wanted to let the user choose.

AdamPayzant:
@kozlovic That makes sense. I'm reviewing it now, but there's a lot to go through and a few things I'm making sure I understand from the existing implementation so it will take a little while

kozlovic:
@AdamPayzant You should take your time, there is no rush.

src/conn.c
```diff
@@ -1488,68 +1561,206 @@ natsConn_newInbox(natsConnection *nc, natsInbox **newInbox)
     return s;
 }
 
+static void
+_respHandler(natsConnection *nc, natsMsg *msg)
+{
+    const char  *subj   = NULL;
+    respInfo    *resp   = NULL;
+    bool        dmsg    = true;
+    int         kind    = 0; // 1 will be core, 2 JS, 3 subject rewrite.
```
AdamPayzant: Nitpick: Consider adding some constants for 0, 1, 2, and 3. It's just for this function but it will make the subsequent checks more clear

src/js.c
```diff
@@ -768,86 +795,231 @@ _handleAsyncReply(natsConnection *nc, natsSubscription *sub, natsMsg *msg, void
...
+        js->cond        = cond;
+        ar->repliesPfx  = pfx;
+        ar->idOffset    = (int) strlen(pfx);
+        ar->sub         = sub;
+        ar->dispatcher  = t;
+        ar->cond        = rCond;
+        ar->drainMsg    = dMsg;
+        ar->init        = true;
     }
+    if (s != NATS_OK)
```
AdamPayzant: Nitpick: Why not just an else case?

AdamPayzant (commented):
Sorry this took so long. There were a lot of changes so I wanted to make sure I was understanding everything correctly. Outside of the 2 nitpicks, this does look good to me.

Just to confirm, in the previous draft PR you mentioned some performance benchmarking, have you done the same for this changeset as well? Just want to make sure all the bases are covered

src/conn.c
kozlovic: Will do.

src/js.c
kozlovic: Sure, will change.

kozlovic:
@AdamPayzant Regarding perf, yes, I ran benchmark tests "manually" and I am working right now on making them part of the "bench" tests that we have so we can easily run against main branch but also any attempts to improve the PR.

I have switched it to Draft also so that we don't merge by accident. I realized that we change the subject space with wildcard sub from `_INBOX.<NUID>.*` to `_INBOX.<NUID>.*.*`, which seems not a big deal, but what if users use permissions and have already allowed specific users to subscribe to `_INBOX.*.*` (instead of `_INBOX.>`). That would break their deployment.

Of course we could document that and since this PR would land only in the next version (say 3.13), that is ok, but maybe I could do without changing the subject space (just have `0_<respID>` or `<jsCtxID>_<respID>` for the last token.

Once I am done with the benchmarks code, I will run against main, this PR, and modifications and see if that is acceptable. I will publish the results here when I am ready (will take some time I think).

kozlovic:
@AdamPayzant I have addressed your comments and made several other changes based on the results of the new benchmark tests (remove the muxer's lock and restore number of tokens in the response subject).

kozlovic:
@claude re-review the changes I have made. The file `test/bench.c` is a rename of `test/bench_sub_async.c`, so ignore code that has not been added/changed. Same for `test/diffstat_sub_async.go` that has been renamed and moved to `tools/bench_diff/main.go`. Ignore existing code there and focus only on the new additions.

kozlovic:
@claude re-review

AdamPayzant (approved):
At this point, this all looks good to me. If you have any more plans, let me know.

Once again, thank you for taking this on!

kozlovic:
@AdamPayzant Nothing to add. I have switched to "Ready for Review" and since you have already approved the PR, I will merge now.

PR #990, opened 2026-05-07
[Added] Adding fields for counters

AdamPayzant:
This is adding gaps in the nats.c library needed for orbit.c (draft available here: https://github.com/synadia-io/orbit.c/pull/1).


src/nats.h
```diff
@@ -1272,6 +1277,15 @@ typedef struct jsDirectGetMsgOptions
         uint64_t        Sequence;               ///< Get the message at this sequence
         const char      *NextBySubject;         ///< Get the next message (based on sequence) for that subject
         const char      *LastBySubject;         ///< Get the last message on that subject
+        int             Batch;                  ///< Gets up to batch number of messages
```
kozlovic: I am assuming that all those new fields are used in Orbit. We don't have a test in this library that would set those fields. We could make the serialization a function (and call it in `js_DirectGetMsg()`) so we could test that the serialization (marshaling) produces the desired output.

src/external_utils.h
```diff
@@ -0,0 +1,58 @@
+#ifndef _NATS_EXTERNAL_UTILS_H_
```
kozlovic: I would strongly recommend that you revert changes that exports `natsBuffer/natsBuf` APIs. You could copy the code into orbit, maybe have a "common" or "util" source folder there. The content of `buf.c` and `buf.h` is small and pretty stable at this point, so unlikely that you need to fix an issue in 2 places.

I realize that you would need to replace `NATS_MALLOC/NATS_FREE` with `malloc/free` and get rid of the `nats_setDefaultError()` and `NATS_UPDATE_ERR_STACK` calls, but that may be ok to do so.

You could even combine `buf.c` and `buf.h` in orbit in a single "h" file if that makes it easier for you to use.

kozlovic (commented):
I don't like too much the idea of exporting the buffer this way. What do you think about copying the code into orbit instead?

src/external_utils.h
AdamPayzant: I prefer having buffer code shared, but if you think it's stable enough then I'll go ahead

src/nats.h
```diff
@@ -1236,6 +1240,7 @@ typedef struct jsPubAck
         bool            Duplicate;
         char            *Batch;
         uint64_t        Count;
+        char            *Val;
```
kozlovic: nit: you could have named the field `Value` (as it is in `PubAck` in the server), even if the actual json name is `val`.

kozlovic (approved):
LGTM, but not merging yet in case you want to change the field name from `Val` to `Value`? (up to you)

PR #992, opened 2026-05-15
[Added] JS consumer reset

AdamPayzant:
Adds the js consumer reset API


AdamPayzant:
@claude re-review

src/nats.h
```diff
@@ -1184,6 +1184,17 @@ typedef struct jsConsumerPauseResponse
...
 } jsConsumerPauseResponse;
 
+/**
+ * Response returned by js_ResetConsumer.
+ *
+ * @see js_ResetConsumer
+ */
+typedef struct jsConsumerResetResponse
+{
+        uint64_t        ResetSeq;       ///< Stream sequence the consumer was reset to.
```
kozlovic, line 1197: Before going too deep in the code review, if I look at the Go client (and the server), the response contains also a "ConsumerInfo" in the response. So is it ok to limit the reset response object to just the sequence? 

If that is the case, you don't need to define a public structure for that and have the public Reset API just return a `uint64_t` instead of the structure. Can you please check with the rest of the clients team and see if the ConsumerInfo is required in the response or not?

AdamPayzant: I missed that. Response struct has been updated

test/test.c
```diff
@@ -34115,6 +34115,188 @@ void test_JetStreamConvertDirectMsg(void)
...
+    s = js_ResetConsumer(&crr, js, "s1", "c1", 0, NULL, &jerr);
+    testCond((s == NATS_OK) && (jerr == 0)
+                && (crr != NULL)
+                && (crr->Consumer != NULL)
+                && (crr->Consumer->Name != NULL)
+                && (strcmp(crr->Consumer->Name, "c1") == 0)
+                && (crr->Consumer->Stream != NULL)
+                && (strcmp(crr->Consumer->Stream, "s1") == 0)
+                && (crr->Consumer->Delivered.Consumer == 0)
+                && (crr->Consumer->NumRedelivered == 0));
```
kozlovic: What should be the expected value of `crr->ResetReq` here? You could check it.

src/nats.h
```diff
@@ -1184,6 +1184,18 @@ typedef struct jsConsumerPauseResponse
         int64_t         PauseRemaining; ///< Remaining time in nanoseconds.
 } jsConsumerPauseResponse;
 
+/**
+ * Response returned by js_ResetConsumer.
+ *
```
kozlovic, line 1189: I would put a not or warning saying that the user should not attempt to destroy the "embedded" `jsConsumerInfo` here, it will be properly destroyed when calling `jsConsumerResetResponse_Destroy`.

src/jsm.c
```diff
@@ -3553,6 +3553,125 @@ js_DeleteConsumer(jsCtx *js, const char *stream, const char *consumer,
...
+        {
+            s = nats_setDefaultError(NATS_NO_MEMORY);
+        }
+        if (freePfx)
+            NATS_FREE((char*) o.Prefix);
+    }
+
+    if ((s == NATS_OK) && (seq > 0))
+    {
+        bodyLen = snprintf(jsonBuf, sizeof(jsonBuf),
```
kozlovic, line 3654: You could put a comment above indeed saying that the serialization is guaranteed to fit in the provided buffer so there is no need to check for `-1`. Also, maybe you can have a single line here.

kozlovic (approved):
nit: Some suggestions, but otherwise LGTM.

kozlovic (approved):
LGTM

PR #999, opened 2026-06-12
Add reconnect on protocol error option

cbecker:
We found this with a client that should always try to reconnect, as it is in an edge device.

Let me know if this is a good direction, and I can add docs and potential tests.

### Detailed description
When a NATS server sends a generic -ERR on an established connection, for example Maximum Connections Exceeded, Parser Error, or any other error that is not Stale Connection, Permissions Violation, or an auth error, cnats calls _close(NATS_CONN_STATUS_CLOSED) unconditionally, bypassing the reconnect behavior entirely.

This happens regardless of whether allowReconnect is set or maxReconnect is -1 (infinite retries).

This means a transient server-side condition, such as briefly hitting a connection limit, permanently severs the client with no recovery, even when the operator has explicitly configured the client to reconnect forever.

### This PR
The fix adds an opt-in flag reconnectOnProtocolError so that callers who want resilience can route generic -ERR through the same reconnect path as all other errors, while the original close-on-error behavior
remains the default so existing users are unaffected.


cbecker:
@kozlovic I'd like to know what you think about this PR. It's handy to have nats.c resubscribe on reconnect, which is why it'd be useful to pass this new option to allow it always to reconnect whatever the failure was.

AdamPayzant:
The idea seems sound to me. Unless @kozlovic disagrees, I'm fine with you going ahead and adding tests

src/conn.c
```diff
@@ -3183,11 +3183,21 @@ natsConn_processErr(natsConnection *nc, char *buf, int bufLen)
...
     {
-        close = true;
+        bool reconnectOnErr = false;
+
         natsConn_Lock(nc);
         nc->err = NATS_ERR;
         snprintf(nc->errStr, sizeof(nc->errStr), "%s", error);
         natsConn_Unlock(nc);
+
+        natsOptions_lock(nc->opts);
```
kozlovic: When a connection is created, the "user" options are cloned and owned by the connection object and become immutable. So locking here would not be strictly needed. You could even have collected the boolean under the `natsConn_Lock()/Unlock()` above.

But that's ok to keep it this way.

kozlovic (changes requested):
The issue with this approach is that if the server closes a client because it does something wrong, then the client will automatically reconnect and likely do the wrong thing again, causing more strain on the server. Of course, the application "could do the same" by recreating the connection from scratch, so it is not that it can't be avoided at all.

Since this is not the default behavior, I would be ok adding it as an opt-in option. Please add some tests for setting/clearing the option and a test that demonstrates that it does reconnect if receiving an error (not sure on top of my head what test to demonstrate that, or if you need to use the mock server facility).

cbecker:
Thanks @kozlovic and @AdamPayzant . Now updated the docs and added a test. I verified that the same tests with the current main branch does not pass.

Removed the lock as well as suggested. Co-authored with Claude

src/conn.c
cbecker: thanks, removed.

test/test.c
```diff
@@ -18225,6 +18225,98 @@ void test_ServerErrorClosesConnection(void)
...
+
+    if (t != NULL)
+    {
+        natsThread_Join(t);
+        natsThread_Destroy(t);
+    }
+
+    testCond((s == NATS_OK) && arg.disconnected && !closedBeforeCleanup);
+
+    _destroyDefaultThreadArgs(&arg);
```
kozlovic, line 18318: This is causing a data race. The thread that you joined/destroyed still will have triggered the async callback for the closed event on connection close. So try to add this before this call:
```
_waitForConnClosed(&arg);
```

kozlovic (changes requested):

test/test.c
cbecker: thanks, done!

kozlovic (approved):

cbecker:
Thanks @kozlovic ! Are you planning to create a release/tag soon? We'd like to update to this version and we can use commits, though we strive to use tags when possible, e.g. v3.14.0

kozlovic:
@cbecker I understand, but we just got a release 2 weeks ago, so we may wait a bit before the next to have more added to the release. But @AdamPayzant could chime in to see if he wants to do the release soon.

cbecker:
No worries, we can handle without a release, thanks for your response.

PR #1000, opened 2026-06-29
KV per-key TTL (parity with nats.go #1864)

NormB:
Bring nats.c to parity with nats.go #1864 ("KeyValue per key TTL support and limit markers") by adding a KV-level per-key TTL write surface on top of the existing per-message TTL primitives (#863):

- kvConfig.LimitMarkerTTL: at bucket create, enable AllowMsgTTL + SubjectDeleteMarkerTTL on the backing stream.
- kvStore_CreateWithTTL / kvStore_CreateStringWithTTL: per-key TTL on create, reusing the marker-aware re-create retry (a re-create over a DEL/PURGE/MaxAge tombstone succeeds and keeps the TTL).
- PurgeTTL: kvPurgeOptions.TTL threads a per-message TTL onto the purge marker so the purge tombstone auto-expires.
- kvStore_Create / kvStore_Update now delegate to internal _kvCreateWithTTL / _kvUpdateWithTTL with ttl=0, so existing behavior is unchanged.
- test_KeyValueCreateWithTTL (test/test.c + test/list_test.txt): create-with-TTL, get-before-expiry, create-over-live-key-fails, server-side expiry, re-create over the expired marker, and Purge+PurgeTTL. JS_SETUP(2,11,0) skips it on < 2.11; verified green on nats-server 2.12.11 & 2.14.2.


test/list_test.txt
```diff
@@ -153,6 +153,7 @@ _test(KeyValueRePublish)
 _test(KeyValueWatch)
 _test(KeyValueWatchMulti)
 _test(KeyValueLimitMarkerTTL)
+_test(KeyValueCreateWithTTL)
```
AdamPayzant: Nit: Tests are in alphabetical order, please move between _test(KeyValueBasics) and _test(KeyValueCrossAccount)

src/kv.c
```diff
@@ -267,6 +267,15 @@ js_CreateKeyValue(kvStore **new_kv, jsCtx *js, kvConfig *cfg)
         sc.AllowDirect = true;
         sc.RePublish = cfg->RePublish;
 
+        // [parity with nats.go #1864] LimitMarkerTTL enables per-message TTL
```
AdamPayzant: Nit: Remove this and subsequent parity notes

src/kv.c
```diff
@@ -684,8 +693,33 @@ kvStore_PutString(uint64_t *rev, kvStore *kv, const char *key, const char *data)
...
+        po.MsgTTL = ttl;
+    s = _putEntry(rev, kv, &po, key, data, len);
+    return NATS_UPDATE_ERR_STACK(s);
+}
+
+// internal: create-if-absent carrying an optional per-key TTL. Reuses the
+// marker-aware retry so a re-create over a DEL/PURGE/MaxAge tombstone succeeds AND
+// keeps the TTL on both the initial attempt and the re-create-over-marker attempt.
+static natsStatus
+_kvCreateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *data, int len, int64_t ttl)
```
AdamPayzant: `kvStore_CreateWithTTL` just wraps this function. Why not name this function `kvStore_CreateWithTTL`, remove the wrapper, and update `kvStore_Create`?

src/kv.c
```diff
@@ -684,8 +693,33 @@ kvStore_PutString(uint64_t *rev, kvStore *kv, const char *key, const char *data)
...
 }
 
-natsStatus
-kvStore_Create(uint64_t *rev, kvStore *kv, const char *key, const void *data, int len)
+// internal: CAS update carrying an optional per-key TTL (milliseconds).
+// ttl <= 0 means "no TTL" (identical to the pre-existing update). The TTL rides
+// jsPubOptions.MsgTTL (added by #863); combining it with the CAS predicate in the
+// same publish is exactly what callers otherwise have to hand-roll. Parity: nats.go #1864.
+static natsStatus
+_kvUpdateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *data, int len,
```
AdamPayzant: Why shouldn't this function be public alongside `kvStore_CreateWithTTL`?

AdamPayzant (changes requested):
Hi Norm. Thank you for the contribution! I've added some notes

NormB:
@AdamPayzant Thanks for the quick review. Will take care of the first three comments. For the last one, will make it public.

Added kvStore_UpdateStringWithTTL to mirror the CreateString* pair
 
A couple of things I ran into while expanding the tests:

- TTL granularity — I left this server-enforced rather than validating client-side: the server requires ≥ 1s (JSMessageTTLInvalidErr on a sub-second value), but a non-whole-second value like 1500ms is accepted. Documented on both functions, with tests for both cases. Glad to add a client-side check instead if you'd prefer.

- On the create path a rejected TTL comes back with an empty error string — the marker-aware re-create runs a _getEntry that clears the last error; the update path preserves the server message. I assert the server text (invalid per-message TTL) on the update path. I can look at preserving it through the create retry as a follow-up if it's worth it.

AdamPayzant (approved):
LGTM

NormB:
@AdamPayzant Any thoughts on the two items I mentioned above? TTL granularity and Propagating errors to the caller?

AdamPayzant:
I thought it was fine without, personally, but it would be a pretty quick PR if you want to add them

PR #1001, opened 2026-06-30
KV per-key TTL: preserve the create-path error through the marker-aware retry

NormB:
Follow-up to the KV per-key TTL work (PR https://github.com/nats-io/nats.c/pull/1000). Per review, TTL granularity
is left entirely to the server (no client-side check), so a future server-side
change to the minimum needs no client update.

- kvStore_CreateWithTTL: the marker-aware retry calls _getEntry, which
  overwrites the thread-local last error, so a failed create returned an empty
  error string. This is pre-existing behavior inherited from kvStore_Create.
  Suppress error-stack updates across the _getEntry probe (nats_doNotUpdateErrStack)
  and restore the original failure on return; re-enable and clear only when we
  actually re-create over a tombstone, so that path reports its own error. The
  server message (e.g. "invalid per-message TTL", "per-message TTL is disabled")
  now reaches the caller.

- Remove the client-side TTL granularity check (kvStore_*WithTTL and the purge
  marker TTL) and kvErrInvalidTTL; the server enforces the 1s minimum.

- Tests: sub-second create/update now assert the server's error survives the
  retry, plus a TTL-disabled-bucket create that exercises the same path.

- Docs: note the server's 1s minimum without asserting a client-side rejection.

### DIFF
--- a/src/kv.c
+++ b/src/kv.c
@@ -696,7 +696,9 @@ kvStore_PutString(uint64_t *rev, kvStore *kv, const char *key, const char *data)
 // CAS update carrying an optional per-key TTL (milliseconds). ttl <= 0 means
 // "no TTL" (identical to the pre-existing update). The TTL rides
 // jsPubOptions.MsgTTL; combining it with the CAS predicate in the same publish
-// is what callers would otherwise have to hand-roll.
+// is what callers would otherwise have to hand-roll. The server enforces the
+// per-message TTL minimum (currently 1 second); the client does not duplicate
+// that check, so a future server-side change needs no client update.
 natsStatus
 kvStore_UpdateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *data, int len,
                       uint64_t last, int64_t ttl)
@@ -734,15 +736,30 @@ kvStore_CreateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *d
         return s;
 
     // Since we have tombstones for DEL ops for watchers, this could be from that
-    // so we need to double check.
+    // so we need to double check. The _getEntry probe below would otherwise
+    // overwrite (and, on NATS_NOT_FOUND, clear) the original create error, so
+    // suppress error-stack updates around it and restore the original error on
+    // return. We re-enable updates only when we actually re-create over a
+    // tombstone, so that attempt reports its own error.
+    nats_doNotUpdateErrStack(true);
+
     ls = _getEntry(&e, &deleted, kv, key, 0);
     if (ls == NATS_OK)
     {
         if (deleted)
+        {
+            nats_doNotUpdateErrStack(false);
+            nats_clearLastError();
             s = kvStore_UpdateWithTTL(rev, kv, key, data, len, kvEntry_Revision(e), ttl);
-
+            kvEntry_Destroy(e);
+            return NATS_UPDATE_ERR_STACK(s);
+        }
         kvEntry_Destroy(e);
     }
+
+    // Not a tombstone re-create: surface the original create failure (status and
+    // stack as they were before the _getEntry probe).
+    nats_doNotUpdateErrStack(false);
     return NATS_UPDATE_ERR_STACK(s);
 }
 
--- a/src/kv.c
+++ b/src/kv.c
@@ -736,30 +736,22 @@ kvStore_CreateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *d
         return s;
 
     // Since we have tombstones for DEL ops for watchers, this could be from that
-    // so we need to double check. The _getEntry probe below would otherwise
-    // overwrite (and, on NATS_NOT_FOUND, clear) the original create error, so
-    // suppress error-stack updates around it and restore the original error on
-    // return. We re-enable updates only when we actually re-create over a
-    // tombstone, so that attempt reports its own error.
+    // so we need to double check. Suppress error-stack updates across the probe
+    // so a NATS_NOT_FOUND (or any other _getEntry result) cannot overwrite/clear
+    // the original create error we still want to surface below.
     nats_doNotUpdateErrStack(true);
-
     ls = _getEntry(&e, &deleted, kv, key, 0);
+    nats_doNotUpdateErrStack(false);
     if (ls == NATS_OK)
     {
         if (deleted)
         {
-            nats_doNotUpdateErrStack(false);
+            // clear original stack error before calling update.
             nats_clearLastError();
             s = kvStore_UpdateWithTTL(rev, kv, key, data, len, kvEntry_Revision(e), ttl);
-            kvEntry_Destroy(e);
-            return NATS_UPDATE_ERR_STACK(s);
         }
         kvEntry_Destroy(e);
     }
-
-    // Not a tombstone re-create: surface the original create failure (status and
-    // stack as they were before the _getEntry probe).
-    nats_doNotUpdateErrStack(false);
     return NATS_UPDATE_ERR_STACK(s);
 }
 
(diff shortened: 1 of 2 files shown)


AdamPayzant:
@kozlovic Can you look over the error propagation here? I'm not sure how I feel and could use a second set of eyes

src/kv.c
```diff
@@ -704,6 +719,10 @@ kvStore_UpdateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *d
     natsStatus      s;
     jsPubOptions    po;
 
+    s = _validatePerKeyTTL(ttl);
```
kozlovic: I would remove all changes related to the check of the TTL. The Go client does not do that if I am not mistaken. I think it is best to let the server handle the check. If in the future a change is made to allow lower, or have a different lower limit, it will be done in the server and client libraries don't need to be updated.

src/kv.c
```diff
@@ -733,17 +753,39 @@ kvStore_CreateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *d
     if (s == NATS_OK)
         return s;
 
+    // A malformed request (invalid key, sub-second TTL, ...) is rejected
+    // client-side with NATS_INVALID_ARG and already carries the right error;
+    // there is no tombstone to re-check, so return it untouched.
+    if (s == NATS_INVALID_ARG)
```
kozlovic: Given that this will no longer be the case when removing the check, are you saying that the original code still had an issue that the error would be "blanked"?

src/kv.c
```diff
@@ -733,17 +753,39 @@ kvStore_CreateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *d
...
+        return NATS_UPDATE_ERR_STACK(s);
+
+    // Otherwise the create may have failed because a DEL/PURGE tombstone exists
+    // for the key (a re-create over it is allowed) or because the key is live (a
+    // genuine create-if-absent conflict). The re-check below calls _getEntry,
+    // which overwrites the thread-local last error, so snapshot the original
+    // failure now and restore it on the paths that return it -- otherwise the
+    // caller would see this create fail with an empty error string.
+    {
```
kozlovic: If the goal is to return the original error (`s` and stack up to that point), regardless of what `_getEntry` returns or `kvStore_UpdateWithTTL` returns, then I would replace all that with:
```
nats_doNotUpdateErrStack(true);

ls = _getEntry(...
(..)

nats_doNotUpdateErrStack(false);
```
If we want to report what `kvStore_UpdateWithTTL` get, I would do in addition:
```
if (deleted)
{
    nats_doNotUpdateErrStack(false);
    nats_clearLastError();
    s = kvStore_UpdateWithTTL(rev, kv, key, data, len, kvEntry_Revision(e), ttl);
     kvEntry_Destroy(e);
}
(...)
```

src/kv.c
```diff
@@ -803,6 +845,15 @@ _delete(kvStore *kv, const char *key, bool purge, const kvPurgeOptions *opts)
     if (!validKey(key))
         return nats_setError(NATS_INVALID_ARG, "%s", kvErrInvalidKey);
 
+    // A purge marker rides a per-message TTL too (kvPurgeOptions.TTL), so it is
```
kozlovic: Again, would remove this.

src/nats.h
```diff
@@ -1699,8 +1699,9 @@ typedef struct kvPurgeOptions
         // written by kvStore_Purge() is published with this per-message TTL (in
         // milliseconds), so the server auto-expires the purge tombstone after the
         // duration. Requires the bucket created with kvConfig.LimitMarkerTTL > 0 and
-        // nats-server v2.11+. Ignored by kvStore_Delete() (delete markers use the
-        // bucket-level LimitMarkerTTL).
+        // nats-server v2.11+. Must be at least 1 second (1000 ms): a sub-second value
```
kozlovic: Would revert changes. Again, you could mention that the minimum is 1000 ms, but not mention that below would return `NATS_INVALID_ARG`. Again, leave the door open to the server possibly change this enforcement in the future.

src/nats.h
```diff
@@ -8567,8 +8571,10 @@ kvStore_UpdateString(uint64_t *rev, kvStore *kv, const char *key, const char *da
  * server expires this key after `ttl` milliseconds. The TTL is set on the same CAS
  * publish, so a plain update can adopt a TTL atomically. Requires the bucket to have
  * been created with `kvConfig.LimitMarkerTTL > 0` and nats-server v2.11+; otherwise the
- * server rejects the TTL (JSMessageTTLDisabledErr). The server requires the TTL to be at
- * least 1 second (1000 ms); a smaller value is rejected with JSMessageTTLInvalidErr.
+ * server rejects the TTL (JSMessageTTLDisabledErr). The TTL must be at least 1 second
```
kozlovic: Same, revert.

test/test.c
```diff
@@ -37444,9 +37445,11 @@ void test_KeyValueCreateWithTTL(void)
     kvEntry_Destroy(e);
     e = NULL;
 
-    test("CreateWithTTL over a live key fails (create-if-absent preserved): ");
+    test("CreateWithTTL over a live key fails, error preserved (create-if-absent): ");
```
kozlovic, line 37448: Revert all changes related to client-side check. If there was really an issue with error stack the way we report it, apply the fix I mentioned above and have your test updated to verify that.

kozlovic (changes requested):

NormB:
Good call — pushed an update that drops all the client-side TTL checks (kv.c, nats.h docs, and tests) and leaves enforcement to the server.

On the blanking question: yes, it's real and pre-existing — kvStore_Create on main has the same _getEntry-after-failure path, so the probe overwrites the original error before we return it. I applied your nats_doNotUpdateErrStack/nats_clearLastError suggestion as-is, and the test now asserts the server's error survives the retry (verified it goes red without the guard).

Thanks for the review.

src/kv.c
```diff
@@ -734,15 +736,30 @@ kvStore_CreateWithTTL(uint64_t *rev, kvStore *kv, const char *key, const void *d
...
+    // return. We re-enable updates only when we actually re-create over a
+    // tombstone, so that attempt reports its own error.
+    nats_doNotUpdateErrStack(true);
+
     ls = _getEntry(&e, &deleted, kv, key, 0);
     if (ls == NATS_OK)
     {
         if (deleted)
+        {
+            nats_doNotUpdateErrStack(false);
```
kozlovic: Actually, could this all be simplified by just surrounding the `_getEntry()` with the "do not update" calls? Something like this:
```
nats_doNotUpdateErrStack(true);
ls = _getEntry(&e, &deleted, kv, key, 0);
nats_doNotUpdateErrStack(false);
if (ls == NATS_OK)
{
    if (deleted)
    {
        // clear original stack error before calling update.
        nats_clearLastError();
        s = kvStore_UpdateWithTTL(rev, kv, key, data, len, kvEntry_Revision(e), ttl);
    }
    kvEntry_Destroy(e);
}
```
Seems like it would be less changes and does the same, no?

kozlovic (changes requested):
Sorry, looks like my original proposal could have been simplified. Could you verify and if so update the code? Thanks!

src/kv.c
NormB: Your update looks good. Running some tests locally with them.

kozlovic: Thanks. I have updated the PR title/description with the content of your updated commit.

NormB: A little off topic, but if you're interested, have been working toward adding NATS support to OpenSIPS. It's not complete, but getting close.  github.com/NormB/OpenSIPS feature/nats.

SIP registrations make heavy use of TTLs so this improvement is useful.

kozlovic (approved):
LGTM

PR #1002, opened 2026-07-14
Flusher latency optimization

AdamPayzant:
Currently, the flusher thread has a 1ms delay. While improving throughput, this can cause excess latency when sending low volume messages. Configuring SendAsap or using the natsConnection_Send family of functions skips flusher entirely by calling `natsConn_bufferFlush` directly, but this dramatically reduces throughput and blocks the user thread for the write.

These changes allow for the delay to be set by the user and have the flusher publish immediately if no new writes come in during a yield and if no publishes have previously occured within the window. This allows for single or sparse messages to avoid the added latency waiting for aggregation without impacting maximum throughput.


src/conn.c
```diff
@@ -2586,16 +2591,31 @@ _flusher(void *arg)
...
+        // we woke up, and writes are frequent (a flush occurred within the
+        // last accumulation window).
+        // For a lone pending write - sparse traffic or a synchronous
+        // request/reply - flush immediately instead.
+        if ((nc->opts->flusherWait > 0))
+        {
+            kicks = nc->flusherKicks;
+            natsConn_Unlock(nc);
+            natsThread_FastYield();
+            natsConn_Lock(nc);
```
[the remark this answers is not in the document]
AdamPayzant: @kozlovic I would prioritizing publishing what's already come in like we did in the previous solution, but it's an easy enough change if you feel that this would actually be better

src/unix/cond.c
```diff
@@ -89,6 +89,36 @@ natsCondition_AbsoluteTimedWait(natsCondition *cond, natsMutex *mutex, int64_t a
...
+
+    r = pthread_cond_timedwait(cond, mutex, &ts);
+
+    if (r == 0)
+        return NATS_OK;
+
+    if (r == ETIMEDOUT)
+        return NATS_TIMEOUT;
+
+    return nats_setError(NATS_SYS_ERROR, "pthread_cond_timedwait error: %d", r);
```
kozlovic, line 119: Should we fix `_timedWait` then?

src/unix/thread.c
```diff
@@ -130,6 +130,12 @@ natsThread_Yield(void)
     sched_yield();
 }
 
+void
```
kozlovic: I don't think we need this function. See my comment on the Windows's one.

src/win/thread.c
```diff
@@ -146,6 +146,13 @@ natsThread_Yield()
     nats_Sleep(1);
 }
 
+void
```
kozlovic: We may not need this function. What don't we just fix `natsThread_Yield()` to simply use `Sleep(0)` (without any call to `SwithToThread()` check)? According to doc (https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep):
```
A value of zero causes the thread to relinquish the remainder of its time slice to any other thread that is ready to run. If there are no other threads ready to run, the function returns immediately, and the thread continues execution.
```
I know that I have a comment that says that Sleep(0) does not help for the reconnect test on Windows, but maybe we should have a closer look at the reconnect logic instead, there may be something there to fix instead of artificially sleeping for 1ms on Windows, maybe by simply have a condition variable wait in the reconnect thread instead (either for all platforms or just windows).

src/natsp.h
```diff
@@ -965,6 +979,11 @@ natsThread_Detach(natsThread *t);
 void
 natsThread_Yield(void);
 
+// Yields to another runnable thread without deliberately sleeping. Unlike
+// natsThread_Yield(), this must not use a millisecond sleep on Windows.
+void
+natsThread_FastYield(void);
```
kozlovic: Again, we may not need this one.

kozlovic (changes requested):

src/win/thread.c
AdamPayzant: Sounds good. I'll take a look and see

kozlovic:
@AdamPayzant I would like to post some modifications (a new commit) to this PR, let me know if that is ok. This addresses the issues I mentioned.

AdamPayzant:
@kozlovic Please, go ahead!

AdamPayzant:
Your changes look good to me

kozlovic:
@AdamPayzant I updated the PR, let me know what you think.

kozlovic (approved):
LGTM!

PR #1003, opened 2026-07-21
Add docs.nats.io examples under examples/docs

Jarema:
Adds 177 self-contained C examples that back the C code tabs on the new docs.nats.io — one file per documentation snippet, `NATS-DOC-START/END` comments delimiting the region the docs render. Covers Core NATS, JetStream (incl. atomic batch publish, per-message TTL, priority groups, consumer pausing), KV, micro/services, and resilience examples.

Build integration: `examples/docs/CMakeLists.txt` mirrors `examples/getstarted` (globs `*.c`, `docs-` target prefix) and is added to the root CMakeLists — so the default `NATS_BUILD_EXAMPLES=ON` compiles all of them in CI. Verified locally: full CMake build clean on macOS/arm64, and sample examples executed against nats-server 2.14.0 with the documented output.

Not covered because nats.c has no API for them: batched direct get, stream snapshot/restore, object store.

The docs repo fetches these at build time (same mechanism as the other clients' `examples/docs` dirs — see nats-io/nats.go#2106).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AdamPayzant (approved):
Look's good as first draft for doc examples

PR #1006, opened 2026-08-05
libuv adapter: NULL-guard nle->handle at the two observed crash sites

### LINKED ISSUE #1005: libuv adapter: crash after silent-failure disconnect
### Observed behavior

Under silent-failure disconnect (server frozen, no TCP FIN) the NATS library's close sequence re-enters the adapter with `nle->handle already` cleared by a prior close path. The adapter then dereferences the NULL handle through libuv and crashes.

Two dereference sites have been observed to segfault in that state:

  1. **uvPollUpdate line 197**
```
uv_close((uv_handle_t*) nle->handle, uvHandleClosedCb);
```
Triggered when a second call to `uvPollUpdate` removes the last remaining event flag and enters the close  lock with a handle that the first call already closed and nulled at line 198.

  2. **natsLibuv_Read line 441**
 ```
uv_poll_stop(nle->handle);
```
Triggered when natsLibuv_Read is called with add=false to remove a read watch after the underlying handle as been closed and nulled by a prior uvPollUpdate close sequence.

### Expected behavior

No crash, no-op on NULL handle, call libuv operation on valid handle

### Server and client version

* NATS server `v2.10.7`
* NATS.c version `v3.13.0`
* libuv `v1.44.2`

### Host environment

Reproduced on different linux environments:
* debian bookworm-slim amd64
* custom yocto distro arm64

### Steps to reproduce

* Silent-failure disconnect of nats-server (`kill -STOP $(pidof nats-server)`)
* libnats calls `_evStopPolling` twice during reconnect-exhaustion teardown (`conn.c:2260` then `conn.c:2648`), and the nats-c libuv adapter has no NULL guard on nle->handle at two dereference sites.
* Deterministic SIGSEGV crash of client binary ~6 min later

### PR DESCRIPTION
**Pull request by arnaudhe**
### Description

Add NULL-guards in libuv adapter, to prevent crash on silent-failure disconnect from server

### Related Issues

Fixes https://github.com/nats-io/nats.c/issues/1005

### DIFF
--- a/src/adapters/libuv.h
+++ b/src/adapters/libuv.h
@@ -194,8 +194,11 @@ uvPollUpdate(natsLibuvEvents *nle, int eventType, bool add)
     }
     // Both read and write events have been removed, this signal that the socket
     // should be closed prior to a reconnect or during natsConnection_Close().
-    uv_close((uv_handle_t*) nle->handle, uvHandleClosedCb);
-    nle->handle = NULL;
+    if (nle->handle != NULL)
+    {
+        uv_close((uv_handle_t*) nle->handle, uvHandleClosedCb);
+        nle->handle = NULL;
+    }
     // We have stopped polling for events for this socket and are in the event
     // loop thread, so we invoke this so that the NATS C client library can
     // proceed with closing the socket.
@@ -437,7 +440,7 @@ natsLibuv_Read(void *userData, bool add)
     bool            sched;
 
     // If we remove, first stop polling immediately, then proceed as usual.
-    if (!add)
+    if (!add && nle->handle != NULL)
         uv_poll_stop(nle->handle);
 
     sched = ((uv_key_get(&uvLoopThreadKey) != nle->loop) ? true : false);

### CODE REVIEW

**Comment by AdamPayzant** [comment]
The guards are fine and fix the immediate seg fault, but I believe there's a larger bug going on here if `nle->handle` is already set to `NULL`. Let me investigate this further

**Comment by AdamPayzant** [comment]
I believe this should fix the issue at the source, but please verify.

@kozlovic Can you check and make sure this won't have any collateral that I've missed?

```
diff --git a/src/conn.c b/src/conn.c
index c5e2365f..70de0e9b 100644
--- a/src/conn.c
+++ b/src/conn.c
@@ -88,6 +88,9 @@ _processOpError(natsConnection *nc, natsStatus s, bool initialConnect);
 static natsStatus
 _flushTimeout(natsConnection *nc, int64_t timeout);
 
+static natsStatus
+_evStopPolling(natsConnection *nc);
+
 static bool
 _processAuthError(natsConnection *nc, int errCode, char *error);
 
@@ -473,6 +476,13 @@ _createConn(natsConnection *nc)
 
     if (s != NATS_OK)
     {
+        if (nc->sockCtx.fdActive)
+        {
+            natsSock_Close(nc->sockCtx.fd);
+            nc->sockCtx.fd       = NATS_SOCK_INVALID;
+            nc->sockCtx.fdActive = false;
+        }
+
         // reset the deadline
         natsSock_ClearDeadline(&nc->sockCtx);
     }
@@ -1930,14 +1940,23 @@ _doReconnect(void *arg)
             // we will report.
             nc->err = s;
 
+            // If we did re-attach then the adapter owns this socket.
+            // Stop polling so that it releases this socket and closes it
+            // from the event loop thread. Otherwise it would be left
+            // polling a socket that we have closed here.
+            if (nc->el.attached && nc->sockCtx.useEventLoop)
+                s = _evStopPolling(nc);
+
+            if (s != NATS_OK)
+            {
+                natsSock_Close(nc->sockCtx.fd);
+                nc->sockCtx.fd = NATS_SOCK_INVALID;
+            }
+            nc->sockCtx.fdActive = false;
+
             // Reset status
             s = NATS_OK;
 
-            // Close the socket since we were connected, but a problem occurred.
-            // (not doing this would cause an FD leak)
-            natsSock_Close(nc->sockCtx.fd);
-            nc->sockCtx.fd = NATS_SOCK_INVALID;
-
             // We need to re-activate the use of pending since we
             // may go back to sleep and release the lock
             nc->usePending = true;
```

**Comment by kozlovic** [comment]
@AdamPayzant For the change:
```
+            if (nc->el.attached && nc->sockCtx.useEventLoop)
+                s = _evStopPolling(nc);
+
+            if (s != NATS_OK)
+            {
+                natsSock_Close(nc->sockCtx.fd);
+                nc->sockCtx.fd = NATS_SOCK_INVALID;
+            }
+            nc->sockCtx.fdActive = false;
+
```
I would add before the `if (s != NATS_OK)` something like:
```
// For non event loop cases, or if stop polling failed, close the socket.
if (s != NATS_OK)
...
```
For the code that adds the socket close in `_createConn`, I guess this is unrelated to event loop, and it is for when the socket is created but we would fail creating a buffer, correct?

kozlovic:
Also, note that I was not able to reproduce the reported issue on macOS with libuv 1.51.0. So it could be an issue that was resolved in a more recent libuv version that the OP uses or on different platform.

AdamPayzant:
> @AdamPayzant For the change:
> 
> ```
> +            if (nc->el.attached && nc->sockCtx.useEventLoop)
> +                s = _evStopPolling(nc);
> +
> +            if (s != NATS_OK)
> +            {
> +                natsSock_Close(nc->sockCtx.fd);
> +                nc->sockCtx.fd = NATS_SOCK_INVALID;
> +            }
> +            nc->sockCtx.fdActive = false;
> +
> ```
> 
> I would add before the `if (s != NATS_OK)` something like:
> 
> ```
> // For non event loop cases, or if stop polling failed, close the socket.
> if (s != NATS_OK)
> ...
> ```
> 
> For the code that adds the socket close in `_createConn`, I guess this is unrelated to event loop, and it is for when the socket is created but we would fail creating a buffer, correct?

Yes, if _createConn fails after the socket has been cleaned up, it'll be leaked. Unrelated but claude caught reviewing my changes and it's probably worth adding.

I was able to reproduce on Linux with libuv 1.52.1 off of main.

PR #1008, opened 2026-08-05
core: destroy parser on disconnect in libuv-adapter mode

### LINKED ISSUE #1007: libuv adapter: parser not destroyed on disconnect
### Observed behavior

In libuv-adapter mode, `nc->ps` (the wire-protocol parser) survives every reconnect cycle because there is no _readLoop thread whose exit cleanup would destroy it.

In blocking-IO mode the parser is destroyed per cycle at `src/conn.c:2398-2399` when _readLoop exits on status=RECONNECTING; the libuv path simply has no equivalent.

If a slow-consumer kick (or any TCP-level disconnect) lands while the parser is mid-protocol-token. Most commonly state=MSG_ARG, drop=1 (`\r` seen, awaiting `\n` of a `MSG <subject> <sid> <size>\r\n` header). The stale parser corrupts the new socket's INFO greeting on every subsequent reconnect attempt. The \n-handler in MSG_ARG computes
```
  start = buf + afterSpace   // ~32750, far past INFO's ~447 bytes
  len   = (i - drop) - afterSpace   // garbage
```
and reads into adjacent heap → NATS_PROTOCOL_ERROR → _processOpError → reconnect → same stuck parser → forever loop with no application recovery.

### Expected behavior

Make sure the parser to be destroyed at each disconnect.

Proposed fix:
Deferred destroy via a bool `psReset` flag on natsConnection. Set in `_processOpError` under the connection lock from any thread, inside the existing `if (nc->el.attached)` block (so blocking-IO mode is untouched).
Consumed in ProcessReadEvent at entry, under the lock, on the event-loop thread (the same thread that runs natsParser_Parse for this nc) so the destroy can never race with an in-flight Parse.

This restores symmetry with blocking-IO's per-cycle parser lifecycle:
  blocking-IO: _readLoop start (create) ... _readLoop exit (destroy)
  libuv:             ProcessReadEvent lazy-create ... CONSUME psReset destroys

### Server and client version

* NATS server v2.10.7
* NATS.c version v3.13.0
* libuv v1.44.2

### Host environment

_No response_

### Steps to reproduce

_No response_

arnaudhe:
### Description

Fix an issue where protocol parser is not destroyed at disconnect when running with libuv adapter. Implements deferred destroy via a bool psReset flag on natsConnection.

### Related Issues

Fixes https://github.com/nats-io/nats.c/issues/1007

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -2469,6 +2469,9 @@ _processOpError(natsConnection *nc, natsStatus s, bool initialConnect)
             // but the actual socket close will be done from the event loop
             // adapter by calling natsConnection_ProcessCloseEvent().
             ls = _evStopPolling(nc);
+
+            // Libuv equivalent of _readLoop's per-cycle parser destroy.
+            nc->psReset = true;
         }
 
         // Fail pending flush requests.
@@ -4579,6 +4582,15 @@ natsConnection_ProcessReadEvent(natsConnection *nc)
     {
         natsConn_Unlock(nc);
         return;
+    }
+
+    // Same event-loop thread runs both this destroy and natsParser_Parse below,
+    // so the in-flight Parse on the previous socket has already returned.
+    if (nc->psReset)
+    {
+        natsParser_Destroy(nc->ps);
+        nc->ps = NULL;
+        nc->psReset = false;
     }
 
     if (nc->ps == NULL)
--- a/src/natsp.h
+++ b/src/natsp.h
@@ -866,6 +866,7 @@ struct __natsConnection
     char                errStr[256];
 
     natsParser          *ps;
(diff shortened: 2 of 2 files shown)


AdamPayzant (changes requested):
The changes look good to me, but please add a test for these changes. If you need help or want me to do it just let me know

kozlovic:
@AdamPayzant Not sure how the user can test that in the test suite to be honest. Would have to be using a mock, capture the nc->ps address, send a bad protocol, make sure that nc->ps is changed I guess?

AdamPayzant:
@kozlovic Maybe I didn't look at it deep enough, but I thought `test_EventLoop` essentially already had that groundwork done? At least enough for testing changes in `natsConnection_ProcessReadEvent`

PR #1011, opened 2026-08-14
comsock: apply socket setup after the proxy connect callback

### LINKED ISSUE #1010: natsSock_ConnectTcp skips socket setup on the proxyConnectCb path, stalling publishes when TLS is enabled
### Observed behavior

When a connection is established through `natsOptions_SetProxyConnHandler()` **and** TLS is
enabled, every `natsConnection_Publish*()` call blocks until the server happens to send something
— in practice until the server's next connection PING. With a default `ping_interval` that means
minutes per publish. There is no error, no reconnect and nothing in the logs; the connection simply
appears to stall.

Cause: `natsSock_ConnectTcp()` invokes the proxy callback and returns immediately, skipping the
socket setup that the regular connect path performs a few lines below:

```c
    // Call the proxy connect callback if provided
    if (ctx->proxyConnectCb != NULL)
    {
        // Invoke the proxy connect callback.
        s = ctx->proxyConnectCb(&ctx->fd, host, port, ctx->proxyConnectClosure);

        // If there was a deadline, reset the deadline with whatever is left.
        if (totalTimeout > 0)
            _resetDeadline(ctx, start, totalTimeout);

    }
```

Not applied on this path:

```c
    s = natsSock_SetBlocking(ctx->fd, false);
    ...
    s = natsSock_SetCommonTcpOptions(ctx->fd);
```

The callback is documented as "creates a socket, uses it for proxy verification, and returns it in
`fd`", so the socket it returns is in whatever state the application's own connect code left it in
— typically blocking and without `TCP_NODELAY`.

TLS then turns that into a lock chain, because `_processConnInit()` deliberately keeps the socket
non-blocking when TLS is in use and therefore never fixes it up:

```c
    // If there is no write deadline option, switch to blocking socket here...
    // unless it is SSL, in which case we need to keep it non-blocking.
    if ((s == NATS_OK) && (nc->opts->writeDeadline <= 0) && (nc->sockCtx.ssl == NULL))
        s = natsSock_SetBlocking(nc->sockCtx.fd, true);
```

| thread | holds | waits for |
|---|---|---|
| `_readLoop` | `ctx->sslMu` — `natsSock_Read()` takes it around `SSL_read()` | `SSL_read()` parked inside a blocking `recv()`, i.e. until the next inbound byte |
| `_flusher` | `nc->mu` — held across `natsConn_bufferFlush()` | `ctx->sslMu` |
| application thread | — | `nc->mu` in `natsConn_publish()` |

A characteristic symptom is that publish durations **alternate**: a publish that only buffers and
signals the flusher returns instantly, and the *next* one runs into the held `nc->mu`.

Measured in our application over an HTTP CONNECT proxy with TLS, before and after applying the
two missing calls locally — same machine, same environment, only the client library differing:

| operation | v3.11.0 as shipped | with the two calls added |
|---|---|---|
| 1st publish after connect | 0.0 s | 0.0 s |
| 2nd publish | **235.7 s** | 0.0 s |
| 3rd publish | 0.0 s | 0.0 s |
| 4th publish | **230.8 s** | 0.0 s |
| periodic heartbeat publish, steady state | **~180 s each** | 0.0 s each |
| last publish before shutdown | **235.5 s** | 0.0 s |
| `natsSubscription_Unsubscribe()` at shutdown | **239.9 s + 240.0 s** | 4 calls, 0.0 s each |
| application startup until ready | ~8 min | 25 s |
| shutdown, last protocol op to process exit | 480.0 s | 63 ms |

In steady state the publishing thread was blocked 180 out of every 240 seconds. All blocked calls
returned on the same 240-second grid; the grid was wall-clock independent and its phase changed
from run to run — consistent with the server's per-connection PING being the only inbound traffic.
Neither run logged a reconnect, disconnect or error.

Note that `test_ProxyConnectCb` cannot catch this: its `_proxyConnectCb` delegates straight back to
`natsSock_ConnectTcp()` on a fresh `natsSockCtx` with no callback set, so the library performs the
setup itself and the omission is invisible.

### Expected behavior

A connection established through `proxyConnectCb` should behave like any other connection:
`natsConnection_Publish*()` returns as soon as the message is buffered, and the socket carries the
same options the regular connect path sets (`TCP_NODELAY`, `SO_REUSEADDR`, `SO_LINGER`, and
non-blocking mode so the library's own deadline handling applies).

Concretely: `natsSock_ConnectTcp()` should apply `natsSock_SetBlocking(ctx->fd, false)` and
`natsSock_SetCommonTcpOptions(ctx->fd)` to the socket returned by the callback, before returning.

Alternatively, if the intent is that the application configures the socket, that requirement should
be documented on `natsProxyConnHandler` — though it cannot be met portably without the library's
internal `natsSock_*` helpers, which is why we think the library should do it.

### Server and client version

- Client: **nats.c v3.11.0**. The affected code is unchanged on `main` at the time of writing.
- Server: **nats-server v2.12.2**

### Host environment

- Client: Windows 11 x64, MSVC, OpenSSL, TLS enabled (`natsOptions_SetSecure(opts, true)`),
  `natsOptions_SetProxyConnHandler()` set; the callback establishes an HTTP CONNECT tunnel through
  a Squid proxy and returns the resulting blocking socket.
- Server: RHEL 9, reachable only through that proxy (NAT), TLS listener.
- Also relevant: no write deadline configured (`opts->writeDeadline == 0`), so the flusher's write
  has no deadline while it holds `nc->mu`.

### Steps to reproduce

Our own measurements above come from the proxy/TLS setup described under *Host environment*. From
reading the code, the proxy and the NAT should not be needed to reproduce it — all that should be
required is a `proxyConnectCb` that returns a **blocking** socket, plus TLS. Lowering the server's
`ping_interval` should shorten the stall from minutes to seconds. We have not run this reduced case
ourselves, so please treat it as a suggested starting point rather than a verified recipe.

1. Start a TLS-enabled `nats-server` with a short ping interval in its config:

   ```
   ping_interval: "10s"
   ```

2. Connect with TLS and a callback that creates the socket itself, without using the `natsSock_*`
   helpers — i.e. what an application implementing its own tunnel handshake does:

   ```c
   static natsStatus
   _blockingProxyConnectCb(natsSock *fd, char *host, int port, void *closure)
   {
       struct addrinfo hints, *res = NULL;
       char            sport[6];
       natsSock        s;

       snprintf(sport, sizeof(sport), "%d", port);
       memset(&hints, 0, sizeof(hints));
       hints.ai_family   = AF_UNSPEC;
       hints.ai_socktype = SOCK_STREAM;

       if (getaddrinfo(host, sport, &hints, &res) != 0)
           return NATS_SYS_ERROR;

       s = socket(res->ai_family, res->ai_socktype, res->ai_protocol);
       if (s == NATS_SOCK_INVALID)
       {
           freeaddrinfo(res);
           return NATS_SYS_ERROR;
       }

       /* Blocking socket, no options set - this is the state an application's
          own connect code typically leaves it in. */
       if (connect(s, res->ai_addr, (natsSockLen) res->ai_addrlen) != 0)
       {
           freeaddrinfo(res);
           return NATS_SYS_ERROR;
       }

       freeaddrinfo(res);
       *fd = s;
       return NATS_OK;
   }
   ```

   ```c
   natsOptions_SetSecure(opts, true);
   natsOptions_SetProxyConnHandler(opts, _blockingProxyConnectCb, NULL);
   natsConnection_Connect(&nc, opts);
   ```

3. Publish in a loop with a few seconds between messages and time each call:

   ```c
   for (i = 0; i < 10; i++)
   {
       int64_t t0 = nats_Now();
       natsConnection_PublishString(nc, "foo", "bar");
       printf("publish %d took %" PRId64 " ms\n", i, nats_Now() - t0);
       nats_Sleep(3000);
   }
   ```

Observed: the durations alternate between ~0 ms and roughly the remaining time until the server's
next PING (~10 s with the config above; minutes with the default `ping_interval`).
Expected: all ~0 ms.

Dropping `natsOptions_SetSecure()` makes the symptom disappear, because without TLS there is no
`ctx->sslMu` shared between the read loop and the writer.

We have the two-line fix running in production and are happy to open a PR with it.

wolfkor:
Fixes #1010

`natsSock_ConnectTcp()` invokes `proxyConnectCb` and returns right away, skipping the
`natsSock_SetBlocking(fd, false)` and `natsSock_SetCommonTcpOptions(fd)` calls that the regular
connect path applies a hundred lines below. The callback hands back a socket in whatever state the
application's own connect code left it in — typically blocking and without `TCP_NODELAY`.

With TLS that stalls every publish, because `_processConnInit()` deliberately keeps the socket
non-blocking when TLS is in use and so never fixes it up:

| thread | holds | waits for |
|---|---|---|
| `_readLoop` | `ctx->sslMu` — `natsSock_Read()` takes it around `SSL_read()` | `SSL_read()` parked inside a blocking `recv()` |
| `_flusher` | `nc->mu` — held across `natsConn_bufferFlush()` | `ctx->sslMu` |
| application thread | — | `nc->mu` in `natsConn_publish()` |

Every `natsConnection_Publish*()` then blocks until the server sends something on its own — in
practice until the next connection PING. Measurements from a production setup over an HTTP CONNECT
proxy are in #1010: publish durations alternating between 0.0 s and 180–240 s, startup 8 minutes
instead of 25 seconds, shutdown 480 s instead of 63 ms. With this change all of it disappears.

### Changes

- `src/comsock.c` — apply the same socket setup the regular path applies, and close the socket if
  either call fails.
- `test/test.c` — extend `test_ProxyConnectCb` with `_plainSocketProxyConnectCb`, which returns a
  plain blocking socket without using any `natsSock_*` helper, and assert via `getsockopt()` that
  `TCP_NODELAY` ended up set on `nc->sockCtx.fd`.

The existing `_proxyConnectCb` cannot catch this: it delegates straight back to
`natsSock_ConnectTcp()` on a fresh `natsSockCtx` with no callback set, so the library performs the
setup itself and the gap stays invisible. The new assertion fails without the `comsock.c` change.

### Testing

The `comsock.c` change has been running in production over an HTTP CONNECT proxy with TLS, which is
where the numbers above come from.

`ProxyConnectCb` was run locally on Windows (MSVC 19.51, `-DNATS_BUILD_WITH_TLS=OFF`,
nats-server v2.12.2) — the new assertion is a genuine regression test:

```
# with this change
#07 Set connectCb that returns a plain socket: PASSED
#08 Connect with connectCb returning a plain socket: PASSED
#09 Socket setup applied by the library: PASSED
ALL PASSED

# same test, src/comsock.c reverted to main
#09 Socket setup applied by the library: FAILED
*** TEST FAILED ***
```

Note that the test does not need TLS: it asserts on the socket setup itself, which is what goes
missing. TLS is only what turns the missing setup into the stall described above.

### Notes

After this change the proxy path behaves like the non-proxy path, including that the connect
deadline `_createConn()` installs from `opts->timeout` now also covers the TLS handshake, where it
previously did not apply at all on this path. Applications behind a slow proxy may need a larger
`natsOptions_SetTimeout()`.

### DIFF
--- a/src/comsock.c
+++ b/src/comsock.c
@@ -187,6 +187,24 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
         // Invoke the proxy connect callback.
         s = ctx->proxyConnectCb(&ctx->fd, host, port, ctx->proxyConnectClosure);
 
+        // The callback hands back a socket it created itself, so it has not been
+        // through the setup that the regular connect path below performs. Apply
+        // the same settings here. Leaving the socket in blocking mode is not an
+        // option: with TLS, natsSock_Read() holds ctx->sslMu across SSL_read(),
+        // so a blocking read would stall every concurrent write - and with it
+        // the flusher thread, which holds nc->mu while flushing.
+        if (s == NATS_OK)
+            s = natsSock_SetBlocking(ctx->fd, false);
+
+        if (s == NATS_OK)
+            s = natsSock_SetCommonTcpOptions(ctx->fd);
+
+        if (s != NATS_OK)
+        {
+            _closeFd(ctx->fd);
+            ctx->fd = NATS_SOCK_INVALID;
+        }
+
         // If there was a deadline, reset the deadline with whatever is left.
         if (totalTimeout > 0)
             _resetDeadline(ctx, start, totalTimeout);
--- a/test/test.c
+++ b/test/test.c
@@ -6421,6 +6421,60 @@ _proxyConnectCb(natsSock *fd, char *host, int port, void *closure)
     return s;
 }
 
+// Returns a plain blocking socket, without going through any of the natsSock_*
+// helpers. This is what a callback that performs its own connect (say an HTTP
+// CONNECT tunnel to a proxy) hands back, and the library is responsible for
+// applying the socket setup to it.
+static natsStatus
+_plainSocketProxyConnectCb(natsSock *fd, char *host, int port, void *closure)
+{
+    struct threadArg    *arg      = (struct threadArg*) closure;
+    struct addrinfo     *servinfo = NULL;
+    struct addrinfo     hints;
+    natsStatus          s         = NATS_OK;
+    natsSock            sock      = NATS_SOCK_INVALID;
+    char                sport[6];
+    int                 res;
+
+    natsMutex_Lock(arg->m);
+    arg->sum++;
+    natsMutex_Unlock(arg->m);
+
+    snprintf(sport, sizeof(sport), "%d", port);
+
+    memset(&hints, 0, sizeof(hints));
+
+    hints.ai_family   = AF_INET;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+
+    if ((res = getaddrinfo(host, sport, &hints, &servinfo)) != 0)
+        s = NATS_SYS_ERROR;
+
+    if (s == NATS_OK)
+    {
+        sock = socket(servinfo->ai_family, servinfo->ai_socktype,
+                      servinfo->ai_protocol);
+        if (sock == NATS_SOCK_INVALID)
+            s = NATS_SYS_ERROR;
+    }
+    if ((s == NATS_OK)
+        && (connect(sock, servinfo->ai_addr, (natsSockLen) servinfo->ai_addrlen) != 0))
+    {
+        s = NATS_SYS_ERROR;
+    }
+
+    if (servinfo != NULL)
+        freeaddrinfo(servinfo);
+
+    if (s == NATS_OK)
+        *fd = sock;
+    else if (sock != NATS_SOCK_INVALID)
+        natsSock_Close(sock);
+
+    return s;
+}
+
 void test_ProxyConnectCb(void)
 {
     natsStatus          s;
@@ -6472,6 +6526,37 @@ void test_ProxyConnectCb(void)
     s = (args.sum == 1 ? NATS_OK : NATS_ERR);
     args.sum = 0;
     natsMutex_Unlock(args.m);
+    testCond(s == NATS_OK);
+
+    natsConnection_Destroy(nc);
+    nc = NULL;
+
+    test("Set connectCb that returns a plain socket: ");
+    s = natsOptions_SetProxyConnHandler(opts, _plainSocketProxyConnectCb, (void*) &args);
+    testCond(s == NATS_OK);
+
+    test("Connect with connectCb returning a plain socket: ");
+    s = natsConnection_Connect(&nc, opts);
+    testCond(s == NATS_OK);
+
+    // The socket the callback returned had no options set. Check that the
+    // library applied its own setup to it, as it does on the regular path.
+    test("Socket setup applied by the library: ");
+    {
+        int         nodelay = 0;
+        natsSockLen len     = (natsSockLen) sizeof(nodelay);
+
+        s = NATS_OK;
+        if (getsockopt(nc->sockCtx.fd, IPPROTO_TCP, TCP_NODELAY,
+                       (char*) &nodelay, &len) != 0)
+        {
+            s = NATS_SYS_ERROR;
+        }
+        else if (nodelay == 0)
+        {
+            s = NATS_ERR;
+        }
+    }
     testCond(s == NATS_OK);
 
     natsConnection_Destroy(nc);


wolfkor:
The failing check, `Sanitize (clang, thread, NO-pool)`, looks unrelated to this PR.

It fails in `test/no_crash_on_exit`, which runs before `testsuite`, on a data race on the global
`gLib` between the `atexit` handler and `nats_CloseAndWait()`:

```
WARNING: ThreadSanitizer: data race (pid=6368)
  Write of size 4 at 0x55cc7bf969bc by thread T4:
    #1 _freeLib src/glib/glib.c:117:5
    #2 natsLib_Release src/glib/glib.c:61:9
    #3 _libTearDown src/glib/glib.c:186:5
    #4 nats_closeLib src/glib/glib.c:361:5
    #5 nats_CloseAndWait src/nats.c:163:12
    #6 _close test/no_crash_on_exit/nocrashonexit.c:22:5

  Previous read of size 4 at 0x55cc7bf969bc by main thread (mutexes: write M0):
    #0 natsLib_Destructor src/glib/glib.c:84:15
    #1 at_exit_callback_installed_at() tsan_interceptors_posix.cpp.o
    #2 _doInitOnce src/glib/glib.c:167:5

Location is global 'gLib' of size 296
SUMMARY: ThreadSanitizer: data race in __tsan_memset
```

`nocrashonexit` does not set a `proxyConnectCb`, so the `comsock.c` change in this PR is not
reachable from it, and `testsuite` — which contains the changed `ProxyConnectCb` test — never ran
in that job. The sibling job `Sanitize (gcc, thread, NO-pool)` passed.

Could someone with write access re-run the job? I am happy to open a separate issue for the `gLib`
race if that would be useful.

AdamPayzant:
This looks okay to me, but @kozlovic could you also take a look and make sure I'm not missing something here?

src/comsock.c
```diff
@@ -187,6 +187,24 @@ natsSock_ConnectTcp(natsSockCtx *ctx, const char *phost, int port)
         // Invoke the proxy connect callback.
         s = ctx->proxyConnectCb(&ctx->fd, host, port, ctx->proxyConnectClosure);
 
+        // The callback hands back a socket it created itself, so it has not been
+        // through the setup that the regular connect path below performs. Apply
+        // the same settings here. Leaving the socket in blocking mode is not an
+        // option: with TLS, natsSock_Read() holds ctx->sslMu across SSL_read(),
+        // so a blocking read would stall every concurrent write - and with it
+        // the flusher thread, which holds nc->mu while flushing.
+        if (s == NATS_OK)
```
kozlovic: I wonder if we should then add the following (that is done in non proxy cases):
```
#ifdef SO_NOSIGPIPE
    if (s == NATS_OK)
    {
            int set = 1;
            if (setsockopt(ctx->fd, SOL_SOCKET, SO_NOSIGPIPE, (void*)&set, sizeof(int)) == -1)
            {
                s = nats_setError(NATS_SYS_ERROR,
                                  "setsockopt SO_NOSIGPIPE error: %d",
                                  NATS_SOCK_GET_ERROR);
            }
    }
#endif
```

kozlovic (approved):
Code change and test addition looks good to me!

PR #1017, opened 2026-08-31
Adding kvStore_GetAsync

### LINKED ISSUE #1014: Async/event driven/non-blocking KV/Core
Add non-blocking async request/reply and kvStore_GetAsync()

Follow-up to #744 / #1012.

PR #1012 adds asynchronous callbacks for KV watches. That solves the long-running/watch side of our use case.

The other operation we need is a one-shot KV Get that does not block an application worker while waiting for JetStream/NATS to reply.

Use case

Our primary use case is FreeRADIUS v4 integration.

FreeRADIUS is designed around asynchronous I/O: a request yields while waiting for an external operation and is resumed when that operation completes. Blocking a worker thread for each outstanding lookup is therefore undesirable.

Conceptually, we would like to be able to do:

```c
kvStore_GetAsync(kv, key, callback, closure);

/* returns immediately; application can yield */

...

callback(entry, status, closure);
```

The callback does not need to execute on the FreeRADIUS event-loop thread. It can signal the event loop, which can then resume the yielded RADIUS request in the correct execution context.

KV Get is the first concrete requirement, but the same non-blocking primitive would also be useful for other NATS / JetStream operations later.

What we are asking for

The important requirement is not merely a callback wrapper around the existing synchronous kvStore_Get().

We would like the request to remain non-blocking all the way down to the NATS request/reply machinery:

```text
application
    |
    v
kvStore_GetAsync()
    |
    v
JetStream Direct/Get
    |
    v
non-blocking NATS request/reply
    |
    +---- return immediately

         ...

reply arrives
    |
    v
correlate outstanding request
    |
    v
decode JetStream / KV response
    |
    v
callback
```

In particular, an implementation that simply runs the existing blocking kvStore_Get() on another worker thread would work functionally, but would not solve the scalability/event-loop problem we are trying to address.

Proposed building blocks

1. Reusable non-blocking request/reply

Provide a reusable asynchronous request/reply mechanism in nats.c that can be used by JetStream Direct Get and other APIs.

The exact public API is open for discussion.

Illustratively:

```c
typedef void (*natsRequestHandler)(
    natsConnection *nc,
    natsMsg *reply,
    natsStatus status,
    void *closure);

natsStatus natsConnection_RequestAsync(
    natsAsyncRequest **req,
    natsConnection *nc,
    const char *subject,
    const void *data,
    int dataLen,
    int64_t timeout,
    natsRequestHandler cb,
    void *closure);
```

The important semantics would be:

• register/send the request and return immediately;
• invoke completion when the reply arrives;
• report timeout, no-responders and other errors asynchronously;
• support many concurrent outstanding requests;
• not require one blocked thread per outstanding request;
• not require one dedicated reply subscription per outstanding request;
• preferably use shared connection-level response multiplexing;
• clearly document which thread / execution context invokes the callback.

Conceptually, the connection would maintain outstanding logical requests and correlate replies back to them:

```text
shared reply subscription

_INBOX.<connection-id>.*
        |
        +-- token A -> request A
        +-- token B -> request B
        +-- token C -> request C
```

Each outstanding request would carry the state needed to complete it, for example:

```text
request A
    |
    +-- correlation token
    +-- callback
    +-- closure
    +-- deadline
    +-- lifecycle state
```

The exact internal implementation is of course up to the maintainers; the important requirement is efficient multiplexing without a blocked thread or dedicated subscription per logical request.

2. Explicit in-flight operation lifecycle

It seems useful for an asynchronous request to have an explicit operation handle, rather than exposing only callback + closure.

For example:

```c
natsAsyncRequest *req = NULL;

natsConnection_RequestAsync(
    &req,
    nc,
    subject,
    data,
    dataLen,
    timeout,
    callback,
    closure);

/* originating application request disappears */
natsAsyncRequest_Cancel(req);

natsAsyncRequest_Destroy(req);
```

This would give the application a clean way to represent and cancel an in-flight operation without exposing internal inbox/correlation details.

This is especially relevant for FreeRADIUS: the RADIUS request may be cancelled or disappear before the NATS reply arrives.

The important lifecycle guarantee would be that reply, timeout and cancellation have well-defined race semantics and exactly one of them completes the operation.

Conceptually:

```text
                 PENDING
                /   |   \
               /    |    \
          reply   timeout  cancel
             \      |      /
              \     |     /
               one path wins
                    |
                    v
                COMPLETE
```

A late reply to a cancelled/timed-out request should be safely discarded.

The exact ownership/reference model of the operation handle is open for discussion; the goal here is to make request lifetime and cancellation explicit.

3. Asynchronous JetStream Get

KV Get currently ultimately performs a JetStream message lookup / Direct Get.

That layer therefore also needs a non-blocking path.

Whether APIs such as these should be public:

```c
js_DirectGetMsgAsync(...)
js_GetMsgAsync(...)
js_GetLastMsgAsync(...)
```

or remain an internal implementation detail is probably a separate API design decision.

The important point is that the JetStream lookup should use the non-blocking request/reply primitive rather than calling the existing synchronous request API internally.

4. kvStore_GetAsync()

Once the underlying request path is asynchronous, expose it through KV.

Illustrative API only:

```c
typedef void (*kvGetHandler)(
    kvStore *kv,
    kvEntry *entry,
    natsStatus status,
    void *closure);

natsStatus kvStore_GetAsync(
    kvAsyncGet **req,
    kvStore *kv,
    const char *key,
    kvGetHandler cb,
    void *closure);
```

With an equivalent operation for a specific revision if appropriate.

Internally this should remain a thin KV layer:

```text
NATS reply
    |
    v
JetStream response validation
    |
    v
KV metadata / entry decoding
    |
    v
kvEntry
    |
    v
user callback
```

Whether KV exposes its own operation type (kvAsyncGet) or reuses the underlying async request handle is also an API-design choice.

Desired semantics

For an event-loop integration, the important guarantees are:

1. Starting a Get does not wait for the server response.
2. No dedicated blocked thread is required per outstanding Get.
3. No dedicated reply subscription is required per outstanding Get.
4. Completion is delivered exactly once.
5. Completion covers success, not-found, timeout, no-responders, cancellation and other errors.
6. The callback receives the caller-supplied closure.
7. Ownership/lifetime of the returned kvEntry is clearly defined.
8. Cancellation, timeout and response races are well-defined.
9. The callback execution context/thread is documented.
10. The design works both with the normal nats.c threading model and external event-loop integrations.

Why start at core request/reply?

It would be possible to implement only kvStore_GetAsync(), but if it internally performs a synchronous JetStream/NATS request on another worker thread, the underlying problem remains.

A genuine asynchronous request/reply primitive is also reusable by other JetStream and NATS APIs that currently need to wait for a response.

So the layering we have in mind is roughly:

```text
Core non-blocking request/reply
          |
          v
In-flight request lifecycle / cancellation
          |
          v
JetStream async request operations
          |
          v
kvStore_GetAsync()
```

PR #1012 establishes the callback model for KV watches. A non-blocking Get built on a reusable request/reply primitive would complete the other half of the event-driven KV use case.

Contribution

We can help test the client-side behavior against the FreeRADIUS v4 yield/resume model.

We do not have significant C experience, so the API sketches above are intentionally illustrative rather than prescriptive. If assumptions about the current nats.c internals or preferred API direction are wrong, feedback is very welcome.

AdamPayzant:
Adding new public KV API for kvStore_GetAsync, as well as internal jetstream functions: `js_directGetMsgAsync`, `js_requestAsync`, and `js_getMsgAsync`.

Resolves #1014


src/nats.h
```diff
@@ -8561,6 +8593,32 @@ kvEntry_Destroy(kvEntry *e);
...
+ *
+ * @see kvGetCb
+ *
+ * @param kv the pointer to the #kvStore object.
+ * @param key the name of the key.
+ * @param cb the callback function to invoke when the entry is retrieved.
+ * @param cbClosure a pointer to an user defined object (can be `NULL`).
+ */
+NATS_EXTERN natsStatus
+kvStore_GetAsync(kvStore *kv, const char *key, kvGetCb cb, void *cbClosure);
```
kozlovic, line 8620: I wonder if we should not instead have a "kvStore_GetWithOptions()` or something like that, which will cover all possible cases (including async with revision which you don't cover here). The point is that maybe we could have a version that accepts options where you can set all information that is needed (some options may be mutually exclusive). For the internal JS APIs, you could have a generic "get" that accepts some options structure that covers all possible cases (direct or not, last msg, per subject, async, etc...)

kozlovic (commented):
Wonder if we should then go a step further and have the `kvStore_GetAsync` be something like `kvStore_GetWithOptions()` where we use an options structure instead that would cover all possible cases (and easier to add options in the future without having a need for a new API. But I understand if this is too much, and if so, do you need then one for the "kvStore_GetRevision"?
Again, this is more of a question if we think that the API is really stable and we should not have too many other "get" versions. If that is the case, you can add the GetRevisionAsync and not go the "WithOptions" way.

src/nats.h
AdamPayzant: For distinguishing between async/sync I prefer them as distinct functions as I think that's more clear what's being done and how the code around it should be handled. As far as I can recall that's also consistent with how it's done elsewhere, but maybe I'm forgetting something. Regardless I do think unifying the internal JS APIs would be worth looking at so I'll take a look at that and maybe go back on that once I have that in place

kozlovic (approved):
LGTM

AdamPayzant:
I'm still experimenting with making single jetstream function for it, but at this point it should probably be it's own PR rather than tacked onto this one. Will merge this one for now

sdktr:
wauw that was fast:) thanks for adding async KV support — this looks useful, shared it with Freeradius maintainers. Their initial question is regarding thread safety:

>One important concern: these callbacks appear to run on internal dispatch/worker threads rather than the thread that initiated the request.

>Could we explicitly document the callback threading model for kvStore_GetAsync (and the related async JS helpers)? In particular:

>Whether callback execution is guaranteed to be off-thread / unspecified thread.
>Whether multiple callbacks can run concurrently.
>Any ordering guarantees per subscription/request context.
>Without this, consumers may accidentally touch non-thread-safe state from callbacks. A short note in API docs + header comments (and maybe an example showing queueing work back to an app-owned thread) would help prevent subtle races.

AdamPayzant:
I'll look at updating the docs, but just to go ahead and answer your questions here:

> > Whether callback execution is guaranteed to be off-thread / unspecified thread.

It will always run on the internal thread tied to it's associated jsCtx.

> > Whether multiple callbacks can run concurrently.

The callbacks would be run sequentially as recieved. If you need them to run concurrently, you would need to dispatch to your own threads from the callback.

> > Any ordering guarantees per subscription/request context.

There should be one response per call in order of the responses recieved, not the order of calls.

kozlovic:
@AdamPayzant I am not sure the following is entirely correct:

> It will always run on the internal thread tied to it's associated jsCtx.

I think you are also invoking the callback on jsCtx destroy, which means the callbacks could then run from a different thread, and possibly concurrently. Also, if a request times out, it is invoked from the timer thread. So there are really multiple places from where the callback can be invoked, which is not ideal. I wonder if you should not have queued all these occurrences to the subscription's queue, so they are at least coming from the same thread...

AdamPayzant:
> I think you are also invoking the callback on jsCtx destroy, which means the callbacks could then run from a different thread, and possibly concurrently. Also, if a request times out, it is invoked from the timer thread. So there are really multiple places from where the callback can be invoked, which is not ideal. I wonder if you should not have queued all these occurrences to the subscription's queue, so they are at least coming from the same thread...

That's a fair point, good catch! I'm already working on unifying the jetstream builder side, so I'll see what I can do about queuing them as well.

PR #1019, opened 2026-09-07
Add natsConnection_GetConnectedServerName

pandazz77:
The `NATS Server` reports its name in the `INFO` protocol via the `server_name`
field (present since server v2.0.0), but the C client had no way to expose it.
Applications that need to know which server in a cluster they are talking to
currently have to fall back on the server ID, which is a generated identifier
rather than the operator-assigned name.

This PR parses `server_name` and adds a public getter for it.


kozlovic (approved):

kozlovic:
@AdamPayzant I have reviewed, but leaving it for your consideration.

PR #1020, opened 2026-09-08
[FIXED] Data race between library's destructor and `nats_Close[AndWait]()`

kozlovic:
This was reported by the test `nocrashonexit` when running with the sanitizer. The issue was that we were clearing the memory of the `natsLib` structure starting at `refs` instead of past that field.

### DIFF
--- a/src/glib/glib.c
+++ b/src/glib/glib.c
@@ -100,7 +100,7 @@ natsLib_Destructor(void)
 static void
 _freeLib(void)
 {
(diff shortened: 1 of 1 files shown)


AdamPayzant (approved):
I believe the failing checks are unrelated, but please verify

Otherwise LGTM

kozlovic:
@AdamPayzant Yes, I am re-running the failing tests, let's wait for green run.

kozlovic:
@AdamPayzant All green, ready to merge.

PR #1021, opened 2026-09-09
[Updated] kvStore_GetAsync follow-up

AdamPayzant:
Follow-up from discussion in https://github.com/nats-io/nats.c/pull/1017.

Contains the following changes:
* Improves the documentation and adds example for kvStore_GetAsync
* Improved documentation for thread model. Moving the callback on one thread caused too many issues and added more complexity to the callback than just requiring it to be threadsafe
* Unifies jetstream get message to better unify get, get-last, direct-get for both sync and async kv gets

Update: Unified thread model has too much friction and caveats, I believe requiring the callback to be threadsafe is the best option. Reverting back to original design with better documentation.

### DIFF
--- a/examples/kv-get-async.c
+++ b/examples/kv-get-async.c
@@ -1,0 +1,285 @@
+// Copyright 2026 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


AdamPayzant:
@kozlovic I had to remove the thread guarantees. I tried making it so that the natsStatus would distinguish would determine which thread it was coming from, but it was already clunky from a callback perspective and wasn't great from an implementation side either. This still has the unified jetstream get message and improved documentation

AdamPayzant:
@sdktr Please also feel free to take a look and let me know if the documentation covers what you need to know. I'd rather fix it now than opening another PR

kozlovic (approved):
LGTM

PR #1022, opened 2026-09-14
[FIXED] ObjectStore: puts honor the context's PublishAsync flow control

zsombor-szalay:
`objStore_Put()` publishes chunks through a private JetStream context created with `jsOptions_Init()`, so the `PublishAsync.MaxPending` / `StallWait` configured on the store's context are ignored and an upload has no in-flight limit. Against a server whose per-stream ingest buffer (`jetstream.max_buffered_size`) is smaller than the object, chunks are dropped with `429 too many requests` and the put fails.

This copies `MaxPending` and `StallWait` from the store's context into the put context. Defaults are unchanged (`MaxPending == 0` remains unlimited); the private context is kept for the per-put error handler and completion scope.

```c
o.PublishAsync.MaxPending = 48;   // 48 x 128 KiB chunks in flight
o.PublishAsync.StallWait  = 5000;
natsConnection_JetStream(&js, nc, &o);
js_CreateObjectStore(&obs, js, &cfg);
```

Adds `ObjectStore_PutFlowControl`: a put on a `MaxPending = 1` context carries the values, uploads a 16 KiB object in 1 KiB chunks and reads it back.


kozlovic (changes requested):
Would be nice to update the test...

test/test.c
```diff
@@ -39997,6 +39997,72 @@ void test_ObjectStore_PutAndGet(void)
     JS_TEARDOWN;
 }
 
+void test_ObjectStore_PutFlowControl(void)
```
kozlovic, line 40000: The test would be fine, but you check really that the options are set, which could be enough. But if I comment out the code change and remove the check to see if the options are set, the test still pass, where with the test below it would not. Could you change the test to this:
```
diff --git a/test/test.c b/test/test.c
index 13543d32..fee4641c 100644
--- a/test/test.c
+++ b/test/test.c
@@ -40000,7 +40000,9 @@ void test_ObjectStore_PutAndGet(void)
 void test_ObjectStore_PutFlowControl(void)
 {
     natsStatus      s;
-    jsCtx           *js2    = NULL;
+    natsConnection  *nc     = NULL;
+    jsCtx           *js     = NULL;
+    natsPid         pid     = NATS_INVALID_PID;
     objStore        *obs    = NULL;
     objStorePut     *put    = NULL;
     objStoreInfo    *info   = NULL;
@@ -40011,8 +40013,28 @@ void test_ObjectStore_PutFlowControl(void)
     jsOptions       o;
     objStoreConfig  cfg;
     objStoreMeta    meta;
+    char confFile[256] = {'\0'};
+    char datastore[256] = {'\0'};
+    char cmdLine[1024] = {'\0'};
 
-    JS_SETUP(2, 10, 0);
+    ENSURE_JS_VERSION(2, 11, 0);
+
+    test("Start server: ");
+    _makeUniqueDir(datastore, sizeof(datastore), "datastore_");
+    _createConfFile(confFile, sizeof(confFile),
+        "jetstream: {\n"
+        "   enabled: true\n"
+        "   max_buffered_size: 10Kb\n"
+        "   max_buffered_msgs: 100\n"
+        "}\n");
+    snprintf(cmdLine, sizeof(cmdLine), "-js -sd %s -c %s", datastore, confFile);
+    pid = _startServer("nats://127.0.0.1:4222", cmdLine, true);
+    CHECK_SERVER_STARTED(pid);
+    testCond(true);
+
+    test("Connect: ");
+    s = natsConnection_Connect(&nc, NULL);
+    testCond(s == NATS_OK);
 
 #ifndef NATS_HAS_TLS
     nats_hashNoErrorOnNoSSL(true);
@@ -40022,13 +40044,13 @@ void test_ObjectStore_PutFlowControl(void)
     jsOptions_Init(&o);
     o.PublishAsync.MaxPending = 1;
     o.PublishAsync.StallWait  = 5000;
-    s = natsConnection_JetStream(&js2, nc, &o);
-    testCond((s == NATS_OK) && (js2 != NULL));
+    s = natsConnection_JetStream(&js, nc, &o);
+    testCond((s == NATS_OK) && (js != NULL));
 
     test("Create ObjectStore: ");
     objStoreConfig_Init(&cfg);
     cfg.Bucket = "FLOW";
-    s = js_CreateObjectStore(&obs, js2, &cfg);
+    s = js_CreateObjectStore(&obs, js, &cfg);
     testCond((s == NATS_OK) && (obs != NULL));
 
     test("Put inherits the store's flow control: ");
@@ -40058,7 +40080,6 @@ void test_ObjectStore_PutFlowControl(void)
     objStoreInfo_Destroy(info);
     objStorePut_Destroy(put);
     objStore_Destroy(obs);
-    jsCtx_Destroy(js2);
 
     JS_TEARDOWN;
 }
```

kozlovic (changes requested):

test/test.c
zsombor-szalay: did as requested

kozlovic (approved):
LGTM

